PR #3, opened 2015-04-27
Added devBuild at Trunk level to debug SolDesktop.java without the "res/...

theotherjay:
To debug "out-of-the-box", the devBuild file needed to exist in trunk otherwise DebugOptions.DEV_ROOT_PATH is never defined.  This is a problem at line 45 in SolDesktop.java because the relative path to the resource is incorrect (yes, we could copy the icon somewhere else or change this logic, but adding the devBuild file was the simplest solution):

    c.title = "Destination Sol";
    if (DebugOptions.DEV_ROOT_PATH == null) {
      c.addIcon("res/icon.png", Files.FileType.Internal);
    } else {
      c.addIcon(DebugOptions.DEV_ROOT_PATH + "res/icon.png", Files.FileType.Absolute);
    }


Cervator:
Thanks! This along with our talk on IRC helped me simplify the local setup without needing a custom working directory in IntelliJ :-)

First merged PR woo!

PR #4, opened 2015-04-28
Renamed SolCmp to SolApplication

PrivateAlpha:
Made SolApplication implement ApplicationListener
Removed the SolAppListener, since it was just a wrapper for SolApplication, with no logic in it


PrivateAlpha:
This is really just a bit of a simple clean up, SolAppListener is just a wrapper class for SolCmp(and has no logic in it), so removed it and made SolCmp(renamed to SolApplication) the entry point for the application

theotherjay:
I'm glad someone is doing this.  Useful class names will help as well as
removing unnecessary files.

PR #15, opened 2015-05-04
Removed SolFile class in favor of a new FileManager class

LinusVanElswijk:
The code contained hardcoded filepaths in many classes that had to read from or write to files.
The SolFiles class also required a boolean flag for each file access, so the user had to know whether or not the file was located in the readonly file directory or the writable file directory.

The FileManager solves both these issues. It internally contains (the most common) directory paths
and can provide users with the correct filehandle for these directories, without having to provide any additional arguments.

This PR also changes the terms for "readonly" and "writable" files to "static" and "dynamic" files.


Cervator:
Excellent cleanup, thanks!

PR #16, opened 2015-05-05
Disabled FPS throttling and added FPS output to console while in 'debug mode'

theotherjay:
- Disabled FPS throttling when in 'debug mode' (when devBuild file is present)

- Added FPS output to console when in 'debug mode' and FPS logging is enabled.  Previously, this was only being rendered to the game UI.

### DIFF
--- a/desktop/src/com/miloshpetrov/sol2/desktop/SolDesktop.java
+++ b/desktop/src/com/miloshpetrov/sol2/desktop/SolDesktop.java
@@ -23,14 +23,18 @@ public class SolDesktop {
             return;
         }
 
+        LwjglApplicationConfiguration c = new LwjglApplicationConfiguration();
         boolean devBuild = java.nio.file.Files.exists(Paths.get("devBuild"));
         if (devBuild) {
             DebugOptions.DEV_ROOT_PATH = "main/"; // Lets the game run from source without a tweaked working directory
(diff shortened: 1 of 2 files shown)


Cervator:
Thanks for the change :-)

I missed your visit on IRC, don't feel out of place there! It is often quiet, but it isn't from people getting ignored. Just mostly AFK :-)

Hoping to test this ASAP but am currently trying to get something else to work

theotherjay:
No rush bro.  I'll probably continue improving/tweaking small things like this until I'm familiar enough with the lay of the land to start thinking about larger changes.

Cervator:
Looks good, and hah, finally I get to see something log to the console! :D

Sheesh, getting near 3,000 FPS here, that's a bit. Any specific use case for disabling throttling?

theotherjay:
Throttling can hide performance impacting changes.  I like to know as I
develop changes what my performance impact is instead of waiting until down
the road when other work could mask a high impact change.  This way, they
can be caught quickly.  3000 fps, eh?  That's quite fast!  My laptop was
only pumping out ~ 900 :D

Cervator:
Fair enough, sounds good :-)

Eventually we can probably put in some automated performance testing, but that's a ways out / low priority I imagine

PR #19, opened 2015-05-20
Update LibGDX

SimonC4:
I started out with the intention of supporting building on OS X using Java 8. There is currently an exception that occurs using Java 8 that doesn't occur if you use Java 7.  To achieve this I updated LibGDX to the latest version, 1.6.0. There were some deprecated API calls that needed to be fixed up including:
* Updated to OpenGL ES 2.0 as 1.0 is no longer supported. http://www.badlogicgames.com/wordpress/?p=3311
* Box2D moved to an extension: http://www.badlogicgames.com/wordpress/?p=3404
* Bitmap Font refactoring: http://www.badlogicgames.com/wordpress/?p=3658

### Android Concerns
I'm not sure how the Android app gets built so I'm not sure what the effect will be on that. I suspect that at the least there need to be Android specific LibGDX libraries that need to be updated.


Cervator:
Hi @SimonC4 and welcome, thanks for the nice PR :-)

Yep we should check Android for this one. I hope you don't mind if we wait to merge till the weekend so there's a bit of extra time to test?

SimonC4:
Thanks. That's fine. I'd prefer it if some testing was done to see if there are any other issues.

Cervator:
Good stuff!

I've tested this PR normally now and it works fine on LibGDX 1.6.0, thank you very much for finding the needed tweaks for the upgrade!

I had previously attempted to just use Gradle to resolve the LibGDX dependencies remotely, but it failed since there were changes needed and I didn't go further at the time. This allowed me to go revisit that and splice in the proper LibGDX Gradle project template.

As part of doing that I went ahead and upgraded to 1.6.1 since that's latest and only a patch-level release so it shouldn't cause trouble. I also got the Android piece mostly working, woo!

I put our combined work into a temporary branch:https://github.com/MovingBlocks/DestinationSol/tree/pr19plusAndroid

It *almost* works but not quite. I can launch the game fine from IntelliJ as a normal Java app. Actually writing out a game package fails to launch, probably from the changed library jar lineup, I need to sort that out.

Android I've gotten to a point where I can launch an emulator and have it start loading a signed version of the app using our super secret key store that'll supposedly allow us to publish on Google Play. But then it fails to load an image directory I haven't correctly copied to the right place yet. That's fine - simple fix, but it is 3:30 am here :D

Feel free to look around! There's a bit more work to go, hopefully can finish up and merge the combined work tomorrow.

Also curiously I'm guessing https://github.com/SimonC4/DestinationSol/commit/ae66bb16612ed983e847ad102e0b280c8096db0f only was needed after the upgrade? If I put that if back I got two weird wobbly cursors, never seen that before.

SimonC4:
Nice work using to use Gradle for LibGDX.

https://github.com/SimonC4/DestinationSol/commit/ae66bb16612ed983e847ad102e0b280c8096db0f was an interesting one. The setCursorPosition function changed its point of origin as described in this post: https://github.com/libgdx/libgdx/issues/1003

On further review, it looks like I've deleted some functionality. That function was also "clamping" the mouse, by not letting it leave the screen area. I'll submit another PR to fix that up.

Cervator:
Arrr this might take a bit longer than expected. I pulled in the other PR just fine, but am having trouble getting the Android emulator to behave.

Previously (Saturday) I was able to get the new structure far enough to where it complained about not being able to find some game assets. 

Just to test I used the old strategy to simply copy the `res` dir into `assets` under the `android` project. Tried that, but no luck, today the emulator doesn't even get that far anymore.

Old hacky project shows the same behavior. It was able to find and uninstall the app installed from my main workspace on the emulator, so at least that part works. But then it stalls the same way, so probably there's something wrong on my local system now that I need to fix.

If you're interested I can upload the old project to GitHub to see if you can make it work. You'd need to make a keystore and supply its info but that shouldn't be terribly difficult.

Anyway that's just in case you want to give it more of a try, I'll keep at it :-)

Last bit of logging for reference (in old project), after this the execution just exits

```
Launching application: com.miloshpetrov.sol2.android/com.miloshpetrov.sol2.android.SolAndroid.
DEVICE SHELL COMMAND: am start -n "com.miloshpetrov.sol2.android/com.miloshpetrov.sol2.android.SolAndroid" -a android.intent.action.MAIN -c android.intent.category.LAUNCHER
Starting: Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] cmp=com.miloshpetrov.sol2.android/.SolAndroid }
```

PR #21, opened 2015-05-24
Fixed up issue identified in PR19

SimonC4:
As mentioned at: https://github.com/MovingBlocks/DestinationSol/pull/19. The mouse is supposed to be "clamped" so it doesn't go off the edge of the screen. Includes the fix to prevent the two mouse pointers due to this change: https://github.com/libgdx/libgdx/issues/1003


Cervator:
You don't need to worry about closing these early, they'll auto-merge if #22 gets merged, or any other version of it :-)

Then the ultimate state of this would become Merged instead of Closed

SimonC4:
Ah, ok. Good to know. :-)

PR #17, opened 2015-05-06
Refactoring: new hull folder structure

LinusVanElswijk:
All data required for a hull config is now contained within a single folder.
Adding new hulls is as simple as adding a new folder under res/configs/hulls and adding the texture, icon and properties to it.

I also tried to clean up method and attribute names in the classes that had to be changed; making them more descriptive.

There is still some future work to be done:

1) For some reason, the old files can not be erased completely. This makes the game crash on the first update, without giving any hint on what went wrong.
This is the case for config/paths/hulls.json and config/hull.json.
These are left as empty json files as a temporary solution.

2) I had to add additional functions to the PathLoader class, to bypass all hardcoded file paths. I could not change them as that would break the game in other places, that still rely on these hardcoded file paths. The PathLoader class should be removed/replaced or recoded entirely; In it currents form, it greatly hinders modifications to the file structure.

The reason I'm doing a PR with these issues still open is because these issues are not introduced by the new code. The modification to the code just made these already existing issues visible.

PR #22, opened 2015-05-26
Great convergence all over again

Cervator:
NOT READY YET

This is a draft PR just to show progress and maybe let others play with it some + provide feedback.

It includes PRs #19 #20 and #21 (merged before I started moving everything around)

It has an Android facade that doesn't work any more after I moved the "res" folder into a more traditional spot, akin to how we handle assets in Terasology. I'm pretty sure that's because I changed the path/file handling to work with the new location and bundling in a jar, which isn't getting along with how files load on Android.

https://github.com/libgdx/libgdx/wiki/File-handling has more on it and I see a fair bit of that in the code. That whole area is somewhat messy and could use some more improvements.

In short I was trying to just make Android correctly load the assets from where they were already, rather than having to copy paste the res dir into a separate Android project (how it worked before). That led to simply moving the res dir to somewhere more conventional so I could use Gradle more cleanly, and then since I was already changing everything might as well rename "main" to "core" and update the package and before this is over I'm sure there'll be some Checkstyle config, massive code reformatting, and shiny new file headers ...

Yeah this is the Great Convergence for Terasology all over again, sigh :D

I might redo this to make the commits more readily reflect what is going on. It might also simply not be worth even trying to move the resource directory. I did go a little crazy.

Anyway. Will sleep on it and keep tinkering. Feedback appreciated :-)


SimonC4:
This is my feedback. 

I was able to get the Desktop version up and running, but not without pain caused by switching branches. There were a few leftover configuration settings in IntelliJ that caused issues, so I needed to do a git clean. I've added my detailed steps below.

### Desktop
These are the steps I followed to get the Desktop up and running. 
**Warning. Git Clean removes your local files.**
1. Close IntelliJ
2. checkout new branch
3. Git Clean using the command line: git clean -fdx
4. Prepare for IntelliJ using the command line: sh gradlew idea
5. Start IntelliJ
6. Open DestinationSol.ipr
7. Add new run configuration for Desktop: 
7.1 Run -> Edit Configurations
7.2 New Application
7.3 Name = Desktop
7.4 Main Class = org.DestinationSol.desktop.SolDesktop
7.5 Use class path of module = desktop
8. Run -> Run ‘Desktop’

### Android
Most of these steps are the ones listed in the readme. 
1. I made sure I had the same Android build tools as per the manifest: 20.0.0
2. I made sure I had the same SDK as per the manifest: 18
3. Make a local.properties in the project root with the path to my SDK. Example: sdk.dir=~/android-sdk/
4. To prepare in IntelliJ go to Project Structure and under Modules click the android "folder" icon and set the SDK
5. In the same window go to Artifacts and add a new Android Application, created from the android module
6. Go back to Modules -> android. Remove the Android sub-folder and add it back in again, as per this post: http://stackoverflow.com/questions/26910664/unexpected-node-android-packaging-in-android-gradle-build
7. I was then able to run the project from the menu: Run -> 'android'. It would build and prompt for the emulator and then error immediately. The error I received was:
Unexpected Error
Local path doesn't exist.
The project may need to be synced with Gradle files.

It looks like the apk has not been built. Not sure on the reason for this.

Cervator:
Huh, interesting on the Android bit, it sounds like you got most/all the steps. Could probably compare more exactly later when we get a chance. I don't hit step 6 for some reason. In the artifact config I set up code signing, but that's the only bit that sticks out.

Cervator:
On thinking about this some more I'll probably go back a few commits and looking at merging the three PRs first, then maybe sort out #17 before moving everything around.

Sort of lost in the shuffle was the point that testing LibGDX 1.6.1 on Android was a success, the project structure and automatic build of Android is just still somewhat on the drawing board :-)

Might even need to do that in the separate repo, akin to how alternate facades (and modules) in Terasology are in their own repos. Right now with this branch you need to have a working Android setup to even compile (unless that's not until you've finished adding more Androidy bits - which you could probably remove anyway). Not super friendly for new contributors.

I've done the crazy reorg thing before and with this as reference redoing it later shouldn't be too bad.

Still, need actual time for it instead of busy weeknights :-)

SimonC4:
That might be a good idea. That would allow this branch to be merged and allowing other changes to progress with the latest LibGDX libraries.

I did have some Android success though! I went back 2 commits and was able to get the Android app working by:
1. Copy the contents of folder main/res/ to android/assets/res/
2. Build from the command line: gradlew installDebug

I wasn't able to get it working through IntelliJ though. I also didn't add any code signing as I think for debug builds it's not necessary.

Having a reproducible build process for the Android would probably a pretty good start. Even if there is an ugly manual copy process involved.

Cervator:
Interesting! I was able to get Android to work earlier via IntelliJ, but not via Gradle :D

Got complaints about the platform tools directory not being correct, yet it clearly went to the right place. Weird

I'll probably sort this out pre-Android and push just the existing pieces (and the tweak needed for Android to compile separately with Box2D). Then can fix up the Android facade probably in the separated approach noted above. Dunno if the assets should move or not, want to look more at @LinusVanElswijk's PR and see how that might one day lead into optional modules.

vampcat:
@Cervator What's the status of this PR? Are we good to go, or is something remaining?

GooeyHub:

Cervator:
Closing this as obsolete, especially now with initial Gestalt integration merged :-)

The remaining potential worth of the PR boils down to restructuring some dirs to better fit Gradle/Android vs full adherence to LibGDX dir standards, I can play with that separately and re-PR if something of interest comes up.

Android is temporarily broken in develop until we figure out an Android compatible tweak to Gestalt (NIO not available on Android) sooo... there'll be a fair bit more tweaking to get all that working again. It'll be like the good ole "Oops we broke the applet again" days for Terasology :D

Issue #23, opened 2015-05-28
Repair Item Lost On Restart


SimonC4:
Each time you start a new game using your previous ship, you lose a repair item. 

#### Steps to reproduce
1. Have one repair item in your prevShip.ini file. An example file is below.
2. Start the game using your previous ship.
3. Remain idle for a few seconds.
4. Exit the game.  Use the exit buttons so the game is saved.
5. Inspect the prevShip.ini file.

#### Expected result
The repair item remains in the prevShip.ini file as it was not used.

#### Actual result
The repair item has been consumed and is no longer in prevShip.ini file.

#### Example prevShip.ini file
hull=imperialSmall
money=360
items=rep a4 s4 blaster waveGun sloMoCharge sloMoCharge sloMoCharge plasmaGun plasmaClip unShieldCharge missileClip missileClip missileClip missileClip missileClip missileLauncher knockBackCharge

PR #24, opened 2015-05-28
Fixes Issue #23 - Repair Item Lost on Restart

SimonC4:
The current repair behaviour works by consuming a repair item and converting it in to Repair Points. These Repair Points persist during your current game session and are consumed as needed when repairing your ship (when your ship is idle). 

The issue is that if your Repair Points were zero, as they would be at the start of a game, and a repair item is available in your inventory it would immediately get removed and converted to Repair Points. This would happen regardless of whether or not your ship needed repairing.

I've added a check at the start of the tryRepair function that will see if the ship needs repairing before attempting to repair it by consuming a repair item. I think this function gets called to repair all ships, not just the player's ship.

I've tested that repair items don't get consumed unless required and that the repair process still occurs when needed.


Cervator:
That's a fun little bug, thanks for finding it!

And there is stuff in place for non-player ships too? Interesting :-)

PR #25, opened 2015-06-01
Proper facades + lots of other PRs

### LINKED ISSUE #10: Bug:  Changing ship from Hunter to default ship: null reference exception
The exception occurred at the starting port.  I attempted to change to default ship and received the error in the attached screenshot.  I was able to reproduce this a second time with the same prevship.ini.  I"ve attached the screenshot and the prevship.ini containing the hunter ship configuration.

prevship.ini
hull=desertMedium
money=2104
items=bulletClip bulletClip bulletClip bulletClip bulletClip bulletClip bulletClip bulletClip bulletClip bulletClip bulletClip bulletClip bulletClip rep rep rep rep rep rep rep rep rep rep rep rep rep rep rep rep rep rep rep rep rep rep rep teleportCharge teleportCharge teleportCharge teleportCharge teleportCharge teleportCharge teleportCharge teleportCharge teleportCharge teleportCharge fixedGun a4 s3 fixedShotGun shellClip shellClip shellClip shellClip shellClip shellClip shellClip shellClip shellClip fixedGun plasmaGun plasmaClip plasmaGun s1 plasmaGun unShieldCharge s2 blaster 

huntertodefaultship-nullreferenceexception

Cervator:
Okay this approach is working pretty well. It doesn't do the big crazy restructure just yet as far as the paths to source code, package names, etc go. But it does properly hook up Android and even the GWT facade as separate Git repos you can fetch simply via Gradle: `gradlew fetchAndroid`

* Android facade: https://github.com/MovingBlocks/DestSolAndroid
* GWT facade: https://github.com/MovingBlocks/DestSolGwt

Haven't tried to automate the `res` copy into `assets` under `android` yet but if I do that and prep IntelliJ the Android version works on my phone just fine.

@SimonC4 - what did you do to get Android working via Gradle execution? I renamed the task so `gradlew android` will now attempt to run from the command line. I don't seem to have any devices set up so we need to add a quick bit to the instructions on how to do that. I did some initial readme updates.

Includes most other PRs: #19, #20, #21, and #24 so it should fix #10 and #23

Builds in Jenkins fine: http://jenkins.terasology.org/job/DestinationSolTest - since that builds without Android or GWT none of that stuff gets in the way.

*However* one problem remains: We've changed around the jars shipped in `/libs` and it looks like `sol.exe` isn't super happy for that. We may have to regenerate the exe. I tried renaming the plain GDX jar to its old name and got a Java error popup rather than nothing at all. But some of the other jars have been split out and we have more of them now.

After we figure that out I think this can finally be merged to master :-)

I'll leave around #22 just for review for when we do move more stuff around, but hopefully after merging / redoing #17


SimonC4:
I am able to run this on an Android emulator following these steps. This will install the Destination Sol app and can then be run from the device.
1. gradlew fetchAndroid
2. create local.properties file with sdk.dir=[sdk path]
3. Copy res into assets
4. Make sure emulator is running - or Android device is connection
5. gradlew android:installDebug

I quickly tried using IntelliJ, but was unsuccessful. Maybe list your steps as I think I'm missing something.


I'm curious about the distribution issue. It looks like all the jars are there. I was able to get it to run on Mac by using my own installed java as there is no Mac dist yet. I essentially did what was stated in this thread: http://steamcommunity.com/app/342980/discussions/1/611698195147641156/ I'd be curious to see if you can get that to run.

I'm assuming that sol.exe doesn't do anything special and would be the same as the linux equivalent, which is one line:
lwjreLinux/bin/java -jar libs/solDesktop.jar

I'll try it out on Windows when I get a chance.

SimonC4:
Quick update: I copied the DestinationSol.zip over to Windows and I was able to get it to work without any issues. Not sure what's different to your setup. I did do a git clean when I checked this branch out, to remove any leftover bits.

SimonC4:
Here's a link to the zip file if you want to try it out. 
https://www.dropbox.com/s/uip8cogwoa6z8qh/DestinationSol.zip?dl=0

Cervator:
Oh I can make it run with java -jar libs/solDesktop.jar just fine. It is only sol.exe that doesn't work for me. Does it work for you on Windows?

SimonC4:
Sorry, I should have been more specific, sol.exe does work for me on
Windows.

Cervator:
Thanks - huh that's weird. Was that the zip from Jenkins or one built locally? Not that it should matter. We might need to dig deeper but if more can confirm it as OK then at least we can merge.

Pinging @PrivateAlpha @theotherjay @LinusVanElswijk 

As for the Android setup in IntelliJ I put the rough steps I use in the readme, but can make it more specific when I get a chance :-) It is weirdly inconsistent - sometimes I get it working first try, other times even with a valid key store it refuses to accept it as signed. I haven't tried entirely unsigned yet though.

SimonC4:
That zip was built locally using "gradlew distZip". It was built on Mac, but I don't think it will matter.

When I get a a chance, I might create a thread post for the Android to remove it from this discussion and to get help working through the issues I'm having.

All dependent on having enough time though.

Cervator:
Yeah I hear you on the time bit - nearly 1:30 am here and I'm just getting to start on catching up on the day - or would, but I'm going to push it out and go to bed ;-)

I was able to test the .exe on my wife's PC and it worked fine. So I think that's a non-issue.

I'm ready to merge this unless you can think of any concerns :-) Then we can probably sort out the Android specifics later, weekend if needed for time zones. And yeah forum would be suitable too and might help invite others in to tinker.

SimonC4:
I think it's good to merge. 

I was able to do a clean checkout and following the instructions I was able to repeatedly run both the desktop and android app from both within IntelliJ and using gradle on the command line.

Some notes:
1. An alternative to running _gradlew idea_ is from selecting _Import Project_ from within IntelliJ. https://github.com/libgdx/libgdx/wiki/Gradle-and-Intellij-IDEA
2. I didn't need to do the steps of setting the android SDK or adding an artifact. 
3. You don't need to sign the app for local dev, only on releases.
4. I couldn't run using _gradlew android_ but installing using _gradlew android:installDebug_ does work. 
5. The recommended Android tools versions: http://libgdx.badlogicgames.com/versions.html

Cervator:
Done :-)

Yay for instructions!

The Android version stuff is pretty much arbitrary at this point. I spliced together what numbers I found in existing places with what I had handy locally.

We probably need to look at what the code base actually uses and how far back we want to support devices. API level 18 for instance is probably too new. But that takes some more thought and this week so far is bonkers :D

SimonC4:
API 18 is the target SDK, not the minimum SDK. Currently the minimum is set to API 4 from memory. It's in the android manifest XML.

http://developer.android.com/guide/topics/manifest/uses-sdk-element.html

Cervator:
Ah yep. Lots of numbers to get used to :-)

SimonC4:
Indeed. Nice job with all the Pull Requests. :-)

Cervator:
Not to get into a bad habit to comment on closed PRs but since it came up here I went through the Android setup a little more and tried `gradlew android:installDebug` - which worked, but just said it installed the app on my device, it didn't actually do anything.

I noticed the package (rather than the target class) in Gradle was still set to a placeholder LibGDX value and fixed it, now `gradlew android` works for me with my device connected, huzzah!

I dunno if that was in part because I did the installDebug first, but hey. It works :-)

PR #26, opened 2015-06-03
Configurable keys

SimonC4:
I've added key mapping in to the settings.ini file which will appear after you next run the game. The key mappings can be changed by using a text editor. The definitive list of possible key definitions can be found here: https://github.com/libgdx/libgdx/blob/02ccad70e85c615962e0fa0932ab01d3da5feba1/gdx/src/com/badlogic/gdx/Input.java#L247

The goal of this work was to remove the hardcoded keys and replacing them with configurable settings. There should be no other functionality change, however:
1. The settings.ini file is always saved on exit. It only used to be saved when settings were changed. This is to make the new settings show in the file immediately. The config settings aren't saved on mobile.
2. Fixed a bug in the tutorial where the instruction said "left" when it should have been "right"
3. There was a "hack" where in KB + Mouse setup the number keys of 5, 6, and 7 were used for shoot primary, shoot secondary, and ability. I replaced the hack with the normal keys. The consequence is that the number keys will no longer work and in KB + Mouse mode you can now use "Space" and "L-Shift" to shoot.
4. There was also another, non-apparent, bug where the Escape key was mapped to several menu options on one menu screen. It's now only mapped to the close button. 

Ideally the key mappings should be a setting within the options menu, but I think that can come at a later date.


Cervator:
Tested this out: very cool! I saw the left/right tutorial thing and each time tried to remember that I should find and fix that real quick - naturally never remembered :D

Mouse + KB feels much more natural now

I think @PrivateAlpha had started looking at what would be needed to make an input mappings GUI screen but yeah agreed that could be done separately, just wanted an excuse to ping him :-)

Cervator:
Tested again and merged :-)

The build out of Jenkins works fine, other than the previously mentioned weirdness of sol.exe not working on my PC (but fine on my wife's).

I added the new key binding defaults to the settings.ini in source control just so we have them handy. That has no impact on the build game though - it starts fresh.

Thanks again! With this in place next Steam release we can just tell people to go tweak their keys in settings.ini :-)

PR #27, opened 2015-06-03
Merge Master in to Hull

SimonC4:
Merged Master in to Hull and resolved the conflicts caused by PR #20 and PR #24.


SimonC4:
Fixes issue where the hull's approximate radius wasn't getting set. This caused the assertion error and it appears it was the cause of why the AI ships behaved oddly.

Cervator:
Thanks for tweaking at this! It does look a little better :-)

Still seems to be a bit to go, the pirate towers still topple over:

destination sol_2015-06-04_00-48-45

And some AI ships still like taking naps:

destination sol_2015-06-04_00-49-11

Another that can provoke one of the new quirks is hopping into the heavy fighter you can buy at the starting station (the one where one wing is a kind of stump). Buy that and just bump around the starting station / ships there, the ship boundary doesn't seem right.

Finally mercenaries are really whacky with this, hadn't noticed that before :D Not that they're super complete and helpful even normally! Just fly to the first planet and buy a pile of the mercs there, see it rain ships!

On a Git related note it looks like something went wrong with the merge. You've got the commits you and I have been working on lately showing up as duplicates here, and @LinusVanElswijk's original commits consumed inside the merge rather than shown normally. 

Could be the order of things. Try a fresh version of the master branch from MovingBlocks then make sure to merge hull into *it* rather than the other way around, if that's what happened :-)

SimonC4:
Ok, I'm still pretty new to git. My intention of this PR was to get master merged in to the hull branch so the hull branch was up to date to continue working on it. I didn't want to merge hull in to master as there are still some issues with it. So should I still branch master and merge hull in to that branch?

As for the duplicate comments are you referring to this commit? https://github.com/SimonC4/DestinationSol/commit/93c6ec329d5091697b48bbc441fd5ec377ae2841. If so, SourceTree automatically adds that as the commit message and I didn't change it. I'll make sure to remove them in future.

Nice job pointing out those issues. I'll try and have a look at them.

Cervator:
Yeah Git takes a while to get used to :-)

A better way to simply update a branch is using rebase. We'd rebase hull on top of master, moving all the commits in it forward so they're placed in front of the head commit. Rebase can be scary too though, I still sometimes copy my whole workspace just in case :D

SourceTree? The commit message is okay, generally merge commits just get "Merge x into y" and the merged commits themselves hold further descriptions. But if you have a tool that does it automatically that in itself shouldn't hurt anything.

I'm realizing now though that I was confused at your goal. This PR makes sense if you're trying to update hull rather than master. In this case sure we'd see our commits instead of those from @LinusVanElswijk. That's on me :-)

Generally I try to avoid merges to anything other than the primary branch, as having merges in multiple places can cause commits to be duplicated (as they end up appearing in different places in history). But that isn't technically the case here after all as the hull branch hasn't seen these commits yet.

Rebase is the ideal choice here, but the conflicts may complicate it. I'd have to check and see.

Sorry about the confusion!

Cervator:
Tried the rebase and yeah it doesn't look too bad :-)

Got too late tonight though, too sleepy to do it right :D

SimonC4:
Ok cool. As discussed I'll close this pull request.

PR #28, opened 2015-06-07
Hull branch rebased with a quick loot drop fix on top

Cervator:
This replaces #17 and #27 with a rebase of the `hull` branch on top of latest master

There were conflicts with #20 and #24 that I believe are resolved - I tried first to resolve them during the rebase, but then there was a little more to do at the end. Bit confused about gun mounts and guns sometimes being off by 1.

Tests out okay with the existing known issue where some ships don't quite behave their size leading to some confused AI among other things like being able to poke into structures.

At least one mercenary ship type is broken - on the first planet I ran into a blank mercenary entry that if hired produced very directionally challenged ships :dizzy: However on another planet I hired some bomber fellows that actually were able to follow me decently in formation. Well, occasionally. Most the time (as before this PR) they would overcompensate quite a bit and zoom by me at high speed only to reverse and try again :-)

I found an issue where the SolMath utility random method would cause an assertion error crash just because an `myHull.config.getApproxRadius()` call returned 0. The math class rng with max + min parameters would crash the game if they were the same rather than just log and continue. I added some logging including a stack dump (but no crash / thrown exception) so we can track down any other issues like this - I *think* I might have seen that elsewhere in the past and it is a pretty severe penalty for feeding two like numbers to an RNG.

Maybe some of the new jsons are not setting something right? Could that explain the broken mercenary and how just *some* of the ships/structures are behaving weirdly?

@SimonC4 check out the network graph now for a good illustration of rebase :-) In IntelliJ it was pretty easy, the two conflicts would stop the rebase at a particular commit, ask me to resolve the conflict same as if I was merging something, then I could continue the rebase until done. I just didn't quite resolve the conflicts correctly/fully the first time through.


SimonC4:
Nice work with the rebase. The graph shows it well. 

I've submitted PR #29 as a fix for the approxRadius. 

The issues pointed out in PR #27 still exist and need to be resolved.

Cervator:
Doh, I completely missed your extra fix in there. Clear as day :imp:

SimonC4:
With PR #30 merged, as far as I know the only remaining issues are on the planets where the turrets fall over and some ships don't move. Although I haven't done much testing. Do you know of any others?

Cervator:
At least one of the mercenaries is broken, check out the station at the first planet. It probably isn't loading right - no name, acts funny, etc.

Wondering if we should just prep up for a release before worrying about stabilizing this one. Good fixes already :-)

SimonC4:
@Cervator, do you want to look at rebasing this branch again? The configurable keys might cause a conflict.

Cervator:
Done! Closing in favor of #41

PR #29, opened 2015-06-07
Fix for approxRadius. 

SimonC4:
Pull request for the hullrebased branch. This fixes the approxRadius being set to zero. The multiplication of size by 0.4f was lost during the refactor: https://github.com/MovingBlocks/DestinationSol/blob/852dd290cb2e9ba1f52be31f1550be34396976ac/main/src/com/miloshpetrov/sol2/game/ship/HullConfig.java#L53


Cervator:
Tested it out real quick, the poor guards at the starting station finally stopped spinning and were able to just hang out and chill. Huzzah! And no more log spam over the previous crash when loot drops.

Thanks :-)

PR #30, opened 2015-06-08
Hull - Origin Fix

SimonC4:
Pull request for the hullrebased branch.

This fixes the issue where the ship boundaries were not quite right as observed by using the imperialMedium "Fighter" ship and bouncing in to things.

The cause was the origin value within hullConfig was not being set correctly.

### DIFF
--- a/main/src/com/miloshpetrov/sol2/files/HullConfigManager.java
+++ b/main/src/com/miloshpetrov/sol2/files/HullConfigManager.java
@@ -123,7 +123,7 @@ public final class HullConfigManager {
     }
 
     private void parseGunSlotList(JsonValue containerNode, HullConfig.Data configData) {
-        Vector2 builderOrigin = shipBuilder.getOrigin(configData.internalName);
+        Vector2 builderOrigin = new Vector2(configData.shipBuilderOrigin);
 
         for(JsonValue gunSlotNode: containerNode) {
             Vector2 position = readVector2(gunSlotNode, "position", null);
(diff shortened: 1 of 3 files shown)


Cervator:
Yay more fixes! Thanks :-)

By the way you might want to check your local Git config. Does the email configured with it match the email on your GitHub account? Usually if that's the case you get your clickable username on the commit entry, rather than a generic Octocat + no clicky. Can always add the Git email to your account on GitHub, should fix it.

SimonC4:
Thanks for pointing that out. I think that must have happened when I updated SourceTree. I've fixed it now, but it won't change previous commits.

Cervator:
As long as you still had an email on the commits (even if it is wrong) I think you might still be able to add it to your GitHub account as a secondary. But no big deal in any case :-)

SimonC4:
Unfortunately not, there was no email.

SimonC4:
Hmm, looks like there was a default email added. I've added this to my GitHub account, but it doesn't look like it's made a difference.

Cervator:
Oh well, keep an eye on future commits hopefully behaving, at least :-)

Checked this out, the fighter does feel better. Am realizing even before the hull branch there is plenty of overlap with different objects, either imperial ship is able to interact funny with the main station, asteroids, etc. Like their bounding box isn't quite right in the first place. So we need to be wary over attempting to fix all the things when some were broken to begin with :-)

PR #31, opened 2015-06-09
Resolution screen show supported resolutions

SimonC4:
This had been bugging me. The resolution screen showed 3 hardcoded resolutions regardless of whether they were supported or not. This would cause crashes on my machine if selected incorrectly and is the cause of other user's issues as reported on Steam.

The solution is to cycle through supported resolutions as provided by libGDX. This isn't the most elegant as some of the resolutions are very small, but from my testing on Mac and Windows 7, all of the supplied resolutions work in windows or full screen, so at least it will prevent crashes. 

I also made another change that after you leave the resolution screen, the new settings are applied immediately without having to restart the game. 

It's not perfect and could do with some refinements, but I think it's better than what currently exists. Would be keen for someone else to do some testing on it.


Cervator:
Awesome! Yeah I had been wondering about that too and it is a great fix for users.

Couple easy things come to mind:

* Text still says to restart the game to apply changes :D Maybe replace with change applies on clicking "Back" - or if not terribly difficult maybe split button into Apply | Cancel?
* You can end up with a whole lot of duplicate resolutions - easy suggested fix follows!

A quick bit of debug logging showed I have a ton of resolution options because of variants of refresh rates and bpp. I have 16 versions of 1680x1050

```
    Gdx.app.log("GameOptions", "Listing available display modes");

    for (Graphics.DisplayMode d : displayModes) {
      Gdx.app.log("GameOptions", "Display mode: " + d);
    }
```

Rather than go through the displayModes array when cycling how about loading it once then filtering out all the dupes? Then cycle through that collection. Although I'm not sure how we'd pick the favored variant. First come first serve? We can do something fancier when we're more on top of changing the UI. With the right collection you can probably just toggle it for the next variant rather than worry about an index.

Finally, does this change anything on Android? I was trying to test but crash either via IntelliJ run config or `gradlew android` - probably that's just my setup.

SimonC4:
Cheers, good feedback. I'll look in to it and update the PR. 

Without the duplicates, how many options do you have? Is it manageable to cycle through them or are there too many?

This shouldn't change anything on Android as the settings screen doesn't get used. But I'll double check it to make sure.

Cervator:
Probably manageable - *just* 20 ;-)

Some quick Set action:

```
GameOptions: Unique: 600x800
GameOptions: Unique: 1200x1920
GameOptions: Unique: 576x720
GameOptions: Unique: 1200x1600
GameOptions: Unique: 480x720
GameOptions: Unique: 1024x1280
GameOptions: Unique: 1000x1776
GameOptions: Unique: 480x640
GameOptions: Unique: 768x1360
GameOptions: Unique: 768x1280
GameOptions: Unique: 1050x1680
GameOptions: Unique: 1080x1920
GameOptions: Unique: 720x1280
GameOptions: Unique: 960x1280
GameOptions: Unique: 768x1024
GameOptions: Unique: 768x1366
GameOptions: Unique: 900x1440
GameOptions: Unique: 800x1280
GameOptions: Unique: 648x1152
GameOptions: Unique: 864x1152
```

Cervator:
FYI: There is a very minor merge conflict now with #26 in

PR #32, opened 2015-06-11
Update Windows JRE to 32-bit version of 1.8.0_45.

SimonC4:
Hopefully this will mean we can use the build from Jenkins on Windows machines and I can update the Steam build and have a working Beta for Windows.

I wasn't able to use the Jenkins JRE build as javaw.exe is missing. 

I don't like having this stuff in the main repo, but this gets the build back up and working. A future task can be to put the launcher and Steam build process in to a different repo.

I also looked at Linux and found a can of worms. It doesn't look like the existing Linux build on Steam works as it's missing stuff in the JRE. I managed to get it working on my Linux machine but will look in to that more later.


Cervator:
Cool, yep works fine for me now :-)

Yep javaw.exe isn't in there on purpose since the custom .exe we came up with doesn't use it. But it isn't super finished yet so oh well.

Agreed on fixing up stuff more later.

PR #35, opened 2015-06-18
Linux JRE

SimonC4:
The 32-bit JRE wasn't working on 64-bit Linux. Have added the JRE 1.8.0_45 for both 32-bit and 64-bit versions of Linux. Also added a new bash file for starting on 64-bit Linux. Removed the optional files as listed here: http://www.oracle.com/technetwork/java/javase/jre-8-readme-2095710.html

In Steam, the 32-bit and 64-bit versions can be set up separately, so that will be the next step for Linux.

Disclaimer note: I dislike this being in the main repo, but it will do for now.


Cervator:
Yeah with the big files in the repo it bloats a bit and unless we reset history you'll always get hit with the size impact on the initial clone.

While it is still relatively early we could potentially do a reset of sorts when we have the JREs sorted out and a better option on where to keep them (separate repo, maybe shared with Terasology). The few active people would need to redo branches but that's relatively minor at this point, much worse with something like the 4-year history in MovingBlocks/Terasology :-)

PR #36, opened 2015-06-19
Steam scripts

SimonC4:
Adding the Steam build scripts to the repo as there have been a few changes for Linux. There's nothing confidential within these build scripts.

### DIFF
--- a/steam/ContentBuilder/scripts/app_build_342980.vdf
+++ b/steam/ContentBuilder/scripts/app_build_342980.vdf
@@ -1,0 +1,17 @@
+"appbuild"
+{
+	"appid"	"342980"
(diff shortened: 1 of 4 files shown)


Cervator:
Thanks for these :-)

I went ahead and updated the resolution PR (with a new one) for you to look at.

After we merge that want to throw together another Beta build for Steam that we'll publish the key for on the forum on both Steam and our main site? Try to get a few other testers just to make sure all is well.

If all looks good we could try to release next weekend. In the meantime I'll try to find some time to look at how to publish a new build on Google Play (oh, double-checked latest code including resolution changes - no impact as expected, works as usual)

We should get some basic patch notes too.

PR #37, opened 2015-06-22
Resolution changes - round two

Cervator:
Builds on  #31 and makes it a little fancier.

Pretty sure this is still fairly primitive but then I usually organize more than actually code - it works though :D

@SimonC4 what do you think? Can you give it a test and confirm all is well?

It now sets up a SortedSet once then iterates over it, restarting when it gets to the end. Resolutions come up sorted, although they start with the first after 1000, good ole String sorting (so yeah, could totally be even more shiny)

One quirk is that after restarting the game the next resolution to come up doesn't care what you had saved. I tried to initialize earlier, even wanted to try picking a sane default, but GameOptions' constructor gets called before we have graphics. And I'm just trying a quick fix :-)

Also highlights a difference between rebases and merges to keep a branch up to date - merge commits get pulled in. Hardly a big issue, just a nitpick from being an organization nut :D


SimonC4:
Looks good. Works fine on my Mac. I'd be happy that this is merged.

Did you fix the issue where duplicate resolutions where showing up on your Windows machine?

Cervator:
Yep and merged :-)

Probably ready for our first release candidate!

PR #39, opened 2015-07-01
Added license comments

Halamix2:
Now all .java files have license info

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@
  * - View tasks for this project: $ gradlew tasks
  */
  
- buildscript {
+buildscript {
     repositories {
         mavenCentral()
     }
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,27 @@
-buildscript {
+/*
+ * Copyright 2015 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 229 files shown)


Halamix2:
I forgot about "  package com.miloshpetrov.sol2.game.gun;" (in two files) but I already made two commits so this could be added to another commit maybe?

Cervator:
Thanks for joining the good fight :-)

You don't need to worry about license headers in .gradle files, we generally leave those header-less (two are included here).

Rebase will make quick work of additional commits to catch another couple files, if you want just squash them back into a single commit. I'm not very worried about 2-3 commits though.

However, one tricky bit - I was hoping to get to these kinds of big refactoring efforts done after we fix and merge the hull_rebased branch (#28). It has changes to a *ton* of Java files too meaning we'll hit massive conflicts.

Do you have a nice way to apply this change in bulk or were you copy pasting the header into individual files? I think with the right kind of IDE setup it may be possible to bulk-format the codebase (not sure though)

Cervator:
Big hull branch has been merged into a new develop branch now. Found a nicer way to re-apply this without piles of copy paste? :-)

We can cram this in there if easy enough to do sometime. There will be more mass-changes and refactoring later.

Cervator:
Closing as superseded by #74

PR #46, opened 2015-07-10
Fixes issue with loading textures

SimonC4:
As reported here: http://forum.terasology.org/threads/version-1-2-0-hull-restructuring-needs-testing.1342/

@Cervator, I'm submitting this PR so if you have the time you can review and possibly merge to try out the new build from Jenkins. As mentioned on the forums, the icon.png files are just placeholders at this stage and need to be replaced, but that shouldn't be too hard.


SimonC4:
I've updated the icon.png images to the correct images and also fixed an issue where the distZip wouldn't start. The code should probably be tidied up a bit, but it should now be in a working state.

Cervator:
Tests out well, including as binary from Jenkins. Woo! Thanks :-)

Ran into an Android issue (#47) but I suspect that's from the hull refactoring

PR #49, opened 2015-07-22
Controller and Input Mapping

SimonC4:
Added partial controller support. This allows the user to control the ship using a controller. Menu tasks still need to use either the keyboard or mouse. 

Added a new screen to the options menu to configure input mapping. This was needed to map controller inputs and have also added mapping keyboard inputs.

I don't think the controllers library needs to be added to the android gradle script as it should never get invoked when running on Android, but could do with testing. 

I might change the default controller mapping too. Works well with Xbox 360 on Mac, but inputs are different on Windows. Will do some more testing around this and with a different controller.


Cervator:
Very cool! I have a controller laying around, I forget exact which one (probably Xbox or PS, one generation behind or so). USB. I'll try to test it out this coming weekend :-)

Cervator:
Tested out! Posting here just for later reference if not fixable right away: hit a crash on opening the tutorial after having set my controller mappings

```
3:59:19 PM - Cervator: Controllers: 1
3:59:19 PM - Cervator: #0: Afterglow controller for PS3
3:59:19 PM - Cervator: java.lang.ClassCastException: com.miloshpetrov.sol2.game.screens.ShipControllerControl cannot be cast to com.miloshpetrov.sol2.game.screens.ShipKbControl
3:59:19 PM - Cervator: 	at com.miloshpetrov.sol2.ui.TutorialManager.<init>(TutorialManager.java:40)
```

PR #50, opened 2015-07-23
libGDX Update to 1.6.4

SimonC4:
Updated libGDX to the latest version of 1.6.4. Only minor changes, change log here: http://libgdx.badlogicgames.com/news.html

Removed some unused libGDX extensions from the gradle script, but left RoboVM for possible future iOS dev.

Also bumped the version of Destination Sol to 1.2.1

Tested on desktop and android.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,14 +15,11 @@ allprojects {
     apply plugin: "eclipse"
     apply plugin: "idea"
 
-    version = '1.2.0'
+    version = '1.2.1'
     ext {
         appName = 'DestinationSol'
-        gdxVersion = '1.6.1'
-        roboVMVersion = '1.2.0'
-        box2DLightsVersion = '1.3'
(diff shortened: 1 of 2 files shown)


Cervator:
Cool, hunting any particular upgrade or just to keep up with the latest version?

With the Android issue for me with v1.2.0 being a local problem maybe we should release it this weekend and at the same time take this and #49 to make a v1.3.0 beta? Looking forward to getting a chance to play with a controller. Also getting a sneaky idea about split-screen multiplayer on OUYA ... no networking, but still hardly a small change ;-)

SimonC4:
Mostly keeping up to date, but there's a small chance I might look in to iOS and RoboVM 1.5 has dev tools that can be used on Mac.

Yep, I'd be happy if these become a 1.3 release, assuming no issues with #49.

PR #58, opened 2015-07-27
Added some documentation to class ChunkFiller, fixed a bug in getFreeRndPos...

grauerkoala:
… where the returned position could be outside the chunk.

The documentation is, of course, by no means complete, but it's a start.

### DIFF
--- a/main/src/com/miloshpetrov/sol2/game/chunk/ChunkFiller.java
+++ b/main/src/com/miloshpetrov/sol2/game/chunk/ChunkFiller.java
@@ -45,30 +45,40 @@ public class ChunkFiller {
     myDustTex = textureManager.getTex("deco/space/dust", null);
   }
 
-
+  /**
+   * Fill the background of a given chunk with floating junk.
+   *
+   * @param game    The {@link SolGame} instance to work with
+   * @param chunk   The coordinates of the chunk
+   * @param remover
+   * @param farBg   Determines which of the background layers should be filled. <code>true</code> fills the layers furthest away, <code>false</code> fills the closer one.
+   */
   public void fill(SolGame game, Vector2 chunk, RemoveController remover, boolean farBg) {
     if (DebugOptions.NO_OBJS) return;
 
+    // Determine the center of the chunk by multiplying the chunk coordinates with the chunk size and adding half a chunk's size
     Vector2 chCenter = new Vector2(chunk);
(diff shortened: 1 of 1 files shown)


main/src/com/miloshpetrov/sol2/game/chunk/ChunkFiller.java
```diff
@@ -163,83 +172,162 @@ private void fillAsteroids(SolGame game, RemoveController remover, boolean forBe
...
+   * @param g        The {@link SolGame} instance to work with
+   * @param chCenter The center of a chunk in which a random position should be found
+   * @return A random, unoccupied position in a chunk centered around chCenter, relative to the entire map, or <code>null</code> if within 100 tries no unoccupied position has been found
+   */
   private Vector2 getFreeRndPos(SolGame g, Vector2 chCenter) {
-    Vector2 pos = new Vector2(chCenter);
     for (int i = 0; i < 100; i++) {
-      pos.x += SolMath.rnd(Const.CHUNK_SIZE/2);
-      pos.y += SolMath.rnd(Const.CHUNK_SIZE/2);
+      Vector2 pos = getRndPos(new Vector2(chCenter));
```
SimonC4: Does this change work as expected? Previously it was adding a random amount to x and y positions every iteration in the for loop. After the change you are getting a new vector position every iteration.

grauerkoala: This works differently, but on purpose. The way the method was originally designed, a position outside of the chunk whose center chCenter represents could be chosen; for example, if the first value hits exactly (-1)*Const.CHUNK_SIZE/2, but an object is already occupying that spot, a new random value is going to be added on top of that. If that value is negative again, the position will end up being outside the chunk. I'm fairly confident that this is unintended behaviour, which is why I changed it in such a manner that it now always chooses a random position based on the center of the chunk and not some random position therein.

SimonC4: Ok, sounds good and I'm glad it was intentional :+1:

grauerkoala: I'm glad you're keeping an eye out nonetheless. :)

Cervator:
Tested around a bit, seems we still have plenty of junk, dust, and debris! Merged :-)

Issue #61, opened 2015-07-29
Mouse Off Centre After Changing Resolution


SimonC4:
I'm fairly certain these is the issue reported on Steam: http://steamcommunity.com/app/342980/discussions/1/535151589908633094/

Steps to reproduce
1. Start the game with the settings that have: resolution 800 x 600, windowed, and with mouse control. These are the default settings.
2. Change resolution to 1366 x 768 (or similar widescreen aspect ratio)
3. Start a new game
4. Move the mouse pointer to the right of the ship

Expected result:
The ship turns to the right

Actual result:
The ship doesn't turn to the right until the mouse is very far to the right.

SimonC4:
Fixed in #62.

Cervator:
Ah, hadn't noticed this didn't auto-close. I think it would have with a commit message like "Fixes #61" - the "issue" is probably throwing off GitHub's detection.

PR #62, opened 2015-07-29
Fixes issue #61

SimonC4:
The screen UI uses a ratio to determine where the items appear on the screen. The x and y coordinates of the screen are converted to a float, where the top is y=0 and the bottom is y=1. The left is x=0 and the right is x=ratio (width / height), e.g. in 800x600 the right hand side is 1.33.

The ratio that the UI elements is used is initialised when the game starts, so once the screen resolution is changed to something like 1366x768, ratio of 1.78, the UI elements are not reinitialised. This isn't too much of a problem, but you will observe some oddities, like the text "Controls: KB + Mouse" don't fit in to the button, but after a restart the will.

The possible options are:
1. Restart the game, but there's no easy way other than asking the user. 
2. Reinitialise everything that uses this ratio - which is every UI element. This seems like it's essentially restarting the game while the game is running. 
3. Scale the mouse pointer by the same ratio so that it's in the correct place.

I've gone with option 3 which scales the mouse x pointer by the ratio of the initial ratio over the current ratio. I've written up steps to reproduce the issue in #61.

### DIFF
--- a/main/src/com/miloshpetrov/sol2/ui/SolInputManager.java
+++ b/main/src/com/miloshpetrov/sol2/ui/SolInputManager.java
@@ -23,6 +23,7 @@ public class SolInputManager {
   private static final float CURSOR_SHOW_TIME = 3;
   public static final float CURSOR_SZ = .07f;
   public static final float WARN_PERC_GROWTH_TIME = 1f;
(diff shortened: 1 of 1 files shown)


Cervator:
Nice. Will do another round of testing when I get home from work :-)

Cervator:
Was able to replicate exactly and confirm the fix. Nicely done good sir! :-)

PR #63, opened 2015-07-31
Fix #53

### LINKED ISSUE #53: Can map duplicate keys in the new input map
So `space` could be bound to both fire and secondary fire, for instance.

SimonC4:
Key or controller button can't be mapped to more than one control. There's no restriction on controller axis inputs as the same axis can be both up and down or left and right. This could be implemented, but I think it's fine as it is.


Cervator:
Tested and merged :-) Agreed on leaving controller input more open.

Issue #67, opened 2015-08-02
Better option to free mouse from windowed game


Cervator:
Currently the mouse is captured a bit too thoroughly when in windowed mode, especially in some cases like http://steamcommunity.com/app/342980/discussions/1/535150948597573921 (which may be helped by recent fixes and impacted by resolution issues

Some related discussion in #48 and some work toward a fix.

Cervator:
Fixed

PR #69, opened 2015-08-04
Mouse Freedom

SimonC4:
This no longer confines the mouse to the window. This was an annoying issue where the mouse could never leave the game when in windowed mode. The mouse is still confined when playing the game with the mouse but is unconfined when in the in-game menu.

I think this is a much better and simpler solution than changing the way the mouse cursor is drawn, which is what I was attempting in my other branch.


GooeyHub:
Refer to this link for build results (access rights to CI server needed): 
http://jenkins.terasology.org/job/DestinationSolPRs/2/

SimonC4:
I have noticed that the mouse can jump to a different position when opening the in-game menu when the input is "keyboard and mouse". Not sure what can be done about this. I think it is acceptable as the overall situation is much better.

SimonC4:
Hadn't realised there was a bug raised. This is a fix for #67.

Cervator:
freedom

Cool stuff. Will check it out after work :-)

Cervator:
Tests out fine. And yeah I wouldn't worry about anything that small, I couldn't even notice it :-)

PR #101, opened 2016-07-17
V1.4 Final tweaks

Cervator:
Hopefully, anyway :-)

I found a rather critical bug in the last round of equipping logic changes - NPC ships were no longer equipping stuff, whoops! Wouldn't I feel silly if I was the supposed tester who missed that. Which I was so... yeah. Nice going me!

Took me an odd amount of time to figure out, probably in part from being out of practice and from the equipping logic being shared a bit funny between player, NPCs, etc. Mercs too I suspect but something still seems to keep them from firing :/ 

They were already borderline broken anyway, so forging onward. Fix for now was simply using the new logic only if the "pilot" being processed is the player, if not use the old logic.

Found and fixed one related bug where a later format change hadn't been applied to the player's default starting config, resulting in nothing being equipped. And during my long journey I rarely encountered a crash bug (is it a crash bug if the game commits suicide when it detects a state it doesn't like?) which it seems the game can actually bypass with no apparent ill effect.

Hopefully.

On a completely different note recent Gradle changes had left the Android part broken as well, which can happen easily since we still haven't got it building automatically on commit / for testing new PRs. Hey I wonder whose job that's supposed to be, hmm.

TODO when merging:

- [ ] Also merge the tweak on the Android side
- [ ] Add the logo change on top if ready
- [ ] Version bump for great justice (non-snapshot)
- [ ] Release v1.4.0 already!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,23 +13,8 @@ buildscript {
 import org.ajoberstar.gradle.git.tasks.*
 
 allprojects {
-    apply plugin: 'java'
     apply plugin: 'eclipse'
     apply plugin: 'idea'
-    apply plugin: 'project-report'
-    apply plugin: 'checkstyle'
-    apply plugin: 'pmd'
-    apply plugin: 'findbugs'
-
-    // Computes code coverage of (unit) tests
-    apply plugin: 'jacoco'
-
-    dependencies {
-        checkstyle('com.puppycrawl.tools:checkstyle:6.5')
-        pmd('net.sourceforge.pmd:pmd-core:5.3.3')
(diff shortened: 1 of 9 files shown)


GooeyHub:

Cervator:
More Android tweaks applied in the other repo. Played for a good half hour via beta release on Google Play. Looking pretty good!

Also got smushed on the surface twice, first by an oversized meteor then by the traveling marketplace ship crashing into the planet. That one actually made me laugh out loud :-)

Checking Steam build next then done at long last.

PR #108, opened 2016-08-21
Possible fix for #102

### LINKED ISSUE #102: Game lag causes audio problems
I'm not sure if this is new with the soundtrack but it may be easier to detect now that there's more audio active.

Reported at http://steamcommunity.com/app/342980/discussions/1/352788552267722914/ specifically with the music going crazy on a Windows 10 system, happening some minutes into the game in laggy situations. Music playing in the menu is fine.

I experienced something similar while playing on Android and an enemy ship seemingly got stuck under the planet surface, yet it was still very eager to shoot me. Being that it was up against a wall the sfx from firing/landing hits were playing excessively and my phone nearly locked up from the lag with audio again sounding terrible due to the lag.

I wonder if there is something we can do to either limit audio being played if too much triggers or somehow isolate the playing somehow (background thread?) to where game lag won't kill it.

@SimonC4 and @Rulasmur - ping! Since the two of you worked with the MusicManager :-)

JGelfand:
[As recommended by](https://github.com/MovingBlocks/DestinationSol/issues/102#issuecomment-233798624) @SimonC4 and @Cervator, the MusicManager now disposes of music when it stops playing. This may fix #102, but I'm not sure since I've never had the game lag on me to begin with.

### DIFF
--- a/main/src/org/destinationsol/game/sound/MusicManager.java
+++ b/main/src/org/destinationsol/game/sound/MusicManager.java
@@ -18,6 +18,8 @@ package org.destinationsol.game.sound;
 
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.audio.Music;
+import com.badlogic.gdx.files.FileHandle;
+
 import org.destinationsol.GameOptions;
 import org.destinationsol.files.FileManager;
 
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Thanks again! I tested this out on both PC and Android and it seems to work fine. I cannot reproduce it on PC either but I do think I felt it on Android once. Didn't this time but then I'm not sure I reenacted the same events. I'll ping the thread on Steam where the issue was originally reported :-)

Please keep it up! Activity promotes activity, we've got several idle contributors that probably just drifted off from the project being a bit quiet for a while. If you run out of interesting stuff to do let me know, maybe introduce yourself in our dev forum sometime if you like (gets you a contributor badge if you're the collector type)

These were two targets I was hoping to get fixed for a v1.4.1 patch release. The third one was trying to fix mercs equipping items, interested in checking on that? There isn't an issue for it yet but I mention some of the related details in #101, which also touches some of the related files.

I am unsure if it is better to try to fix the equip step or making sure somehow that the mercs are at least born with a weapon equipped. They do not equip a weapon dropped on them right now, which used to work.

@GooeyHub: add to whitelist

GooeyHub:

Issue #113, opened 2016-08-28
Game crash on attempted start


JGelfand:
I played a bit on the $400 ship with one heavy slot, then bought a new ship and realized I should switch to mouse+kb to make handling multiple guns easier. I exited the game, then switched back and got the [following error](http://pastebin.com/g1gVHz1S). The fancy new crash reporter prompted me to file an issue on github, so here we are.

JGelfand:
Oh hey! I tried starting the game again, and got the same error. I guess I somehow broke my savefile. The "reached group size limit" thing makes me think it might have something to do with the way I maxed my bullets for my two heavy machine guns before exiting. I'll try and replicate this tomorrow, I guess.

PR #114, opened 2016-08-28
Handle errors cause by item overflow gracefully

JGelfand:
This prevents the crash on startup I mentioned in #113.


GooeyHub:

Cervator:
Sure does!

Was digging through my notifications and saw your report, figured it might be another one of those odd overly aggressive AssertionErrors, and yep, sure enough kept scrolling and saw you already found and fixed it :D

There are several more spots where the game exits like that when it could just handle the situation gracefully and continue. We fix them as we come across them, so feel free to keep an eye out :-)

Issue #100, opened 2016-06-05
Split apart menu background + logo, update to new logo


Cervator:
TrekMarvel provided a set of nice new logos in the forum: http://forum.terasology.org/threads/a-brand-new-logo.1545

Trouble is without a pure black background any logo may look funny / clash with the black background. We should likely split apart the background from the logo, set the background image to fill the whole background of the menu, then add the logo as an overlay above the buttons.

I've tried updating the Android logos and such, that seems to work OK. May not have found the spot determining the icon for the windows on PC yet.

SimonC4:
I think the Windows executable used Launch4j, which is a tool for wrapping Java applications distributed as jars in lightweight Windows native executables.

More information on Launch4j: http://launch4j.sourceforge.net/
This Gradle plugin might be useful: https://github.com/TheBoegl/gradle-launch4j

Cervator:
Oh that's right. Thanks :-)

I *think* I was even seeing the old icon on the game launched via source, so may still need to hunt that piece. Icon on Android in its task manager was updated, so that was cool to see.

crazywolf132:
The game will work fine without a pure black background. I did some testing and it is fine.

crazywolf132:
Working on PR now.

vampcat:
Closing, as completed by #145 .

PR #107, opened 2016-08-20
Separate volume buttons for music, other sounds

JGelfand:
I made separate volume buttons for the music and other sound effects, per issue #97 .

PR #116, opened 2016-09-16
Logo and background. #100

crazywolf132:
Added the logo and background I even added the desktop icon, feel free to change it to the other as i didn't know which one to use. I also fixed the credits screen and the README. 

I used my version of importing files into the menu. It makes it easier to import images, rather than including files that have nothing to do with it.

If you need any more information... feel free to ask.

P.S. You need the logo to be 512 x 512px              (I did a quick resize, hence why it is blurry)


GooeyHub:

JGelfand:
GooeyHub is a bot. If you're not on the whitelist it asks for admin verification. In either case, it runs a final test before the pull request is merged. @Cervator will notice this eventually, just give them time. See #107 for an example :D.

Cervator:
I noticed this earlier but have been struggling a bit on how to respond to it and the other comments just posted on other issues. There's a bit of ... awkward history in place here, especially when comparing this PR to some discussion yesterday. Probably best if I get a hold of @crazywolf132 in private sometime - you know where to find me.

The biggest issue with this PR right now is that it doesn't actually compile. I figure code was copy pasted from the [custom fork Path to Mani](https://github.com/BurntGameProductions/Path-to-Mani) without adjustments. There are a fair few other issues too.

PR #118, opened 2016-09-21
Add "Can't sell equipped items" option

Avalancs:
Add a new option "Can't sell equipped items" in menu. When enabled, selecting an equipped item in the sell window will have a disabled button "unequip it!" instead of "sell".
menu
sell window

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -119,6 +119,7 @@ Contributors
 * [LadySerenaKitty](https://github.com/LadySerenaKitty)
 * [askneller](https://github.com/askneller)
 * [JGelfand](https://github.com/JGelfand)
+* [AvaLanCS](https://github.com/Avalancs)
 
 ... and your name here? :-) More coming!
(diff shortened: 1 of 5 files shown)


GooeyHub:

Cervator:
Thanks for this!

@GooeyHub: add to whitelist

Would you be interested in adding something similar so the player cannot re-purchase the ship you're already in? Another case of simply disabling the button when that ship is selected.

GooeyHub:

Avalancs:
@Cervator yes, that was the next thing I wanted to look into!

Cervator:
Tested out and merged!

I tweaked the string used for the option to make it a little shorter and a bit clearer (no double negative with "don't" + "no") - it is now "Can sell used items" does that sound fair? :-)

Thanks!

Avalancs:
Yes, thank you :)

Issue #124, opened 2016-10-04
Volume settings in main menu (not just in-game)


Cervator:
Got to thinking about this when I had to go in-game to test volume settings. Maybe another one of interest to you @Avalancs :-)

Eventually we need to figure out nested settings screens, like in Terasology. Am not sure how much work that would be. Really the in-game and main menu could have Settings buttons that both go to the same sub-menu.

Avalancs:
Yes, that sounds reasonable, especially since there could be more options in the future. I need to learn a bit more about the UI code, but making a new screen right now seems simple enough.

Cervator:
Closing as complete :-)

PR #125, opened 2016-10-06
Add sfx and music volume to Main Menu -> Options

Avalancs:
For #124 and @Cervator :)
Added music and sfx volume to main menu -> options. No nested menus yet, sorry.
I also felt it weird that volume goes from low to high, so I added a medium option.
volume_med

### DIFF
--- a/main/src/org/destinationsol/game/sound/MusicManager.java
+++ b/main/src/org/destinationsol/game/sound/MusicManager.java
@@ -108,8 +108,8 @@ public final class MusicManager {
         }
     }
 
-    public void resetVolume(GameOptions options)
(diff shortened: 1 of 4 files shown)


GooeyHub:

main/src/org/destinationsol/game/sound/MusicManager.java
```diff
@@ -108,8 +108,8 @@ public void StopMusic() {
         }
     }
 
-    public void resetVolume(GameOptions options)
+    public void setVolume(float volumeMultiplier)
```
Avalancs: Wanted to break the dependency here.

Issue #119, opened 2016-09-21
When the game is paused, map is too zoomed in


Avalancs:
After pausing the game, accessing the map it is weirdly zoomed in. Closing the map and unpausing the game, then re-accessing the map it is fine. Pause and open again, and it's zoomed in again.
![zoom_bug_when_paused](https://cloud.githubusercontent.com/assets/18668703/18727141/a7a1b7e0-8047-11e6-94f9-4ca827c47fab.jpg)

I'm using the latest version from the development branch, and I could reproduce this from windows 7 as well as XUbuntu

Avalancs:
Update: from what I understand of the code, the SolCam:update() function is responsible for moving the camera, the ship, and managing the map zoom and view distance. But when the game is paused, this code is not run.
When the game starts, first time there is a zooming animation, and I think the same happens with the map when it is opened, but so fast you cannot see it. So when the game is paused, and the map is opened, the animation does not happen, and the view distance stays at a very low value, so you can barely see anything.

It would be nice to separate the camera calculations, the ship movement, and the map related calculations. Then the map calculations+animations could be run even when the game is paused.

vampcat:
Fixed with #126.

PR #126, opened 2016-10-23
fix map doesn't look right when the game is paused (issue #119)

Avalancs:
The issue was that the same camera is used for drawing the game, and the map, and the zoom did not get updated while the game was paused (and the skull icons did not get animated as well).

I extracted the zoom code in SolCamera so that it could be called when the game is paused.


GooeyHub:

PR #127, opened 2016-11-27
Item overflow error fix

Avalancs:
Second part to #114 and also a fix to http://forum.terasology.org/threads/crash-related-to-full-inventory.1643/

If somehow the player got an item that spanned to the 5th page of the inventory, quit the game, and then tried to play again, then an assertion error was thrown because of the 5th page. For example having a ship with a small and a large gun, filling up all the inventory space, then buying a ship with 2*large guns, the small gun would get unequipped and forms the 5th page.

I removed that assertion error, and verified that the game can start now. It is not possible to pick up an item that would add another entry to the 5th page (so you can pick up ammo and buy repair kits, if they stack with existing items), and after the player sells the excess items, he/she can pick up items normally.


GooeyHub:

PR #132, opened 2017-03-12
Soundtrack for Steam

Rulasmur:
Added steam script for soundtrack depot.

I've also created the depot on steamworks for it.

- [x] create SteamWorks DLC
- [x] create Depot 
- [x] Convert Music to mp3
- [x] create depot script
- [ ] create Icon for soundtrack
- [ ] pricing?
- [ ] AlbumArtwork.png 400x400
- [ ] licensing

### DIFF
--- a/steam/ContentBuilder/scripts/depot_build_608570.vdf
+++ b/steam/ContentBuilder/scripts/depot_build_608570.vdf
@@ -1,0 +1,31 @@
+"DepotBuildConfig"
+{
+	// Set your assigned depot ID here
+	"DepotID" "608570"
+
+	// Set a root for all content.
+	// All relative paths specified below (LocalPath in FileMapping entries, and FileExclusion paths)
+	// will be resolved relative to this root.
+	// If you don't define ContentRoot, then it will be assumed to be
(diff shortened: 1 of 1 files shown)


GooeyHub:

msva:
> - [ ] Convert Music to mp3

But why?

Rulasmur:
@msva That is what steam requires

Cervator:
@NeonInsect did express interest in the past to have the soundtrack released on Steam, just never had a chance to. Interesting that they only support MP3, which is tricky legally. I wouldn't want to have them in the main repo in that format, they're large files too for that matter.

Pricing should be free, license is still held by Neon with us allowed to use it, but not charge for it. Icon/album is one of the new ones we prepared, see https://itunes.apple.com/gb/album/destination-sol-1-ep/id1130523355 for instance :-)

I think we should make a new repo for *all* the Steam stuff, like https://github.com/MovingBlocks/DestSolAndroid and let it be included as a SteamFacade only if you need to work on the Steam stuff.

Rulasmur:
Restored branch as I had forgotten it was still in use

PR #152, opened 2017-05-14
Gestaltifying everything!

vampcat:
Gestaltify everything!

Updates to the latest gestalt version (6.0.0), and loads all assets via gestalt/splits apart assets so that they can be loaded via gestalt.


GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

main/src/org/destinationsol/SolApplication.java
```diff
@@ -250,7 +248,8 @@ public OggSoundManager getSoundManager() {
         return soundManager;
     }
 
-    public void paused() {
+    // TODO: Why do we even have this method? Look into it's removal.
+    public void pauseApplication() {
```
Cervator: Guessing it has to do either with the window losing focus or the in-game menu being used?

Cervator: Also: nitpick for `it's` instead of `its` :D But not enough to really worry about ;-)

main/src/org/destinationsol/assets/emitter/EmitterFileFormat.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.file.Paths;
+import java.util.List;
+
+@RegisterAssetFileFormat
+public class EmitterFileFormat extends AbstractAssetFileFormat<EmitterData> {
```
Cervator: Usual tiny nag message about javadoc, even if trivial :D

main/src/org/destinationsol/game/chunk/ChunkFiller.java
```diff
@@ -247,9 +247,9 @@ private void fillFarJunk(SolGame game, Vector2 chCenter, RemoveController remove
         TextureManager textureManager = game.getTexMan();
 
         for (int i = 0; i < count; i++) {
-            // Select a random far junk texture
+            // Select a random far junk atlas
```
Cervator: This reads a little funny. Made sense with "texture" but less so with "atlas" since it indicates sort of a potential collection? Not that a texture couldn't also be a collection if it has multiple things. Maybe something like "Select a random entry from the far junk atlas"

main/src/org/destinationsol/game/planet/PlanetManager.java
```diff
@@ -60,10 +60,10 @@ public PlanetManager(TextureManager textureManager, HullConfigManager hullConfig
         mySysConfigs = new SysConfigs(textureManager, hullConfigs, itemManager);
         myMazeConfigs = new MazeConfigs(textureManager, hullConfigs, itemManager);
 
-        mySystems = new ArrayList<SolSystem>();
-        myMazes = new ArrayList<Maze>();
-        myPlanets = new ArrayList<Planet>();
-        myBelts = new ArrayList<SystemBelt>();
+        mySystems = new ArrayList<>();
```
Cervator: Not necessarily complaining, but why does the type detail go away here? For something like `Planet planet : myPlanets` later wouldn't it complain about generics/type somehow?

main/src/org/destinationsol/game/planet/PlanetTiles.java
```diff
@@ -34,7 +34,7 @@
     private final Map<SurfaceDirection, Map<SurfaceDirection, List<Tile>>> myGroundTiles;
 
     public PlanetTiles(TextureManager textureManager, String groundFolder, FileHandle configFile) {
-        myGroundTiles = new HashMap<SurfaceDirection, Map<SurfaceDirection, List<Tile>>>();
+        myGroundTiles = new HashMap<>();
```
Cervator: Losing some types again, dunno why. Probably makes sense somehow.

main/src/org/destinationsol/SolApplication.java
vampcat: The function for application losing focus is pause(), the paused() (now pauseApplication() for clarity) is an entirely different function that is never used - which is why inspect for removal, with the only thing being maybe saving the logic inside it somewhere.

Also, fixed the typo :D Appreciate the nitpick :)
(Idk why I somtimes use it's instead of its without thinking)

main/src/org/destinationsol/assets/emitter/EmitterFileFormat.java
vampcat: Yeah, I will go through all these classes (2 more to be added, held up due to exams) in one pass and add javadoc to them before this PR is merged :)

main/src/org/destinationsol/game/chunk/ChunkFiller.java
vampcat: Oh damn, this was a refactor thing, not intentional. Definitely should be junk texture.
Sorry! :P

main/src/org/destinationsol/game/planet/PlanetManager.java
vampcat: No, it won't.

The reason I explicitly make it a point to remove the type detail is, CheckStyle complains about it (Java 7.0 and above don't need explicit types, since it can be deduced from the type of myPlanets). Both are identical, the reduced down version is just cleaner. Code simplification, less warnings, I'm happy with that :D
Although if you want, I can revert back - not that I recommend it.

And no, it won't complain about generics. I've removed the type info in almost every class I've touched AND have converted all for loops to foreach loops ( `Planet planet : myPlanets` ) and yet it works perfectly :)

main/src/org/destinationsol/game/planet/PlanetTiles.java
vampcat: Same reason as above, exact same functionality with cleaner code.
Aaaaand I hate the checkstyle warnings.

vampcat: In fact, in most places, I'm shifting such things to the class itself, so that the variable declaration becomes
`private  final Map<SurfaceDirection, Map<SurfaceDirection, List<Tile>>> myGroundTiles = new HashMap<>();`
Makes a lot more sense over there.

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
Bam! Great work. Tested the game in general but quite thoroughly and nothing major stuck out. Woo!

PR #160, opened 2017-06-11
Added new weapon (dualBlaster) used on federal ships.

Sigma-One:
# Description
> Adds new weapon appearing on federal ships. Works just like the blaster but has twice the fire rate.

# Testing
> Find a federal ship, and look at it's weapons.


GooeyHub:

PR #162, opened 2017-06-19
added imperialCapital ship

Sigma-One:
# Description
> Added new imperialCapital ship and moved the patroling imperialTinys to guard it.

# Testing
> Find the new ship in the starting system.


GooeyHub:

modules/core/assets/ships/imperialCapital/imperialCapital.json
```diff
@@ -0,0 +1,443 @@
+{
+    "size": 3.5,
+    "maxLife": 800,
+    "type": "big",
+    "price": 20000,
+    "e1Pos": "0.08 0.37",
+    "e2Pos": "0.08 0.63",
+    "displayName": "Imperial Capital Ship",
```
vampcat: Can't we simply name it as `Imperial Capital` ? I think the Ship might be kinda redundant here.

Sigma-One: Usually in sci-fi they are referred to as "capital ships" so I just named it that. Also, the name doesn't really matter right now as the capitals cannot be bought.

GooeyHub:

modules/core/assets/ships/imperialCapital/imperialCapital.json
Cervator: Alternatively we could give it a flavor name and start referring to a _class_ of capital ships :-)

That's probably a good idea in either case. Gives us an easy way to associate all capital ships with class behavior, like making them avoid dipping below the orbital band of a planet under normal circumstances.

Cervator:
Merged - although it won't say so on GitHub as it'll be missing an extraneous merge commit from the fork.

@Sigma-One: pro tip on updating forks: use Git Rebase to put your commits in front of the latest from upstream, not Git Merge - that just adds another merge commit, in some cases making old commits appear like new commits (did not in this case). I had pulled your stuff a bit earlier before you used a merge to update your branch, just got sidelined so didn't push this at the time :-)

It is good to keep branches up to date, but not critical if there are no conflicts or likely issues. Merging into a fork instead of using rebase just tends to cloud up the history though.

Your fork's develop branch may not show as diverged - unsure if you can cleanly pull latest and get a clean setup again. You may have to force push to your fork and it would be a good idea to start using topic branches. Can advise further on Slack if needed, just let me know :-)

PR #167, opened 2017-06-28
Moving stuff around

vampcat:
- Fixed project structure for PC Facade
- Some more code clean-up, mostly removing unused imports
- Unified settings and save file for gradle and IntelliJ so it behaves the same when you open the same from either source.


GooeyHub:

PR #168, opened 2017-07-02
Refactoring, and addition of new "Station" layer for rendering

vampcat:
Mostly refactoring of code (specifically, finally figured out (more like guessed) that Dra means Drawable), and addition of a new Drawable layer to allow for Big ships + Stations to act normally.


GooeyHub:

PR #171, opened 2017-08-12
Roll back Gestalt to v5.1.3 due to a release quirk 

Cervator:
Some bad v6 artifacts got loose and were stuck in a few Gradle caches here and there. Artifactory has been tidied up and Jenkins' Gradle cache cleaned. Not sure if anybody else got the bad version other than me and @vampcat - but if `develop` doesn't compile after a `gradlew cleanIdea idea` for anybody else that's a thing to try :-)

We'll try a round two for adopting v6 soon enough, but this way we can start building and doing little things again.


GooeyHub:

PR #175, opened 2017-09-03
game balance: lowered time between shots for blaster

Avalancs:
I reduced the time between shots for the blaster so when starting out you have a better chance to take out enemies, and also makes shooting asteroids less tedious.


GooeyHub:

PR #188, opened 2017-12-20
Custom spawn stations

Steampunkery:
# Description
This PR fixes than long outstanding problem that you can't choose the station you initially spawn at. The problem is handled by removing the `engine/src/main/resources/assets/configs/mainStationConfig.json` file and replacing it with a file called `startingStation.json` in each module's `assets/configs` folder. These files are laid out exactly the same as the old `mainStationConfig.json`, with the default file looking like this:
```json
{
    "hull": "core:station",
    "items": "core:gun core:gun",

    "guard": {
        "hull": "core:imperialBig",
        "items": "core:fixedGun core:fixedShotGun core:mediumArmor+core:heavyArmor core:shield+core:bigShield",
        "density": 3
    }
}
```

# Testing
Add the above code to a file named `startingStation.json` in each module's `assets/configs` folder.


GooeyHub:

GooeyHub:

GooeyHub:

vampcat:
Okay, time to take a step back. You know how you can create branches starting from a specific commit? Well, this branch should have originated from MovingBlocks/develop, and not Steampunkery/develop. Another small thing, the ships should have gone in their own branch (formicModule or something) so that your develop would have been clean.

Either way, I would request that you create a new PR for this, with only the required commits.
(Create a new branch from MovingBlocks/develop, and either cherry pick the relevant commits, or copy-paste the final files and commit them manually.)

With that said, I'm still reviewing the relevant commits on this PR, so that you don't have to waste time waiting for my review.

engine/src/main/java/org/destinationsol/assets/Assets.java
```diff
@@ -40,6 +43,8 @@
 public abstract class Assets {
     private static AssetHelper assetHelper;
     private static Set<ResourceUrn> textureList;
+    // This map is initialized after the player selects their ship, it maps the human readable names of ships to their internal URIs
+    public static Map<String, String> playerSpawnConfigIdMap = new HashMap<String, String>();
```
vampcat: I like the idea of having this map, but I don't like the idea of having it here. `Assets` is supposed to be a helper class to assist in loading assets. So, what I would suggest is that you move this map to a different location - maybe in NewShipScreen for the time being?

engine/src/main/java/org/destinationsol/game/GalaxyFiller.java
```diff
@@ -120,6 +122,15 @@ private FarShip build(SolGame game, ShipConfig cfg, Faction faction, boolean mai
         }
         return s;
     }
+    
+    public JsonValue checkRootNull(Json json) {
```
vampcat: CheckRootNull implies you are just checking, and should return a boolean. Since our primary objective is not checking but rather obtatining the JsonValue, this should be names slightly differently. `GetRootNode()` ?

PR #228, opened 2018-01-22
JSON sanitization!

Adrijaned:
# Description
> Adds a JSON file sanitization with more descriptive error messages, currently for ships.

# Testing
> Try running some functional ship ingame. It should work. Now change the ship's JSON to make it not valid. Try running again,

# Outstanding Work
- [ ] Add support for other JSON files

### DIFF
--- a/engine/src/main/java/org/destinationsol/assets/json/SolJsonException.java
+++ b/engine/src/main/java/org/destinationsol/assets/json/SolJsonException.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/assets/json/SolJsonException.java
```diff
@@ -0,0 +1,36 @@
...
+    public SolJsonException(String msg) {
+        super(msg);
+    }
+
+    /**
+     * This Exception is meant to supplement more descriptive error messages.
+     * <p>
+     * If somebody does not accept that and does not supply the exception message, notify the user in case exception gets called.
+     */
+    public SolJsonException() {
```
Steampunkery: Why not just omit this constructor entirely? It would force the user of the exception to write a message.

vampcat:
@GooeyHub: add to whitelist

GooeyHub:

PR #233, opened 2018-01-24
removed settings.ini from repo

NicholasBatesNZ:
removed settings.ini from repo so that it doesn't bother you when committing changes


GooeyHub:

vampcat:
Yeah, not gonna do this :P

As discussed on slack, this is not the way to go. We need to devise an alternate system. Till then, we live with the annoying settings file.

NicholasBatesNZ:
See #237

PR #234, opened 2018-01-25
custom engine speeds and accelerations and relevant json changes

Steampunkery:
# Description
This PR allows creators of engines to specify a top speed for their engine and a custom acceleration. This means that engines can have varying speeds for different prices. `topSpeedMultiplier` and `acceleration` are not values dictating the top speed and acceleration of the engine, but rather multipliers. So, an acceleration of 2 and a topSpeedMultiplier of 1 means that the ship accelerates to the normal top speed twice as fast while an acceleration of 1 and a topSpeedMultiplier of 2 means the ship accelerates like normal but reaches a speed of twice as fast as normal.

### DIFF
--- a/engine/src/main/java/org/destinationsol/common/SolMath.java
+++ b/engine/src/main/java/org/destinationsol/common/SolMath.java
@@ -487,9 +487,13 @@ public class SolMath {
         }
         return res;
     }
+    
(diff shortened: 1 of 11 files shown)


GooeyHub:

Issue #54, opened 2015-07-26
Asteroids can spawn inside of larger asteroids


grauerkoala:
I've come across an asteroid that was stuck within a much larger asteroid today, sadly bouncing around, unable to escape his captor.
I failed to get a screenshot, but maybe I'll come across one of those (space) oddities again (unless I find and fix the issue first, that is).
##

NicholasBatesNZ:
Closing as superseded by #299

PR #238, opened 2018-01-26
Potential fix for issue #54

NicholasBatesNZ:
I have not been able to test the fix, but theoretically this should stop asteroids from spawning inside of each other. When a FarAsteroid is converted into an Asteroid it makes sure that it is not overlapping an existing Asteroid.
Issue #54

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/ObjectManager.java
+++ b/engine/src/main/java/org/destinationsol/game/ObjectManager.java
@@ -22,6 +22,7 @@ import org.destinationsol.Const;
 import org.destinationsol.common.DebugCol;
 import org.destinationsol.common.SolColor;
 import org.destinationsol.common.SolMath;
+import org.destinationsol.game.asteroid.FarAsteroid;
 import org.destinationsol.game.drawables.Drawable;
 import org.destinationsol.game.drawables.DrawableManager;
 import org.destinationsol.game.drawables.FarDrawable;
@@ -132,8 +133,8 @@ public class ObjectManager {
             }
             if (isNear(fod, camPos, ts)) {
                 SolObject o = fo.toObj(game);
-                // Ensure that StarPorts are added straight away so that we can see if they overlap
-                if (o instanceof StarPort) {
+                // Ensure that StarPorts and Asteroids are added straight away so that we can see if they overlap
+                if (o instanceof StarPort || o instanceof FarAsteroid) {
                     addObjNow(game, o);
                 } else {
                     addObjDelayed(o);
--- a/engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
+++ b/engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
@@ -75,8 +75,7 @@ public class FarAsteroid implements FarObj {
                     else {
                         distanceFromAsteroid.scl((size + .5f) / distance);
                     }
-                    position = asteroidPosition.cpy().add(distanceFromAsteroid);
-                    SolMath.free(SolMath.distVec(asteroidPosition, position));
+                    position = asteroidPosition.add(distanceFromAsteroid);
                 }
                 SolMath.free(distanceFromAsteroid);
             }
--- a/engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
+++ b/engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
@@ -57,7 +57,7 @@ public class FarAsteroid implements FarObj {
     }
 
     /**
-     * Ensures that this asteroid does not overlap any other asteroid, by adjusting its position
+     * Ensures that this asteroid does not overlap any other asteroid, by adjusting its position.
      */
     private void adjustDesiredPosition(SolGame game) {
         List<SolObject> objects = game.getObjMan().getObjs();
--- a/engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
+++ b/engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
@@ -26,52 +26,47 @@ import org.destinationsol.game.SolObject;
 import java.util.List;
 
 public class FarAsteroid implements FarObj {
-    private final Vector2 myPos;
-    private final float myAngle;
-    private final RemoveController myRemoveController;
-    private final float mySz;
-    private final Vector2 mySpd;
-    private final float myRotSpd;
-    private final TextureAtlas.AtlasRegion myTex;
+    private Vector2 position;
+    private final float angle;
+    private final RemoveController removeController;
+    private final float size;
+    private final Vector2 speed;
+    private final float rotateSpeed;
+    private final TextureAtlas.AtlasRegion tex;
 
-    public FarAsteroid(TextureAtlas.AtlasRegion tex, Vector2 pos, float angle, RemoveController removeController,
-                       float sz, Vector2 spd, float rotSpd) {
-        myTex = tex;
-        myPos = pos;
-        myAngle = angle;
-        myRemoveController = removeController;
-        mySz = sz;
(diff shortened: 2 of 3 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
```diff
@@ -49,7 +52,41 @@ public boolean shouldBeRemoved(SolGame game) {
 
     @Override
     public SolObject toObj(SolGame game) {
-        return game.getAsteroidBuilder().build(game, myPos, myTex, mySz, myAngle, myRotSpd, mySpd, myRemoveController);
+        Vector2 position = adjustDesiredPosition(game, myPos, mySz);
```
vampcat: After this line, `myPos` and the actual position of asteroid (position) are out of sync. Way better to modify the value of myPos, rather than create a new Vector2. Luckily for you, that's exactly what you've done (by mistake).

engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
```diff
@@ -49,7 +52,41 @@ public boolean shouldBeRemoved(SolGame game) {
...
+
+    /**
+     * Ensures that the position of any Asteroids do not overlap
+     * @param game SolGame instance
+     * @param desiredPosition Vector2 position that you would like the Asteroid to spawn at
+     * @param size float size of the asteroid
+     * @return the new Vector2 position of the asteroid, or desired if it does not overlap
+     */
+    private static Vector2 adjustDesiredPosition(SolGame game, Vector2 desiredPosition, float size) {
+        Vector2 newPosition = desiredPosition;
```
vampcat: Here, `newPosition` is a reference to desiredPosition. That means any change to one reflects in the other. Maybe you meant to do `newPosition = new Vector2(desiredPosition);`?
Either way, I recommend not creating a new vector here, and instead modifying desiredPosition instead (and hence make the return type void).

Also, since this method is already part of the class, why pass size and position as arguments? Simply make the method non-static, and take the variables from the class.

So, after all my suggested changes, this method becomes `private void adjustDesiredPosition(SolGame game)`

GooeyHub:

engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
```diff
@@ -26,52 +26,47 @@
...
+        return game.getAsteroidBuilder().build(game, position, tex, size, angle, rotateSpeed, speed, removeController);
     }
 
     /**
-     * Ensures that the position of any Asteroids do not overlap
-     * @param game SolGame instance
-     * @param desiredPosition Vector2 position that you would like the Asteroid to spawn at
-     * @param size float size of the asteroid
-     * @return the new Vector2 position of the asteroid, or desired if it does not overlap
+     * Ensures that the position of any Asteroids do not overlap by adjusting the position
```
vampcat: `Ensures that this asteroid does not overlap any other asteroid, by adjusting its position.

engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
```diff
@@ -80,13 +75,12 @@ private static Vector2 adjustDesiredPosition(SolGame game, Vector2 desiredPositi
                     else {
                         distanceVector.scl((size + .5f) / distance);
                     }
-                    newPosition = fromPosition.cpy().add(distanceVector);
-                    SolMath.free(SolMath.distVec(fromPosition, newPosition));
+                    position = fromPosition.cpy().add(distanceVector);
+                    SolMath.free(SolMath.distVec(fromPosition, position));
```
vampcat: whaaaat

This line makes no sense to me. Care to explain it?

engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
```diff
@@ -26,52 +26,47 @@
...
      */
-    private static Vector2 adjustDesiredPosition(SolGame game, Vector2 desiredPosition, float size) {
-        Vector2 newPosition = desiredPosition;
+    private void adjustDesiredPosition(SolGame game) {
         List<SolObject> objects = game.getObjMan().getObjs();
         for (SolObject object : objects) {
             if (object instanceof Asteroid) {
                 Asteroid asteroid = (Asteroid) object;
                 // Check if the positions overlap
                 Vector2 fromPosition = asteroid.getPosition();
```
vampcat: Rename `fromPosition` to something more.. sensible.

engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
```diff
@@ -80,13 +75,12 @@ private static Vector2 adjustDesiredPosition(SolGame game, Vector2 desiredPositi
                     else {
                         distanceVector.scl((size + .5f) / distance);
                     }
-                    newPosition = fromPosition.cpy().add(distanceVector);
-                    SolMath.free(SolMath.distVec(fromPosition, newPosition));
+                    position = fromPosition.cpy().add(distanceVector);
```
vampcat: You are discarding old position, and replacing it with a new vector that contains the same value as fromPosition + distanceVector. I think it's better to eliminate the new vector creation bit, and simply do
```position.set(fromPosition)
position.add(distanceVector)```
(Note that this prevents an object creation, which is huge.

File: engine/src/main/java/org/destinationsol/game/ObjectManager.java [hunk 7]
```diff
@@ -126,8 +127,8 @@ public void update(SolGame game) {
             }
             if (isNear(fod, camPos, ts)) {
                 SolObject o = fo.toObj(game);
-                // Ensure that StarPorts are added straight away so that we can see if they overlap
-                if (o instanceof StarPort) {
+                // Ensure that StarPorts and Asteroids are added straight away so that we can see if they overlap
+                if (o instanceof StarPort || o instanceof FarAsteroid) {
```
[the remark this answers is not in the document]
Adrijaned: With addObjectNow(), you are adding objects immediately into the current frame of the game; if you for instance added object if the half of the execution of current frames update(), half of the methods that are executed each frame wouldn't be performed while aware of the object. If some of these methods, for instance, set up these objects, for instance some of their inner variables, like shouldBeRemoved (not the greatest example), the object, inserted with addObjectNow would be removed a frame or two later. With addObjectsDelayed, however, you ensure that the objects will be inserted right before the beginning of the next frame, and thus everything that has to be executed in the frame will be aware of these objects and it may take them in account.

GooeyHub:

GooeyHub:

engine/src/main/java/org/destinationsol/game/asteroid/FarAsteroid.java
```diff
@@ -57,28 +57,28 @@ public SolObject toObj(SolGame game) {
...
+                        distanceFromAsteroid.scl((asteroid.getSize() + .5f) / distance);
                     }
                     else {
-                        distanceVector.scl((size + .5f) / distance);
+                        distanceFromAsteroid.scl((size + .5f) / distance);
                     }
-                    position = fromPosition.cpy().add(distanceVector);
-                    SolMath.free(SolMath.distVec(fromPosition, position));
+                    position = asteroidPosition.cpy().add(distanceFromAsteroid);
+                    SolMath.free(SolMath.distVec(asteroidPosition, position));
```
Adrijaned: What the hell is this line for?

GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

NicholasBatesNZ:
Closing as it is not the ideal solution to #299

PR #241, opened 2018-01-27
Rigid body script

Steampunkery:
Just a script (the python version and the java version have identical outputs) to format and correct ship json files after box-2d does it's thing. It renames `"rigidBodies"` to `"rigidBody"`, removes the wrapper list from the now renamed `"rigidBody"` attribute and also remove the useless `"name"` and `"imagePath"` attributes from `"rigidBody"`.


GooeyHub:

Adrijaned:
@vampcat Terasology has some kind of utility repo, IIRC, where there are stored all the helpful things and stuff, that are not directly part of the code. Couldn't something like that made for DS?

vampcat:
Terasology stores all utility things under the MB org. Which is where we currently house all DS utility (only Box2D for now, admittedly) as well.
However, if we have more than 1 or 2 utilites, then having a separate repo will definitely make sense. Do you forsee any other utilities?

Adrijaned:
Dunno. Maybe I'll force myself one day into writing a GUI app for creating the ships JSON, and guns, and engines... (Not very likely to happen in near future, but you know, the possibilities...)

vampcat:
As soon as I have time to think about this in more detail, or you finish writing the gui - whichever comes first - we'll have another discussion about how to best deal with this :)

PR #244, opened 2018-02-02
Hero encapsulation

Adrijaned:
# Description
> Encapsulation of hero and transcendentHero in single class. See #243 for details

Issue #243, opened 2018-01-30
Ensure game hero is never null


eviltak:
## Problem

Currently, the game searches for the player ship on every update and, if found, a reference to it is assigned to `SolGame.hero`. However, if the hero is in the "transcendental" state (travelling through a StarPort) or no hero exists at all, `SolGame.hero` remains null. 

https://github.com/MovingBlocks/DestinationSol/blob/55d228c45fd94ed45393ec867df5d602a85ee2c6/engine/src/main/java/org/destinationsol/game/SolGame.java#L340-L350

This leads to a number of problems:

1. The player ship is searched for every frame, leading to suboptimal performance especially when a large number of `SolObject`s exist.
2. `SolGame.getHero` returns null when a null reference is not expected. A large chunk of the codebase - I daresay even the entire game - considers that a hero exists without fail. As such, NPEs may crop up here and there without explanation.
3. The `hero` should _never_ be null. As stated above, Destination Sol gameplay is wholly hero-centric; a player must therefore exist at all times. However, the code currently does not provide assurance that `getHero` will never return null (as evident from Slack discussions entailing the `Optional` refactor in #236).

## Proposed solution
Currently, a null `hero` has only two purposes:
- to signify that the hero is in the transcendental state or
- to allow for free camera movement without a hero for debugging.
https://github.com/MovingBlocks/DestinationSol/blob/55d228c45fd94ed45393ec867df5d602a85ee2c6/engine/src/main/java/org/destinationsol/game/SolCam.java#L66-L71

To work around this, the following can be done:
- `SolGame.hero` is set only once, on game load. `SolGame.transcendentalHero` is set only once, either on load or when the player enters the transcendental state for the first time.
- A boolean, `SolGame.isHeroTranscendental` is added to signify if the hero is in the transcendental state. If the value returned by the getter is true, callers should use `getTranscendentalHero` instead.
- Instead of checking for the non-existence of a hero, simply allow free camera movement if `DebugOptions.DIRECT_CAM_CONTROL` is true:
```java
public class SolCam {
    // ...
    public void update(SolGame game) {
        // ...
        float ts = game.getTimeStep();
        if (DebugOptions.DIRECT_CAM_CONTROL) {
            applyInput(game);
        }
        else {
            Vector2 heroPos = hero.getHull().getBody().getWorldCenter();
            // ...
        }
```

## Points to consider
[...]

PR #246, opened 2018-02-18
Hero encapsulation

Adrijaned:
Re-opened PR #244, because that got _funny_, and, well, broken see #244, resp. #243 for details on this.

This is to simplify working with the players ship when programming, since previously, there was only a global pool of objects, from each on every iteration players ship was found, and then returned, and everyone had to check everywhere what is the player's actually at that given point in time.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/Hero.java
+++ b/engine/src/main/java/org/destinationsol/game/Hero.java
@@ -16,7 +16,7 @@
 package org.destinationsol.game;
 
 import com.badlogic.gdx.math.Vector2;
-import org.destinationsol.common.SolDescriptiveException;
+import org.destinationsol.common.SolException;
 import org.destinationsol.game.input.Pilot;
 import org.destinationsol.game.item.Armor;
 import org.destinationsol.game.item.ItemContainer;
@@ -41,18 +41,18 @@ public class Hero {
     public Hero(SolShip shipHero) {
         this.shipHero = shipHero;
         if (shipHero == null) {
-            throw new SolDescriptiveException("Something tries to create hero, when there is no ship linked to him. Please report this..");
+            throw new SolException("Something tries to create hero, when there is no ship linked to him.");
         }
         isTranscendent = false;
     }
 
-    public void toTranscendent(StarPort.Transcendent transcendentHero) {
+    public void setTranscendent(StarPort.Transcendent transcendentHero) {
         this.transcendentHero = transcendentHero;
         transcendentHeroShip = transcendentHero.getShip();
         isTranscendent = true;
     }
 
-    public void toSolShip(SolShip hero) {
+    public void setSolShip(SolShip hero) {
         isDead = false;
         this.shipHero = hero;
         isTranscendent = false;
@@ -70,20 +70,20 @@ public class Hero {
         return isTranscendent ? transcendentHeroShip.getPilot() : shipHero.getPilot();
     }
 
-    public SolShip getShipHero() {
+    public SolShip getShip() {
         if (isTranscendent) {
-            throw new SolDescriptiveException("Something is trying to get a SolShip hero while the hero is Transcendent state.");
+            throw new SolException("Something is trying to get a SolShip hero while the hero is Transcendent state.");
         }
         return shipHero;
     }
 
-    public SolShip getShipHeroUnchecked() {
+    public SolShip getShipUnchecked() {
         return shipHero;
     }
 
     public StarPort.Transcendent getTranscendentHero() {
         if (!isTranscendent) {
-            throw new SolDescriptiveException("Something is trying to get a Transcendent hero while the hero is in SolShip state.");
+            throw new SolException("Something is trying to get a Transcendent hero while the hero is in SolShip state.");
         }
         return transcendentHero;
     }
@@ -191,7 +191,7 @@ public class Hero {
 
     private void assertNonTranscendent() {
         if (isTranscendent) {
-            throw new SolDescriptiveException("Something is trying to get a property of hero that doesn't exist in transcendent state, while the hero is in transcendent state.");
+            throw new SolException("Something is trying to get a property of hero that doesn't exist in transcendent state, while the hero is in transcendent state.");
         }
     }
 
--- a/engine/src/main/java/org/destinationsol/game/Hero.java
+++ b/engine/src/main/java/org/destinationsol/game/Hero.java
@@ -15,43 +15,148 @@
  */
 package org.destinationsol.game;
 
+import com.badlogic.gdx.math.Vector2;
 import org.destinationsol.common.SolDescriptiveException;
 import org.destinationsol.game.input.Pilot;
+import org.destinationsol.game.item.Armor;
+import org.destinationsol.game.item.ItemContainer;
+import org.destinationsol.game.item.Shield;
+import org.destinationsol.game.item.TradeContainer;
+import org.destinationsol.game.ship.FarShip;
+import org.destinationsol.game.ship.ShipAbility;
 import org.destinationsol.game.ship.SolShip;
+import org.destinationsol.game.ship.hulls.Hull;
 
 public class Hero {
     private SolShip hero;
     private StarPort.Transcendent transcendentHero;
+    private FarShip transcendentHeroShip;
     private boolean isTranscendent;
-    private final String transcendentNotExistingPropertyExceptionMessage = "Something is trying to get a property of hero that doesn't exist in transcendent state, while the hero is in transcendent state.";
 
-    public Hero(SolShip hero, StarPort.Transcendent transcendentHero) {
+    public Hero(SolShip hero) {
         this.hero = hero;
-        this.transcendentHero = transcendentHero;
-        if (hero != null && transcendentHero != null) {
-            throw new SolDescriptiveException("Something tries to set both hero and transcended hero. This should not be possible. Please report this.");
+        if (hero == null) {
+            throw new SolDescriptiveException("Something tries to create hero, when there is no ship linked to hiim. Please report this..");
         }
-        isTranscendent = transcendentHero != null;
+        isTranscendent = false;
+    }
+
+    public void toTranscendent(StarPort.Transcendent transcendentHero) {
+        this.transcendentHero = transcendentHero;
+        transcendentHeroShip = transcendentHero.getShip();
+        isTranscendent = true;
+    }
+
+    public void toSolShip(SolShip hero) {
+        this.hero = hero;
+        isTranscendent = false;
     }
 
     public boolean isTranscendent() {
         return isTranscendent;
     }
 
-    public Pilot getPilot(){
-        if (isTranscendent) {
-            throw new SolDescriptiveException(transcendentNotExistingPropertyExceptionMessage);
-        }
-        return hero.getPilot();
+    public Pilot getPilot() {
+        return isTranscendent ? transcendentHeroShip.getPilot() : hero.getPilot();
     }
 
     public SolShip getHero() {
+        if (isTranscendent) {
+            throw new SolDescriptiveException("Something is trying to get a SolShip hero while the hero is Transcendent state.");
+        }
+        return hero;
+    }
+
+    public SolShip getHeroUnchecked() {
         return hero;
     }
 
     public StarPort.Transcendent getTranscendentHero() {
+        if (!isTranscendent) {
+            throw new SolDescriptiveException("Something is trying to get a Transcendent hero while the hero is in SolShip state.");
+        }
         return transcendentHero;
     }
 
+    public float getAngle() {
+        return isTranscendent ? transcendentHeroShip.getAngle() : hero.getAngle();
+    }
 
+    public Shield getShield() {
+        return isTranscendent ? transcendentHeroShip.getShield() : hero.getShield();
+    }
+
+    public Armor getArmor() {
+        return isTranscendent ? transcendentHeroShip.getArmor() : hero.getArmor();
+    }
+
+    public ShipAbility getAbility() {
+        onlySolShipHero();
+        return hero.getAbility();
+    }
+
+    public float getAbilityAwait() {
+        onlySolShipHero();
+        return hero.getAbilityAwait();
+    }
+
+    public boolean canUseAbility() {
+        onlySolShipHero();
+        return hero.canUseAbility();
+    }
+
+    public Vector2 getSpd() {
+        return isTranscendent ? transcendentHero.getSpd() : hero.getSpd();
+    }
+
+    public float getAcc() {
+        onlySolShipHero();
+        return hero.getAcc(); // Transcendent hero doesn't accelerate
+    }
+
+    public Vector2 getPosition() {
+        return isTranscendent ? transcendentHero.getPosition() : hero.getPosition();
+    }
+
+    public Hull getHull() {
+        onlySolShipHero(); // Transcendent hero has no hull
+        return hero.getHull();
+    }
+
+    public float getLife() {
+        return isTranscendent ? transcendentHeroShip.getLife() : hero.getLife();
+    }
+
+    public float getRotAcc() {
+        onlySolShipHero();
+        return hero.getRotAcc();
+    }
+
+    public float getRotSpd() {
+        onlySolShipHero();
+        return hero.getRotSpd();
+    }
+    public float getMoney() {
+        return isTranscendent ? transcendentHeroShip.getMoney() : hero.getMoney();
+    }
+
+    public void setMoney(float money) {
+        onlySolShipHero(); // Transcendent hero should by design be not modified
+        hero.setMoney(money);
+    }
+
+    public TradeContainer getTradeContainer() {
+        onlySolShipHero(); // Trading should by no means be possible when in transcendent state
+        return hero.getTradeContainer();
+    }
+
+    public ItemContainer getItemContainer() {
+        return isTranscendent ? transcendentHeroShip.getIc() : hero.getItemContainer();
+    }
+
+    private void onlySolShipHero() {
+        if (isTranscendent) {
+            throw new SolDescriptiveException("Something is trying to get a property of hero that doesn't exist in transcendent state, while the hero is in transcendent state.");
+        }
+    }
 }
--- a/engine/src/main/java/org/destinationsol/game/Hero.java
+++ b/engine/src/main/java/org/destinationsol/game/Hero.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2017 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.game;
+
+import org.destinationsol.common.SolDescriptiveException;
+import org.destinationsol.game.input.Pilot;
+import org.destinationsol.game.ship.SolShip;
+
+public class Hero {
+    private SolShip hero;
+    private StarPort.Transcendent transcendentHero;
+    private boolean isTranscendent;
+    private final String transcendentNotExistingPropertyExceptionMessage = "Something is trying to get a property of hero that doesn't exist in transcendent state, while the hero is in transcendent state.";
+
+    public Hero(SolShip hero, StarPort.Transcendent transcendentHero) {
+        this.hero = hero;
+        this.transcendentHero = transcendentHero;
+        if (hero != null && transcendentHero != null) {
+            throw new SolDescriptiveException("Something tries to set both hero and transcended hero. This should not be possible. Please report this.");
+        }
+        isTranscendent = transcendentHero != null;
+    }
+
+    public boolean isTranscendent() {
+        return isTranscendent;
+    }
+
+    public Pilot getPilot(){
+        if (isTranscendent) {
+            throw new SolDescriptiveException(transcendentNotExistingPropertyExceptionMessage);
+        }
+        return hero.getPilot();
+    }
+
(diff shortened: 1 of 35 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -0,0 +1,205 @@
...
+    private SolShip hero;
+    private StarPort.Transcendent transcendentHero;
+    private FarShip transcendentHeroShip;
+    private boolean isTranscendent;
+    private boolean isDead;
+
+    public Hero(SolShip hero) {
+        this.hero = hero;
+        if (hero == null) {
+            throw new SolDescriptiveException("Something tries to create hero, when there is no ship linked to hiim. Please report this..");
```
Cervator: Typo "hiim"

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -0,0 +1,205 @@
...
+        isDead = false;
+        this.hero = hero;
+        isTranscendent = false;
+    }
+
+    public boolean isTranscendent() {
+        return isTranscendent;
+    }
+
+    public boolean isNonTranscendent() {
```
Cervator: Maybe this is just copy pasted / moved, but wouldn't it make sense to just have the single `isTranscendent` method?

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -0,0 +1,205 @@
...
+        return !isTranscendent;
+    }
+
+    public Pilot getPilot() {
+        return isTranscendent ? transcendentHeroShip.getPilot() : hero.getPilot();
+    }
+
+    public SolShip getHero() {
+        if (isTranscendent) {
+            throw new SolDescriptiveException("Something is trying to get a SolShip hero while the hero is Transcendent state.");
```
Cervator: Maybe briefly explain why that matters in a parenthesis at the end of the string "(this makes no sense because X)"

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+package org.destinationsol.game.screens;
+
+import org.destinationsol.GameOptions;
+import org.destinationsol.ui.SolUiControl;
+
+import java.util.List;
+
+class CameraKbControl {
```
Cervator: Would it be worth writing out this class name fully so `CameraKeyboardControl` ? No real harm in spelling it out

engine/src/main/java/org/destinationsol/game/screens/ChangeShip.java
```diff
@@ -119,7 +120,8 @@ private void changeShip(SolGame game, SolShip hero, ShipItem selected) {
         SolShip newHero = game.getShipBuilder().build(game, hero.getPosition(), new Vector2(), hero.getAngle(), 0, hero.getPilot(),
                 hero.getItemContainer(), newConfig, newConfig.getMaxLife(), hull.getGun(false), g2, null,
                 ei, new ShipRepairer(), hero.getMoney(), hero.getTradeContainer(), hero.getShield(), hero.getArmor());
-        game.getObjMan().removeObjDelayed(hero);
+        game.getObjMan().removeObjDelayed(hero.getHero());
         game.getObjMan().addObjDelayed(newHero);
+        game.getHero().toSolShip(newHero); // HACK
```
Cervator: Helps to explain why, maybe convert into a `// TODO: This is a hack because X which we need to fix with by doing Y` :-)

engine/src/main/java/org/destinationsol/game/screens/MainScreen.java
```diff
@@ -122,8 +125,11 @@ public MainScreen(float resolutionRatio, RightPaneLayout rightPaneLayout, SolApp
         mercControl = new SolUiControl(mercArea, true, gameOptions.getKeyMercenaryInteraction());
         mercControl.setDisplayName("Mercs");
         controls.add(mercControl);
+        freeCamControl = new SolUiControl(null, false, gameOptions.getKeyFreeCameraMovement()); // No button, since on mobile, it should be ideally controlled straightly by dragging.
```
Cervator: May want to put the comment above the code line since it is kinda long-ish

Cervator:
Tested by starting a game with 3 systems in loaded imperial large. Went to the closest starlane and tried to enter. Crashed with the following:

```
ERROR [LWJGL Application] (SolApplication.java:123) - Fatal Error:
org.destinationsol.common.SolDescriptiveException: Something is trying to get a property of hero that doesn't exist in transcendent state, while the hero is in transcendent state.
	at org.destinationsol.game.Hero.onlySolShipHero(Hero.java:194)
	at org.destinationsol.game.Hero.getHull(Hero.java:132)
	at org.destinationsol.game.SolCam.update(SolCam.java:70)
	at org.destinationsol.game.SolGame.update(SolGame.java:320)
	at org.destinationsol.SolApplication.update(SolApplication.java:146)
	at org.destinationsol.SolApplication.safeUpdate(SolApplication.java:121)
	at org.destinationsol.SolApplication.render(SolApplication.java:102)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:215)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:120)
```

Beyond that I was curious to perhaps jump into a starlane that crosses a sun, to see if that matters. I figure the transcended stuff relates to whether the player is in invincible mode, like when traveling in a star lane - going through a sun seemed a decent idea to test that!

Any other test cases of interest would be a good thing to add here :-)

I skipped most naming issues in the PR since I know there are lots of things being renamed and still to rename.

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
Adrijaned: Functionality and consistence with already existing code in this PR, renaming in the next PR

engine/src/main/java/org/destinationsol/game/Hero.java
Adrijaned: I kinda see no reason. Imagine you have a bottle with blue ball, and you try to take the red out. This is about the same.

engine/src/main/java/org/destinationsol/game/Hero.java
Adrijaned: This was actually a piece of code specifically requested by vampy. There was `!hero,isTranscendent()` on way too many places.

engine/src/main/java/org/destinationsol/game/Hero.java
Cervator: Are you trying to say we're in the Matrix? :D

(Speaking as somebody who still isn't totally sure what transcendent state even is it just stuck out to me)

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
eviltak: That may be the case, but since you have created this class yourself in this PR I think it better that the name be fixed before the PR is merged.

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -0,0 +1,205 @@
...
+import org.destinationsol.game.ship.FarShip;
+import org.destinationsol.game.ship.ShipAbility;
+import org.destinationsol.game.ship.SolShip;
+import org.destinationsol.game.ship.hulls.Hull;
+
+/**
+ * A wrapper class for the Hero, that handles the normal and transcendent ships transparently.
+ */
+public class Hero {
+    private SolShip hero;
```
eviltak: Better rename this to something else like `ship` to avoid confusion.

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -0,0 +1,205 @@
...
+        onlySolShipHero();
+        return hero.maybeUnequip(game, item, equip);
+    }
+
+    public boolean maybeUnequip(SolGame game, SolItem item, boolean secondarySlot, boolean equip) {
+        onlySolShipHero();
+        return hero.maybeUnequip(game, item, secondarySlot, equip);
+    }
+
+    private void onlySolShipHero() {
```
eviltak: Renaming this to something like `assertSolShipHero` or `assertNonTranscendent` (preferred) shall help clarify things a bit, because the name `onlySolShipHero` does not clearly state what the method does (which is asserting that the hero is not transcendent).

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -0,0 +1,205 @@
...
+
+    public boolean isNonTranscendent() {
+        return !isTranscendent;
+    }
+
+    public Pilot getPilot() {
+        return isTranscendent ? transcendentHeroShip.getPilot() : hero.getPilot();
+    }
+
+    public SolShip getHero() {
```
eviltak: Similar to comment above, renaming this to `getShip` with the accompanying field rename will remove ambiguity. `hero.getHero()` is a bit confusing, isn't it?

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -0,0 +1,205 @@
...
+        if (isTranscendent) {
+            throw new SolDescriptiveException("Something is trying to get a SolShip hero while the hero is Transcendent state.");
+        }
+        return hero;
+    }
+
+    public SolShip getHeroUnchecked() {
```
eviltak: Similar problem as above.

engine/src/main/java/org/destinationsol/game/ObjectManager.java
```diff
@@ -100,15 +100,20 @@ public void update(SolGame game) {
 
             if (o.shouldBeRemoved(game)) {
                 removeObjDelayed(o);
+                if (game.getHero().isAlive() && game.getHero().isNonTranscendent() && o == game.getHero().getHero()) {
```
eviltak: Might make sense extracting `game.getHero()` to a local variable. Also, `game.getHero().getHero()` is the perfect example of how confusing naming the hero's ship `hero` is!

engine/src/main/java/org/destinationsol/game/ObjectManager.java
```diff
@@ -100,15 +100,20 @@ public void update(SolGame game) {
...
+                    game.getHero().die();
+                }
                 continue;
             }
             if (isFar(o, camPos)) {
-                FarObj fo = o.toFarObj();
-                if (fo != null) {
-                    addFarObjNow(fo);
+                if (game.getHero().isAlive() && game.getHero().isNonTranscendent() && o != game.getHero().getHero()) {
```
eviltak: Same as above.

engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -193,7 +189,7 @@ private void createPlayer(String shipName) {
         String itemsStr = !respawnItems.isEmpty() ? "" : shipConfig.items;
 
         boolean giveAmmo = shipName != null && respawnItems.isEmpty();
-        hero = shipBuilder.buildNewFar(this, new Vector2(pos), null, 0, 0, pilot, itemsStr, hull, null, true, money, new TradeConfig(), giveAmmo).toObj(this);
+        hero = new Hero(shipBuilder.buildNewFar(this, new Vector2(pos), null, 0, 0, pilot, itemsStr, hull, null, true, money, new TradeConfig(), giveAmmo).toObj(this));
```
eviltak: Inherited issue: we should probably do something about the excessive arguments passed to the `buildNewFar` method. It does not look like `ShipBuilder` follows the Builder pattern to me.

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -0,0 +1,205 @@
...
+
+    public Hero(SolShip hero) {
+        this.hero = hero;
+        if (hero == null) {
+            throw new SolDescriptiveException("Something tries to create hero, when there is no ship linked to hiim. Please report this..");
+        }
+        isTranscendent = false;
+    }
+
+    public void toTranscendent(StarPort.Transcendent transcendentHero) {
```
eviltak: `toXYZ` methods are generally used to specify object conversions. If we choose to follow that convention, it might make sense to rename this method to something like `setTranscendent`.

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -0,0 +1,205 @@
...
+        this.transcendentHero = transcendentHero;
+        transcendentHeroShip = transcendentHero.getShip();
+        isTranscendent = true;
+    }
+
+    public void toSolShip(SolShip hero) {
```
eviltak: Similar to the comment above.

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
```diff
@@ -0,0 +1,56 @@
...
+        controls.add(controlUp);
+        controlDown = new SolUiControl(null, false, gameOptions.getKeyDown());
+        controls.add(controlDown);
+        controlLeft = new SolUiControl(null, false, gameOptions.getKeyLeft());
+        controls.add(controlLeft);
+        controlRight = new SolUiControl(null, false, gameOptions.getKeyRight());
+        controls.add(controlRight);
+    }
+
+    boolean isDown() {
```
eviltak: It might be better to rename this series of methods to something like `isDownControlOn` or `isDownControlEnabled` to be more descriptive.

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
```diff
@@ -0,0 +1,56 @@
...
+        return controlDown.isOn();
+    }
+
+    boolean isUp() {
```
eviltak: Same as above.

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
```diff
@@ -0,0 +1,56 @@
...
+        return controlUp.isOn();
+    }
+
+    boolean isLeft() {
```
eviltak: Same as above.

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
```diff
@@ -0,0 +1,56 @@
...
+        return controlLeft.isOn();
+    }
+
+    boolean isRight() {
```
eviltak: Same as above.

engine/src/main/java/org/destinationsol/game/screens/MainScreen.java
```diff
@@ -503,6 +511,22 @@ public boolean isAbility() {
         return shipControl.isAbility();
     }
 
+    public boolean isCameraUp(){
```
eviltak: Similar to a set of comments above: should be renamed to `isCameraUpControlEnabled` or similar.

engine/src/main/java/org/destinationsol/game/screens/MainScreen.java
```diff
@@ -503,6 +511,22 @@ public boolean isAbility() {
...
+        return cameraControl.isUp();
+    }
+
+    public boolean isCameraDown(){
```
eviltak: Same as above.

engine/src/main/java/org/destinationsol/game/screens/MainScreen.java
```diff
@@ -503,6 +511,22 @@ public boolean isAbility() {
...
+        return cameraControl.isDown();
+    }
+
+    public boolean isCameraLeft(){
```
eviltak: Same as above.

engine/src/main/java/org/destinationsol/game/screens/MainScreen.java
```diff
@@ -503,6 +511,22 @@ public boolean isAbility() {
...
+        return cameraControl.isLeft();
+    }
+
+    public boolean isCameraRight(){
```
eviltak: Same as above.

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -0,0 +1,205 @@
...
+
+    public ItemContainer getItemContainer() {
+        return isTranscendent ? transcendentHeroShip.getIc() : hero.getItemContainer();
+    }
+
+    public void die() {
+        isDead = true;
+    }
+
+    public boolean maybeEquip(SolGame game, SolItem item, boolean equip) {
```
eviltak: I understand that the underlying method in `SolShip` is named the same, but since we are building a wrapper around it, consider renaming the method to follow the `tryXYZ` convention _if_ it is decided upon.

Adrijaned: In DS, convention is `maybe`. Also, this methods behaves in two ways depending on how the `equip` flag is set. If it is unset, it checks, whether the item can even be equipped. Otherwise, it tries to equip it and reports result

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
Adrijaned: Here I am trying to stay consistent with the rest of the codebase.
(applies for the following 7 comments too)

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
Adrijaned: > consistence with already existing code

engine/src/main/java/org/destinationsol/game/Hero.java
Adrijaned: This is basically the same thing. Hero is just some kind of bottle, that may contain either SolShip hero or Transcendent hero. This exception says that you are trying to get the SolShip hero while there is Transcendent hero (and thus, no solhip hero) (previously, before the Hero, there was just null-checking in EVERY SINGLE PLACE that used `game.getHero()` (well, actually `g.getHero()`, unless `g` was gun, in which case it was `ga`))

engine/src/main/java/org/destinationsol/game/screens/ChangeShip.java
Adrijaned: I don't actually even remember that comment or the reason for it. Guess it might be pretty useless, since I don't see anything that hack-ish in there

engine/src/main/java/org/destinationsol/game/Hero.java
Adrijaned: This is kind of object conversion

GooeyHub:

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
vampcat: Here is the thing: If you make these small changes as separate commits in PRs that already touch a part of the codebase, they will easily get merged. However, if you keep putting them off and then send them out in a mega PR, they will inevitable take way more time.

So, I will *insist* you make the change here, even more so because you introduced this class in the first palce :)

engine/src/main/java/org/destinationsol/game/SolGame.java
vampcat: Yes, DestSol has way too many arguments in every single method. That being said, I kinda don't like the builder pattern for this particular case, since almost nothing is optional, and one of the biggest advantages of builder pattern is the ability to leave something out, so that the default value can be used for that parameter. Rather than have mandatory calls to 10 different methods, I would rather have them in the constructor.

engine/src/main/java/org/destinationsol/game/Hero.java
vampcat: The method name is perfectly fine for the existing DS codebase, true.

However, I would like to rename all "maybe" methods as well, since they make no sense to a newcomer.

engine/src/main/java/org/destinationsol/game/SolGame.java
eviltak: Assumed that the class followed the builder pattern because it has the prefix `Builder` :) Are those 10 arguments really unavoidable? I see a couple of nulls and 0s there. Can they be omitted?

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -67,18 +67,18 @@ public boolean isNonTranscendent() {
     }
 
     public Pilot getPilot() {
-        return isTranscendent ? transcendentHeroShip.getPilot() : hero.getPilot();
+        return isTranscendent ? transcendentHeroShip.getPilot() : shipHero.getPilot();
     }
 
-    public SolShip getHero() {
+    public SolShip getShipHero() {
```
vampcat: Not sure about `getShipHero`. How does `getShip` sound? Short, clear and to the point.

engine/src/main/java/org/destinationsol/game/screens/CameraKbControl.java
eviltak: Since this method has been newly introduced in this PR, I'd argue that naming should be fixed here itself. However, if you feel comfortable tackling this in another PR, I should have no problems.

eviltak:
I have similar thoughts on `SolDescriptiveError`. As an aside, I really like how descriptive your exception messages are, @Adrijaned!

Steampunkery:
Hmm. Well we could always just not merge code using a RuntimeException. In regard to the “please report this” bit, I vote we just remove the empty constructor and force them to use a description.

Adrijaned:
@Steampunkery There is no empty constructor, it just throws an exception if the exception being thrown is shorter than 10 chars. @vampcat @eviltak About c) - that can be easily removed. About b) - it signifies that error happened directly in DS codebase, and is not related to IO/libs/ongoing magnetic storms/broken operating system. About a) - have you heard there is a huge refactor/renaming PR on the way? And afterwards, we could just go the way of Ghost's suggestion.

engine/src/main/java/org/destinationsol/game/SolGame.java
vampcat: I haven't checked every single argument for ShipBuilder, I gave a general statement for most of the classes. I'll have a second look at ShipBuilder arguments though.

GooeyHub:

GooeyHub:

PR #249, opened 2018-02-23
Added first sane tests

Adrijaned:
Adds the first sane unit tests DestinationSol has ever seen! This is a historical moment, that will forever be remembered in history books and the other usual stuff for keeping things in!


GooeyHub:

Issue #251, opened 2018-03-04
Mercenaries from previously saved session carry over to New Game session


manas96:
### What you were trying to do
I tried playing a New Game after saving a previous session in which I had hired mercenaries.
### What actually happened
The mercenaries I had hired in the previous session spawned with me in the New Game session.
### How to reproduce

* Step 1 Ensure there are no previous saves (i.e. the files, "mercenaries.json" and "prevShip.ini" should not be present in /engine/src/main/resources).
* Step 2 Launch the game and make a new game session. (Play Game -> New Game -> Ok). You might need to select Loaded Imperial Large ship to do Step 3 faster.
* Step 3 Visit a planet and hire a bunch of mercenaries.
* Step 4 Exit the game. This session's data will be saved.
* Step 5 Launch game again and start another New Game session. You will spawn with the previously hired mercenaries.

### Game version
Um.. I'm not sure about this. I downloaded the develop branch.

vampcat:
Oooh, nice catch! Thanks for pointing it out :)
Would you like to try your hand at this, to get started with DestSol? :D

@Steampunkery I believe you're the culprit? :P

Steampunkery:
Yeah I did that. I can fix it if you want, @manas96, or I can show you the files and stuff to get started if you want to fix it.

manas96:
I fixed it. I've been having problems with my internet and git (my first time using git). I'll send in a pull request shortly.

Steampunkery:
How exactly does the implementation of your fix work? I just want to make sure it lines up with some slated changes.

PR #252, opened 2018-03-04
Added a boolean in SolApplication class (Fix for issue #251)

manas96:
# Description
A fix for issue #251 .
Added boolean "isNewGame" in SolApplication class.
It is set to TRUE when a new game is made, and FALSE when previous game
is continued.
If TRUE, mercenaries.json file is not loaded. If FALSE, mercenaries.json is loaded.

# Testing
Launch new game after hiring mercenaries in a previous session. Mercenaries won't spawn in the new game now.

### DIFF
--- a/engine/src/main/java/org/destinationsol/SolApplication.java
+++ b/engine/src/main/java/org/destinationsol/SolApplication.java
@@ -234,9 +234,15 @@ public class SolApplication implements ApplicationListener {
         return musicManager;
     }
 
-    public OggSoundManager getSoundManager() { return soundManager; }
+    public OggSoundManager getSoundManager() {
+        return soundManager;
+    }
 
-    public boolean getisNewGame() {return isNewGame;}
+    public boolean getIsNewGame() {
+        return isNewGame;
+    }
 
-    public void setisNewGame(boolean value) {isNewGame = value;}
+    public void setIsNewGame(boolean value) {
+        isNewGame = value;
+    }
 }
--- a/engine/src/main/java/org/destinationsol/SolApplication.java
+++ b/engine/src/main/java/org/destinationsol/SolApplication.java
@@ -60,8 +60,6 @@ public class SolApplication implements ApplicationListener {
     private float timeAccumulator = 0;
     private boolean isMobile;
 
-    private boolean isNewGame;
-
     public SolApplication() {
         // Initiate Box2D to make sure natives are loaded early enough
         Box2D.init();
@@ -184,8 +182,8 @@ public class SolApplication implements ApplicationListener {
         musicManager.playGameMusic(options);
     }
 
-    public void startNewGame(boolean tut, String shipName) {
-        solGame = new SolGame(this, shipName, tut, commonDrawer);
+    public void startNewGame(boolean isNewGame, boolean tut, String shipName) {
+        solGame = new SolGame(this, shipName, tut, isNewGame, commonDrawer);
         inputManager.setScreen(this, solGame.getScreens().mainScreen);
         musicManager.playGameMusic(options);
     }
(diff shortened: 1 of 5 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -232,7 +234,9 @@ public OggMusicManager getMusicManager() {
         return musicManager;
     }
 
-    public OggSoundManager getSoundManager() {
-        return soundManager;
-    }
+    public OggSoundManager getSoundManager() { return soundManager; }
```
NicholasBatesNZ: Why have you changed the formatting? Please revert

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -232,7 +234,9 @@ public OggMusicManager getMusicManager() {
...
+
+    public boolean getisNewGame() {return isNewGame;}
```
NicholasBatesNZ: Same here, revert formatting please

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -232,7 +234,9 @@ public OggMusicManager getMusicManager() {
...
+
+    public void setisNewGame(boolean value) {isNewGame = value;}
```
NicholasBatesNZ: And here :)

engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -236,6 +236,9 @@ private void createMercs() {
         if (!SaveManager.resourceExists(MERC_SAVE_FILE)) {
             return;
         }
+        if(solApplication.getisNewGame()){
```
NicholasBatesNZ: `if (...)` instead of `if(...)` (Add a space)

Cervator:
Hey there thanks for the good report and quick fix :-)

Is your IDE just auto-formatting stuff away from the more common project formats, perhaps? Try to set up and run Checkstyle. There should be a config file for it after you've run just about anything Gradle, but if you can't spot it run `gradlew extractConfig`

Triggering bot for review

@GooeyHub: add to whitelist

GooeyHub:

GooeyHub:

manas96:
Fixed formatting stuff suggested by @NicholasBatesNZ .
@Cervator - seems I didn't have CheckStyle installed in my IntelliJ. When installing it, I got an error similar to [this ](https://github.com/MovingBlocks/Terasology/issues/3113)issue. 
I think the DestSol gradle config needs to be updated with the same fix. (I fixed my issue by deleting the offending module in my local copy)

manas96:
I see your point, @vampcat . However, if I simply pass the variable through all invocations I will eventually have to know the value of `isNewGame` inside
` @Override
void updateCustom(...)` method of the `LoadingScreen` class. I can't simply pass in the boolean to this method, since it is overriding the `solUiScreen` interface method.
So I think the best solution is to move `isNewGame` inside the `LoadingScreen` class.
 If I understand this code correctly, this is the same reason why the `boolean loadTutorial` variable is also present in the `LoadingScreen` class.

vampcat:
I am fine with shifting the boolean to LoadingScreen class. I would like to remove these variables in the future, but since that requires a slightly more complicated refactor.. :D

GooeyHub:

Steampunkery:
Hmm. In the upcoming galaxy generation PR, there is an isNewGame variable. Do we hold off on this until then or just fix the merge conflicts when we merge galaxy generation?

vampcat:
Which class is that variable in?

manas96:
It's in the SolGame class in the Galaxy generation PR.

PR #258, opened 2018-03-28
Unit testing

Adrijaned:
Extremely ugly and non-efficient code, but looks like it works.

### DIFF
--- a/engine/build.gradle
+++ b/engine/build.gradle
@@ -11,6 +11,15 @@ dependencies {
     compile "org.terasology:gestalt-asset-core:6.0.0"
     compile "org.terasology:gestalt-entity-system:6.0.0"
 
+
+    //TODO inserted by someone who has no idea how the gradle thing works. Inspect whether necessary (for tests)
+    compile "com.badlogicgames.gdx:gdx-backend-lwjgl:$gdxVersion"
+    compile "com.badlogicgames.gdx:gdx-platform:$gdxVersion:natives-desktop"
+    compile "com.badlogicgames.gdx:gdx-box2d-platform:$gdxVersion:natives-desktop"
+    compile "com.badlogicgames.gdx:gdx-controllers-desktop:$gdxVersion"
+    compile "com.badlogicgames.gdx:gdx-controllers-platform:$gdxVersion:natives-desktop"
+
+
     // Test lib dependencies
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile group: 'org.mockito', name: 'mockito-core', version: '2.7.22'
--- a/engine/build.gradle
+++ b/engine/build.gradle
@@ -12,12 +12,12 @@ dependencies {
     compile "org.terasology:gestalt-entity-system:6.0.0"
 
 
-    //TODO inserted by someone who has no idea how the gradle thing works. Inspect whether necessary (for tests)
-    compile "com.badlogicgames.gdx:gdx-backend-lwjgl:$gdxVersion"
-    compile "com.badlogicgames.gdx:gdx-platform:$gdxVersion:natives-desktop"
-    compile "com.badlogicgames.gdx:gdx-box2d-platform:$gdxVersion:natives-desktop"
-    compile "com.badlogicgames.gdx:gdx-controllers-desktop:$gdxVersion"
-    compile "com.badlogicgames.gdx:gdx-controllers-platform:$gdxVersion:natives-desktop"
+    //TODO inserted by someone who has no idea how the gradle thing works. Inspect whether necessary (for tests). Copied rom desktop build.gradle. May break things with android (maybe)
+    testCompile "com.badlogicgames.gdx:gdx-backend-lwjgl:$gdxVersion"
+    testCompile "com.badlogicgames.gdx:gdx-platform:$gdxVersion:natives-desktop"
+    testCompile "com.badlogicgames.gdx:gdx-box2d-platform:$gdxVersion:natives-desktop"
+    testCompile "com.badlogicgames.gdx:gdx-controllers-desktop:$gdxVersion"
+    testCompile "com.badlogicgames.gdx:gdx-controllers-platform:$gdxVersion:natives-desktop"
 
 
     // Test lib dependencies
--- a/engine/src/test/java/org/destinationsol/IniReaderTest.java
+++ b/engine/src/test/java/org/destinationsol/IniReaderTest.java
@@ -31,7 +31,7 @@ public class IniReaderTest {
     @Before
     public void initIniReader() {
         String iniFileContents =
-                "# Full line comment\n" +
+                "# Full line comment\n" + // Each of these values is used exactly once in these tests, except for doubleRequestedKey
                         "partLineCommentKey = correctValue1 # Part line comment\n" +
                         "terrible key name        =          terrible key value\n" +
                         "# missingKey = wrongValue\n" +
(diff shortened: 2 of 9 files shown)


GooeyHub:

engine/src/test/java/org/destinationsol/IniReaderTest.java
```diff
@@ -51,7 +51,8 @@ public void initIniReader() {
                         "floatKey = 6\n" +
                         "anotherFloatKey = 7.3f\n" +
                         "invalidFloatKey = 8,6\n" +
-                        "anotherInvalidFloatKey = hi\n"; // Each of these values is used exactly once in these tests, except for doubleRequestedKey
+                        "anotherInvalidFloatKey = hi\n" +
+                        "UnicodeKey çáč🧝 = unicodevalue áśǵj́ḱĺóí⋄«»⋄⋄ǫő"; // Each of these values is used exactly once in these tests, except for doubleRequestedKey
```
Cervator: It would probably help to move this comment to the very top of the list. That way it doesn't wind up being caught in diffs like this (also, seems like it would get read sooner at the top anyway)

engine/build.gradle
```diff
@@ -11,6 +11,15 @@ dependencies {
     compile "org.terasology:gestalt-asset-core:6.0.0"
     compile "org.terasology:gestalt-entity-system:6.0.0"
 
+
+    //TODO inserted by someone who has no idea how the gradle thing works. Inspect whether necessary (for tests)
```
Cervator: Oh I see. This block was copied from the desktop `build.gradle` right? Because if you're running a unit test directly against the engine you wouldn't get the dependencies declared for the desktop.

If you only need them for tests you can use `testCompile` instead of `compile` and they'll only matter when execution is aimed at testing, not running normally. That might also sidestep any issues introduced for *Android* suddenly also depending directly on these for compilation. 

Although I don't know how it would impact a testing context for Android but ... there aren't any Android tests yet, anyway. So maybe just move to `testCompile` and update the comment/todo with some of these notes and a future consideration to update it further

engine/src/test/java/org/destinationsol/game/ShardTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+public class ShardTest {
+
+    private static ArrayList<Drawable> drawables;
+
+    private static void init() {
+        InitializationUtilities.init();
+    }
+
+    private static final SolObject constantShard;
```
Cervator: I thought this looked out of order, but I had to realize the Checkstyle config file was bad before I could confirm that. Although this is an interesting line of static stuffs :-)

This line should be up next to the other plain variable, then the methods get to go next

To fix Checkstyle (I'll submit a separate PR): change `codeMetrics group: 'org.terasology.config', name: 'codemetrics', version: '1.1.0', ext: 'zip'` to use `1.2.0` in the project root `build.gradle` then rerun `gradlew idea. Then pick to use the "Terasology Checkstyle" and run a scan against this file to confirm.

At that point this variable name will also register a style warning since it is a constant, so it should be in all caps :-)

Any point to not just have `InitializationUtilities.init();` inside the static init so the plain `init` method can simply be removed?

engine/src/test/java/org/destinationsol/game/ShardTest.java
```diff
@@ -0,0 +1,144 @@
...
+    @Test
+    public void receivesGravity() {
+        // When ship is shattered in gravity, its shards should fall to the ground
+        assertTrue(constantShard.receivesGravity());
+    }
+
+    @Test
+    public void receiveForce() {
+        // TODO I don't quite know what does this even do, so better leave this for sb else
+        fail("Implement this test!");
```
Cervator: Probably just leave a TODO with the name then, don't implement a failing tests - the tests will fail! :D

Alternatively you can mark the test as ignored while leaving the code in place, then it will just skip till somebody addresses the TODO

engine/src/test/java/org/destinationsol/testingUtilities/BodyUtilities.java
```diff
@@ -0,0 +1,47 @@
...
+import com.badlogic.gdx.math.Vector2;
+import com.badlogic.gdx.physics.box2d.Body;
+import com.badlogic.gdx.physics.box2d.BodyDef;
+import com.badlogic.gdx.physics.box2d.FixtureDef;
+import com.badlogic.gdx.physics.box2d.PolygonShape;
+import com.badlogic.gdx.physics.box2d.World;
+
+public final class BodyUtilities {
+
+    private static final World world = new World(new Vector2(0, 0), true);
```
Cervator: If constant then all caps. But this variable is only ever used once, so maybe just inline it later and remove the constant?

engine/src/test/java/org/destinationsol/testingUtilities/BodyUtilities.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.badlogic.gdx.math.MathUtils;
+import com.badlogic.gdx.math.Vector2;
+import com.badlogic.gdx.physics.box2d.Body;
+import com.badlogic.gdx.physics.box2d.BodyDef;
+import com.badlogic.gdx.physics.box2d.FixtureDef;
+import com.badlogic.gdx.physics.box2d.PolygonShape;
+import com.badlogic.gdx.physics.box2d.World;
+
+public final class BodyUtilities {
```
Cervator: Checkstyle suggests that a utility class shouldn't have a public or default constructor, so I guess you could throw in a 

```java
private BodyUtilities() {
    // empty private constructor for utility class
}
```

Or something like that to make Checkstyle happy

engine/src/test/java/org/destinationsol/testingUtilities/InitializationUtilities.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+package org.destinationsol.testingUtilities;
+
+import com.badlogic.gdx.backends.lwjgl.LwjglApplication;
+import org.destinationsol.SolApplication;
+import org.destinationsol.game.DebugOptions;
+import org.destinationsol.game.SolGame;
+
+public class InitializationUtilities {
```
Cervator: Same note about a private constructor for a utility class

engine/src/test/java/org/destinationsol/testingUtilities/InitializationUtilities.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public static SolGame game;
+    private static boolean initialized = false;
```
Cervator: The ` = false;` is redundant, default value (another Checkstyle nag)

engine/src/test/java/org/destinationsol/testingUtilities/InitializationUtilities.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public static void init() {
+        if (initialized) {
+            return;
+        }
+        initialized = true;
+        DebugOptions.DEV_ROOT_PATH = "engine/src/main/resources/";
+        final LwjglApplication application = new LwjglApplication(new SolApplication());
+        try {
+            Thread.sleep(10000L);
```
Cervator: Any specific reason for this sleep of 10s? Just hoping that'll normally be enough to launch the app until it is in a good enough shape for the tests to run? In that case you might want to add a class-level javadoc snippet explaining what the class is for and why the sleep duration is what it is

Actually, in general throw javadoc on just about anything you can think of ... :D

Adrijaned: Javadocs in progress for other stuff too :smile: . But will add

engine/src/test/java/org/destinationsol/testingUtilities/BodyUtilities.java
Adrijaned: It is an attempt at optimization... sort of? Also, if any other things requiring world are ever needed, this can be reused then.

GooeyHub:

Cervator:
Thanks for the updates :-) Checkstyle says all clear!

PR #264, opened 2018-04-01
Refactored ControlType to enum from integer constants..

Adrijaned:
Enum values are easier to inspect and harder to mess up than integers.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/SolGame.java
+++ b/engine/src/main/java/org/destinationsol/game/SolGame.java
@@ -19,17 +19,6 @@ import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.math.Vector2;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.FileReader;
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.io.UnsupportedEncodingException;
-import java.lang.reflect.Type;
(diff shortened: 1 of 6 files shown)


GooeyHub:

GooeyHub:

engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -62,10 +64,17 @@
...
+import java.io.File;
 import java.io.FileNotFoundException;
+import java.io.FileReader;
+import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.UnsupportedEncodingException;
+import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.HashMap;
+
```
NicholasBatesNZ: All of these imports are unused??

Adrijaned: All imports used.

Cervator: Looks like this might have been a branching quirk - they had been removed on purpose somewhere but may have still been present in this branch pre-merge. I had to resolve conflicts in this file anyway so I removed them again :-)

Cervator:
Merged locally, conflict resolved, tested, and pushed - thanks! :-)

The mystery imports were from prior merge issues, probably specifically the second commit - https://github.com/MovingBlocks/DestinationSol/pull/264/commits/90d9a7ae708dcda04354cf8178988f497d362f4a - when possible try to avoid using merges to update a PR branch. Either just leave it a bit out of date or use rebase if needed for conflicts and such.

PR #265, opened 2018-04-02
Respawn is now handled by Hero

Adrijaned:
Just wrote enough code for Hero so he can handle respawning. createPlayer() is untouched.


GooeyHub:

PR #268, opened 2018-04-02
Test and extract create player from SolGame

oniatus:
This PR moves player creation out of SolGame. The current class does still do a bit too much but at least it is tested and easier to understand.

Sidenotes for review:

- I added assertj for assertions, if we don't want to add this dependency I can replace the fluent assertions with old assertEquals.
- I had to remove final from HullConfig as the class alone is not instanciable and with final not able to mock. The cleaner solution would be to replace the class access with an interface and keep the backing up class final. This would involve chaning all calling places (~130) to the new interface which is something I did not want to include in this PR. The class as-is is still immutable however it could be extended and then break but that may be a bit esotheric depending how often we extend in destsol :)
- Can be reviewed commit-wise. I prepared the method for extraction first without changing the controlflow, then put it under test and refactored everything after the tests were in place without touching the tests again. Also did a short manual test session with respawn, reload and hiring a mercenary then reloading with him.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/PlayerCreator.java
+++ b/engine/src/main/java/org/destinationsol/game/PlayerCreator.java
@@ -47,7 +47,7 @@ class PlayerCreator {
 
         String items = findItems(shipConfig, respawnState);
 
-        boolean giveAmmo = isNewShip && respawnState.getRespawnItems().isEmpty();
+        boolean giveAmmo = shouldGiveAmmo(respawnState, isNewShip);
         Hero hero = new Hero(game.getShipBuilder().buildNewFar(game, new Vector2(position), null, 0, 0, pilot, items, hull, null, true, money, new TradeConfig(), giveAmmo).toObject(game));
         ItemContainer itemContainer = hero.getItemContainer();
         if (!respawnState.getRespawnItems().isEmpty()) {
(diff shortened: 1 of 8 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/game/PlayerCreator.java
```diff
@@ -0,0 +1,170 @@
+/*
+ * Copyright 2017 MovingBlocks
```
Adrijaned: Year

engine/src/main/java/org/destinationsol/game/RespawnState.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2017 MovingBlocks
```
Adrijaned: Year

engine/src/main/java/org/destinationsol/game/RespawnState.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+
+public class RespawnState {
+
+    private boolean isPlayerRespawned;
+    private final List<SolItem> respawnItems = new ArrayList<>();
+    private HullConfig respawnHull;
+    private float respawnMoney;
+
+
```
Adrijaned: Too much blank lines here, it seems to me

engine/src/test/java/org/destinationsol/game/PlayerCreatorTest.java
```diff
@@ -0,0 +1,374 @@
+/*
+ * Copyright 2017 MovingBlocks
```
Adrijaned: Year

GooeyHub:

Cervator:
I'd like to move forward with merging this commit-epic PR, any objections or review desires? :-)

GooeyHub:

NicholasBatesNZ:
I have resolved the merge conflicts and am merging this now :)

PR #261, opened 2018-03-30
Added some javadocs to SolObject

Adrijaned:
Descriptive and exhausting PR title.

PR #269, opened 2018-04-02
Added some javadocs to SolObject

Adrijaned:
re-opened #261

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/SolObject.java
+++ b/engine/src/main/java/org/destinationsol/game/SolObject.java
@@ -47,10 +47,10 @@ public interface SolObject {
     void update(SolGame game);
 
     /**
-     * Denotes whether the object should be removed as of the time of calling.
+     * Denotes whether the object, in its current state, should be removed.
      * <p>
-     * Object should usually be removed when its health reaches zero, or, for instance, if the object is projectile, it
-     * reaches its target. This method should not handle removal of object for optimization purposes when it gets too
+     * An object should usually be removed when its health reaches zero, or, for instance, if the object is projectile,
+     * it reaches its target. This method should not handle removal of object for optimization purposes when it gets too
      * far away, unless the object is specifically designed to exist only in close proximity of player, like some sort
      * of kinetic defense. Removal for optimization purposes is handled in {@link #toFarObject()}.
      *
@@ -64,21 +64,21 @@ public interface SolObject {
      * <p>
      * This method should handle things like throwing out {@link Loot} when ships are destroyed, or creating smaller
      * asteroids when a big {@link Asteroid} is destroyed. This method is also responsible for freeing its resources,
-     * when such is applicable. Note that this method is also called when an object is due to be removed for
-     * optimization purposes, and thus you need to explicitly check for {@link #shouldBeRemoved(SolGame)} if you have
-     * some code specific to destruction of object.
+     * when such is applicable. Please note that this method is called after removal due to {@link #shouldBeRemoved(SolGame)}
+     * returning true, as well as after removal after calling {@link #toFarObject()}. As such, if you have any code
+     * specific to removal of object due to its destruction, you need to explicitly check that the object is really
+     * being removed due to destruction.
      *
      * @param game Game this object belongs to.
      */
     void onRemove(SolGame game);
 
     /**
-     * Called whenever this object should receive damage.
+     * Inflicts object with certain amount of damage of specific type.
      * <p>
-     * Whenever the object is hit by some projectile or other sort of damage dealing thing, this method is called on the
-     * object to allow it to perform its damage handling. This usually means just subtracting the damage from object's
-     * health and playing a hit sound, but if the object has no health pool or should be otherwise indestructible, or
-     * invincible against some types of damage, this method can be freely left blank.
+     * This method usually just subtracts the damage from object's health and plays a hit sound, but if the object has
+     * no health pool or should be otherwise indestructible, or invincible against some types of damage, this method can
+     * be freely left blank.
      *
      * @param dmg      Damage the object receives.
      * @param game     Game this object belongs to.
@@ -88,17 +88,18 @@ public interface SolObject {
     void receiveDmg(float dmg, SolGame game, @Nullable Vector2 position, DmgType dmgType);
 
     /**
-     * Denotes whether this object is affected by gravity, this is usually constant.
+     * Denotes whether this object is affected by gravity.
      *
      * @return True if object is affected by gravity, false otherwise.
      */
     boolean receivesGravity();
 
     /**
-     * Applies external force to object.
+     * Applies external force to the object.
      * <p>
-     * This method usually just scales the {@code force} by objects mass if {@code acc} is set, and then passes the
-     * force to object's internal {@link Body}.
+     * If {@code acc} is true, the supplied {@code force} is treated as an acceleration and considered to not have been
+     * scaled by mass. This method usually just performs the scaling as needed, and then passes the force to the
+     * object's internal {@link Body}.
      *
      * @param force Force to apply to the object.
      * @param game  Game this object belongs to.
@@ -107,7 +108,7 @@ public interface SolObject {
     void receiveForce(Vector2 force, SolGame game, boolean acc);
 
     /**
-     * Returns object's position.
+     * Returns the object's position.
      * <p>
      * This method usually just returns the vector retrieved from object's internal {@link Body}.
      *
@@ -116,35 +117,36 @@ public interface SolObject {
     Vector2 getPosition();
 
     /**
-     * Called when an object is too far from player/camera, and is thus to be converted to {@link FarObject}.
+     * Converts this {@link SolObject} to a {@link FarObject}.
      * <p>
-     * This method is used for optimization of game's objects, by converting to their more resource friendly versions
-     * when they get too far from player/camera. The {@code FarObject} returned by this method should be usable to get
-     * copy of this object back in reverse. If null is returned by this method, this object is designated to be removed
-     * when it gets too far from player/camera, and instead of recreating from {@code FarObject}, to be generated again.
+     * This method is used to convert a {@link SolObject} to a more resource friendly {@link FarObject} when it is too
+     * far from the player/camera. The {@link FarObject} returned by this method must contain enough information to
+     * accurately recreate the {@link SolObject} when it gets in range again. If null is returned by this method, this
+     * object is designated to be removed when it gets too far from the player/camera and is regenerated from scratch
+     * (without persistence).
      *
      * @return FarObject representation of this object, or null if object is to be removed.
      */
     FarObject toFarObject();
 
     /**
-     * Returns list of all {@link Drawable Drawables} this object has assigned.
+     * Returns list of all {@link Drawable Drawables} this object has.
      *
      * @return List of drawables this object has.
      */
     List<Drawable> getDrawables();
 
     /**
-     * Returns object's angle.
+     * Returns object's angle, in degrees.
      * <p>
      * This method usually just returns the angle retrieved from object's internal {@link Body}.
      *
-     * @return Angle of this object, in degrees.
+     * @return Angle of this object.
      */
     float getAngle();
 
     /**
-     * Returns object's speed.
+     * Returns the object's speed.
      * <p>
      * This method usually just returns the vector retrieved from object's internal {@link Body}.
      *
@@ -168,7 +170,7 @@ public interface SolObject {
     void handleContact(SolObject other, float absImpulse, SolGame game, Vector2 collPos);
 
     /**
-     * Used for retrieval of object's debug string.
+     * Used for retrieval of object's debug string, comprising of list of its debug-useful properties.
      * <p>
      * Any kind of information can be used in the debug string, if you don't need/want to display any debug information,
      * you can freely have this method return null. To display debug strings in-game, set the flag {@link DebugOptions#OBJ_INFO}.
@@ -176,6 +178,7 @@ public interface SolObject {
      *
      * @return Debug string with information about the object.
      */
+    //TODO - improve/rework debug collection, displaying and related. See discussion in https://github.com/MovingBlocks/DestinationSol/pull/269 for more details
     default String toDebugString() {
         return null;
     }
--- a/engine/src/main/java/org/destinationsol/game/SolObject.java
+++ b/engine/src/main/java/org/destinationsol/game/SolObject.java
@@ -16,8 +16,11 @@
 package org.destinationsol.game;
 
 import com.badlogic.gdx.math.Vector2;
+import com.badlogic.gdx.physics.box2d.Body;
 import org.destinationsol.common.Nullable;
+import org.destinationsol.game.asteroid.Asteroid;
 import org.destinationsol.game.drawables.Drawable;
+import org.destinationsol.game.item.Loot;
 
 import java.util.List;
 
@@ -173,7 +176,9 @@ public interface SolObject {
      *
      * @return Debug string with information about the object.
      */
-    String toDebugString();
+    default String toDebugString() {
+        return null;
+    }
 
     /**
      * Denotes whether this object is meant to behave as if made from metallic material.
--- a/engine/src/main/java/org/destinationsol/game/SolObject.java
+++ b/engine/src/main/java/org/destinationsol/game/SolObject.java
@@ -16,8 +16,11 @@
 package org.destinationsol.game;
 
 import com.badlogic.gdx.math.Vector2;
+import com.badlogic.gdx.physics.box2d.Body;
 import org.destinationsol.common.Nullable;
+import org.destinationsol.game.asteroid.Asteroid;
 import org.destinationsol.game.drawables.Drawable;
+import org.destinationsol.game.item.Loot;
 
 import java.util.List;
 
--- a/engine/src/main/java/org/destinationsol/game/SolObject.java
+++ b/engine/src/main/java/org/destinationsol/game/SolObject.java
@@ -44,7 +44,7 @@ public interface SolObject {
     void update(SolGame game);
 
     /**
-     * Whether the object should be removed as of the time of calling.
+     * Denotes whether the object should be removed as of the time of calling.
      * <p>
      * Object should usually be removed when its health reaches zero, or, for instance, if the object is projectile, it
      * reaches its target. This method should not handle removal of object for optimization purposes when it gets too
@@ -88,7 +88,7 @@ public interface SolObject {
     }
 
     /**
-     * Whether this object is affected by gravity, this is usually constant.
+     * Denotes whether this object is affected by gravity, this is usually constant.
      *
      * @return True if object is affected by gravity, false otherwise.
      */
@@ -184,7 +184,7 @@ public interface SolObject {
     String toDebugString();
 
     /**
-     * Whether this object is meant to behave as if made from metallic material.
+     * Denotes whether this object is meant to behave as if made from metallic material.
      * <p>
      * This method is used mainly for choosing the sound to play after collision. Value true will play metallic sound,
      * false will play rocky sound. Value of null means no sound will be played.
@@ -197,7 +197,7 @@ public interface SolObject {
     }
 
     /**
-     * Whether this object has a {@link Body} associated with it.
+     * Denotes whether this object has a {@link Body} associated with it.
      * <p>
      * Generally, everything that can be touched has its {@code Body}.
      *
--- a/engine/src/main/java/org/destinationsol/game/SolObject.java
+++ b/engine/src/main/java/org/destinationsol/game/SolObject.java
@@ -21,36 +21,189 @@ import org.destinationsol.game.drawables.Drawable;
 
 import java.util.List;
 
+
+/**
+ * Generally speaking, SolObjects are all the game's foreground objects that are in the proximity of player/visible screen.
+ * <p>
+ * {@code SolObject} can be generally crashed into. Background objects aren't usually {@code SolObject}s, but they are
+ * rather just sprites rendered using some background manager. {@code SolObject} that gets too far from player/camera is
+ * usually transformed to {@link FarObject}, which has less runtime executable code and thus is faster and can be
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     *
+     * @param game Game this object belongs to.
+     */
     void onRemove(SolGame game);
 
+    /**
+     * Called whenever this object should receive damage.
+     * <p>
+     * Whenever the object is hit by some projectile or other sort of damage dealing thing, this method is called on the
+     * object to allow it to perform its damage handling. This usually means just subtracting the damage from object's
```
eviltak: Please remove the first sentence in this paragraph and note the addition of the missing "the" before "object's":
```
* <p>
* This usually means just subtracting the damage from the object's
```

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     * crosses for ships and stuff alike.
+     *
+     * @param game Game this object belongs to.
+     */
+    void update(SolGame game);
 
+    /**
+     * Denotes whether the object should be removed as of the time of calling.
+     * <p>
+     * Object should usually be removed when its health reaches zero, or, for instance, if the object is projectile, it
```
eviltak: _An_ object should usually...

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
 
+    /**
+     * Denotes whether this object is meant to behave as if made from metallic material.
+     * <p>
+     * This method is used mainly for choosing the sound to play after collision. Value true will play metallic sound,
+     * false will play rocky sound. Value of null means no sound will be played.
+     *
+     * @return Whether this object is made from metal.
+     */
+    //TODO allow for more complex collision sounds. Or even better, completely rework the material system.
```
eviltak: Unless I am missing something, this TODO item may belong to an implementor rather than the interface itself.

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     *
+     * @param dmg      Damage the object receives.
+     * @param game     Game this object belongs to.
+     * @param position Position the object was hit at, if hit by point-based damage. Null if not applicable, such as fire.
+     * @param dmgType  Type of the damage object receives.
+     */
     void receiveDmg(float dmg, SolGame game, @Nullable Vector2 position, DmgType dmgType);
 
+    /**
+     * Denotes whether this object is affected by gravity, this is usually constant.
```
eviltak: Please remove "`, this is usually constant`" -- unnecessary details.

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     *
+     * @return True if object is affected by gravity, false otherwise.
+     */
     boolean receivesGravity();
 
+    /**
+     * Applies external force to object.
```
eviltak: `* Applies an external force to the object.`

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     * <p>
+     * This method usually just scales the {@code force} by objects mass if {@code acc} is set, and then passes the
+     * force to object's internal {@link Body}.
```
eviltak: `* If {@code acc} is true, the supplied {@code force} is treated as an acceleration and considered to not have been scaled by mass.` We do not need `, and then passes the force to object's internal {@link Body}.`

Side note: I think that the `acc` parameter is redundant and just leads to more confusion. The method, after all, is called `receiveForce` and not `receiveAcceleration`! This may be something to address in a later PR.

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     *
+     * @param force Force to apply to the object.
+     * @param game  Game this object belongs to.
+     * @param acc   Whether this objects is accelerating by the force, such as in gravity.
+     */
     void receiveForce(Vector2 force, SolGame game, boolean acc);
 
+    /**
+     * Returns object's position.
```
eviltak: `* Returns the object's position.`

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     * <p>
+     * This method usually just returns the vector retrieved from object's internal {@link Body}.
+     *
+     * @return Position of the object.
+     */
     Vector2 getPosition();
 
+    /**
+     * Called when an object is too far from player/camera, and is thus to be converted to {@link FarObject}.
```
eviltak: `* Converts this {@link SolObject} to a {@link FarObject}.`

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     * <p>
+     * This method is used for optimization of game's objects, by converting to their more resource friendly versions
+     * when they get too far from player/camera. The {@code FarObject} returned by this method should be usable to get
+     * copy of this object back in reverse. If null is returned by this method, this object is designated to be removed
+     * when it gets too far from player/camera, and instead of recreating from {@code FarObject}, to be generated again.
```
eviltak: `* when it gets too far from the player/camera and has to be regenerated manually.`

I may be wrong here, so please correct me if the object does not actually have to be regenerated manually in this case.

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     * @return Position of the object.
+     */
     Vector2 getPosition();
 
+    /**
+     * Called when an object is too far from player/camera, and is thus to be converted to {@link FarObject}.
+     * <p>
+     * This method is used for optimization of game's objects, by converting to their more resource friendly versions
+     * when they get too far from player/camera. The {@code FarObject} returned by this method should be usable to get
+     * copy of this object back in reverse. If null is returned by this method, this object is designated to be removed
```
eviltak:
```
* This method is used to convert a {@link SolObject} to a more resource friendly {@link FarObject}
* when it is too far from the player/camera. The {@link FarObject} returned by this method must contain
* enough information to accurately recreate the {@link SolObject} when it is in range. If null is returned...
```

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     * when it gets too far from player/camera, and instead of recreating from {@code FarObject}, to be generated again.
+     *
+     * @return FarObject representation of this object, or null if object is to be removed.
+     */
     FarObject toFarObject();
 
+    /**
+     * Returns list of all {@link Drawable Drawables} this object has assigned.
```
eviltak: `* Returns a list of all {@link Drawable Drawables} this object has.`

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     *
+     * @return List of drawables this object has.
+     */
     List<Drawable> getDrawables();
 
+    /**
+     * Returns object's angle.
```
eviltak: `* Returns the object's angle, in degrees.`

Include all important information (like if the angle is in degrees or radians) in the summary sentence.

Side note: This method should, in my opinion, be renamed to something like `getRotation` or `getOrientation`, because the "angle" of an object makes no sense.

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     * <p>
+     * This method usually just returns the angle retrieved from object's internal {@link Body}.
+     *
+     * @return Angle of this object, in degrees.
+     */
     float getAngle();
 
+    /**
+     * Returns object's speed.
```
eviltak: `* Return the object's speed.`

Side note: This method should be renamed to `getVelocity` as soon as possible.

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     *
+     * @param other      Object this object touches.
+     * @param absImpulse Impulse the objects apply to each other.
+     * @param game       Game this object belongs to.
+     * @param collPos    Position where the two objects touch.
+     */
     void handleContact(SolObject other, float absImpulse, SolGame game, Vector2 collPos);
 
+    /**
+     * Used for retrieval of object's debug string.
```
eviltak: `* Returns a {@link String} containing a well-formatted list of debug information related to the state of the object.`

Side note: The `to` method name prefix conventionally specifies conversion. A good example is `toFarObject`. The usage of the prefix here is invalid. This method should be renamed to something like `dumpDebugInformation` (or better).

engine/src/main/java/org/destinationsol/game/SolObject.java
Adrijaned: There is about 10? 15? implementors of this interface. All of them should have better possibility to choose the sound to be played upon crashing with something.

engine/src/main/java/org/destinationsol/game/SolObject.java
```diff
@@ -16,41 +16,187 @@
...
+     */
+     boolean shouldBeRemoved(SolGame game);
+
+    /**
+     * Called whenever an object is due to be removed from game.
+     * <p>
+     * This method should handle things like throwing out {@link Loot} when ships are destroyed, or creating smaller
+     * asteroids when a big {@link Asteroid} is destroyed. This method is also responsible for freeing its resources,
+     * when such is applicable. Note that this method is also called when an object is due to be removed for
+     * optimization purposes, and thus you need to explicitly check for {@link #shouldBeRemoved(SolGame)} if you have
```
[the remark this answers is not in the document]
Adrijaned: Rewritten whole second part of paragraph.

engine/src/main/java/org/destinationsol/game/SolObject.java
Adrijaned: What does regenerated mmanually mean?

engine/src/main/java/org/destinationsol/game/SolObject.java
Adrijaned: Actually, this method should return something else than null only when really necessary. Only ships on AIPilots return non-null values from this method now. You can think of this method as of cleaner version of good old System.out.println(). When you need to learn contents of some variable for debugging, you put it in, afterwards, you remove it.

Cervator:
Bump - @Adrijaned do you have additional tweaks to add to this in response to the feedback?

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: This still looks like something that should belong to a specific implementor, since an interface contains absolutely no implementation information, and as such a TODO item regarding interface implementation should in my opinion not belong to the interface. Also, CodeFactor (rightly) complains that this comment is unexpected here. 

Anyhow, I don't see this as too big of a problem. It might make sense to move this comment to a better place so that CodeFactor stops complaining. Maybe move it _before_ the JavaDoc comment? I don't know if that will work, but it is worth a try. If nothing works, then we can leave it as is.

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: By "regenerated manually", I mean that the engine does not "optimize" this `SolObject` by converting it to a `FarObject`, it is removed and has to be created again. That is what I could gather from your description. Is that what really happens? If not, we might have to rephrase this to accurately reflect what will happen if `toFarObject` returns null.

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: I think it should be the other way around, actually. The main method description should include _everything_ -- what this method does, what kind of input it takes, and what kind of output it gives and a description of all possible edge cases. The params/returns section must also include information with the same depth of detail, but specific to the parameter/returned value. Necessary or important information may have to be duplicated in the method description and the params/returns description.

Specifically for this case, the `in degrees` detail is very, _very_ important. The summary sentence is the first (and probably only) thing I will read during a quick documentation look-up, and if it does not contain any and all important information about the method, I might create an easily avoidable logic error (by considering that the angle returned is in radians, not degrees, for instance).

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: While that may be true for the current code-base, I believe that there should _always_ be a way to retrieve debug data on the current state of the object, whether it is needed or not. There should also be a mechanism to easily view that debug information in-game, without having to close it, change code, and recompile the game. This will help immensely for things like user bug reports, and locating the problem in hard to replicate one-off bugs. I say this from personal experience :) This may be something to consider in the future, what do you think? Pinging @vampcat, @Cervator (and any others interested) for opinions.

engine/src/main/java/org/destinationsol/game/SolObject.java
Adrijaned: CodeFactor complaints of ANY //TODO

engine/src/main/java/org/destinationsol/game/SolObject.java
Adrijaned: Then why the word "manually"? With your suggested wording, it would make it seem to me that you have to add some special code somewhere to allow the object to be generated new in an area that has been already generated before.

engine/src/main/java/org/destinationsol/game/SolObject.java
Adrijaned: Usually, when you retrieve this value, you'll only be passing it to some other function expecting angle, or performing some crazy hard logic with it. In the first case, everything is in degrees in DS, in the second place, youo will probably still even look at someimplementations and ignore the javadoc whatsover, to make sure that the method really does what you think it does and what's written in the javadoc

engine/src/main/java/org/destinationsol/game/SolObject.java
Adrijaned: Imagine every single ship, every single bullet, every single shard ofeverything, every single planet tile had 10 lines of text below it. I don't think that will help with debugging much, I rather think that it will cause any debug information you may need to retrieve unreadable

Adrijaned:
I have fixes for the complained about in the review, that I have upvoted. Waiting with push till the discussion with @eviltak comes to an end.

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: That's not how debug information works. There should be mechanisms in place to output debug information just for the object that you want when you want it. From what I can understand, DestSol does not have any such mechanisms in place, but this is still something to discuss in the future. My recommendation for this javadoc has not changed.

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: Both the reasons you have outlined are the very reasons we need good, well detailed javadoc in the first place -- so that no assumptions have to be made, and time is not wasted by looking up the method source just to find out what the function does. I still do not have a concrete reason to redact my suggestion -- the change recommendation still stands.

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: I honestly do not think that is the case here -- the placement of the comment is causing the problem, not the fact that it is a TODO comment. Still, I see no point discussing this further, and leave it to your discretion to act further.

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: I am not sure I fully understand your last comment, but yes, I do want to make it seem that we have to add code to allow the object to be regenerated new in an area where it has been already generated before. As far as I can understand, that is what `and instead of recreating from {@code FarObject}, to be generated again` means. If this is not your intention/not actually what happens, you might want to consider rewording that line to reflect what actually happens. If you'd like, you can post a description of what actually happens and we can come up with something together.

Adrijaned: Objects that have their respective FarObjects are usually generated when the system is generated, and are just loaded/unloaded as needed. For other objects, there is a generateChunk() that is being called when player gets close to a chunk that has not been loaded, this function creates all SolObjects in the chunk without persistance, when the player gets too far again, these objects are destroyed, and chunk unloaded. Upon next coming close towards the chunk,, the generateChunk() is executed again.

engine/src/main/java/org/destinationsol/game/SolObject.java
Adrijaned: Javadocs should represent what the method does, not what it should do and be used to in future. For that, I might add a TODO, and I think I will if you do not opose that, possibly with your wording in a non-javadoc comment. I am trying to add TODOs to thinks that would not require much work to be fixed/improved for new contributors, and this possibly seems like the case. Is this option acceptable to you?

eviltak: I completely understand what you are getting at, but I still think we should change the method documentation summary statement.

First, the current summary statement provides no information whatsoever. The method name itself specifies as much information as the summary statement (in that the method returns a debug string) and makes it redundant. 

Furthermore, there are two basic things I expect of **good** debug information:
- It should be well formatted and easy to read.
- It should contain relevant information not easily found elsewhere, especially data which is not displayed in the UI.
Expectations may differ from person to person, but from what I have noticed, these two qualities  are generally considered as absolutely necessary for the debug information to be of any value.

In addition, when documenting _interface_ methods, there is no other option but to document what the method _should_ do. As such, the documentation is often considered as the specifications of the contract defined by the interface. The implementors may choose to deviate from the contract, but they must accept full responsibility. Specifically in this case, if the only non-null implementation in `AiPilot` (as you mentioned) does not return a "well-formatted list of debug information related to the state of the object", it is (according to my definition of ideal debug information) failure on part of the `AiPilot` implementation.

Do the reasons I listed above (specifically the first reason) help change your mind?

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: Adrijaned, thank you for that detailed, well written description! It certainly helps clear all confusion. Keeping that in mind, I accept your suggestion that we drop the word "manually", and change the clause to `when it gets too far from the player/camera and is regenerated from scratch (without persistence)`, which hopefully includes all the required details. What do you think?

engine/src/main/java/org/destinationsol/game/SolObject.java
Adrijaned: A bit of spin-off point: What about creating a DebugInfoCollector, to where depositing each objects debug info would be its own responsibility? Like, a single interface for depositing debug info of everything, not just solObjects, that can be completely turned off/on from game options?

Cervator: Catching back up - there was some chat on Slack on this topic :-)

Or rather, there seem to be 2-3 separate topics:

* More advanced debug capacity. Yep, everybody wants this, I believe. Out of scope for this PR but somebody should totally write up an issue. Also I fired up Factorio just to learn what `F4` did ;-) That looks like a great example to follow to allow enabling/disabling certain debug features 👍 
* What to actually do for this specific review comment on this PR right now. I largely agree with @eviltak's words of wisdom, but also hear @Adrijaned's desire to just float along sprinkling javadoc and todos rather than fix all the things :-)
  * The name does seem wrong, but yep that's pretty much out of scope for this PR and should get a nice TODO
  * The summary description (first sentence) - yeah, doesn't really add anything, although other stuff being added does, which is nice. Little tricky what to do with it since this setup may change a lot over the next few months (if anybody starts working on that area / deeper debug stuffs). Evil's example might be a good alternative for now. I'm honestly not sure how much intent we should bake into javadoc like that, especially with changes on the horizon. Maybe such potentially changing intent should be added in a TODO instead? "TODO: Clarify intent, current setup as of this comment is x, may be advisable to change to y in the future" ?

I'd suggest tweaking at the text just a bit, filing a follow-up issue (or two, or three), then moving on along with merging this soonish :-)

eviltak: @Cervator makes a good point --  we really need to begin to wrap things up, we have bigger fish to fry :)

In that regard, @Adrijaned, I request you to speed through this item, which is the last (as of now). I think we all agree there are quite a few future possibilities with the debug system, and we could choose to go in one of many directions. But, as far as I can envision, this method should largely stay the same for most options. Future changes notwithstanding, I really do feel that the summary statement should be changed to something more relevant and descriptive bearing in mind what this method is _expected_ to do. The exact changes I leave to your discretion, @Adrijaned. Feel free to use my proposed statement as a guideline.

Adrijaned: Somewhat edited/improved, added TODO about debug rework, with link to this PR

GooeyHub:

Cervator:
Thanks for all the nice discussion and javadoc :-)

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: I just realized that CodeFactor does complain of _any_ TODO it sees! That must have been the problem here.

Adrijaned: Told ya :smile:

eviltak:
Yep, sorry for doubting you! :D

engine/src/main/java/org/destinationsol/game/SolObject.java
eviltak: Yep, sorry for doubting you! :D

PR #270, opened 2018-04-05
Added some javadocs to FarObject

Adrijaned:


GooeyHub:

engine/src/main/java/org/destinationsol/game/FarObject.java
```diff
@@ -16,19 +16,85 @@
...
 import com.badlogic.gdx.math.Vector2;
+import com.badlogic.gdx.physics.box2d.Body;
 
+/**
+ * FarObject can be generally considered a {@link SolObject} simplified in terms of runtime code.
+ * <p>
+ * Generally speaking, every {@code SolObject} that gets too far away from player/camera, is transformed to a
+ * {@code FarObject} or deleted, with few specific exceptions. {@code FarObject}s then persist still in the game until
+ * the player/camera gets close to them again, at which point they are again transformed into their specific
+ * {@code SolObjects}. {@code FarObject}s generally have no body associated with them and do not inetract with the game much.
```
Cervator: Typo: "inetract"

On a side note I love how comparable this is to the Sector stuff in Terasology, if a fair bit simpler and easier to deal with :-)

engine/src/main/java/org/destinationsol/game/FarObject.java
```diff
@@ -16,19 +16,85 @@
...
     Vector2 getPosition();
 
+    /**
+     * Used for retrieval of object's debug string.
+     * <p>
+     * Any kind of information can be used in the debug string, if you don't need/want to display any debug information,
+     * you can freely have this method return null. To display debug strings in-game, set the flag {@link DebugOptions#OBJ_INFO}.
+     * These strings should then be rendered in the proximity of their objects.
+     *
+     * @return Debug string with information about the object.
```
Cervator: Traditionally I've found param and return statements usually start with the data type, so in this case maybe "String with debug information about the object" but that's pretty trivial and just a hunch, I don't know if it is a convention.

engine/src/main/java/org/destinationsol/game/FarObject.java
```diff
@@ -16,19 +16,85 @@
...
+     * <p>
+     * The created object should be by all terms similar or same as the {@code SolObject} that created this object.
+     *
+     * @param game Game this object belongs to.
+     * @return SolObject representation of this object.
+     */
     SolObject toObject(SolGame game);
 
+    /**
+     * Called on every game's frame, allows for handling of object-specific runtime needs.
```
Cervator: Not sure what the `'s` actually means here. I figure it is just "every game frame" or "every frame of the game"

PR #271, opened 2018-04-07
Oggsoundmanager - another (mainly) javadoc PR

Adrijaned:


GooeyHub:

GooeyHub:

GooeyHub:

engine/src/main/java/org/destinationsol/game/sound/OggSoundManager.java
```diff
@@ -114,66 +178,47 @@ public void play(SolGame game, PlayableSound playableSound, @Nullable Vector2 po
...
     public void drawDebug(GameDrawer drawer, SolGame game) {
         if (DebugOptions.SOUND_INFO) {
             debugHintDrawer.draw(drawer, game);
         }
     }
 
+    /**
+     * Updates drawer used in {@link #drawDebug(GameDrawer, SolGame)} and removes unnecessary objects.
+     *
+     * @param game Game currently at progress.
```
dave2s: Current game / Game _in_ progress    ?

engine/src/main/java/org/destinationsol/game/sound/OggSoundManager.java
```diff
@@ -89,12 +134,31 @@ public void play(SolGame game, PlayableSound playableSound, @Nullable Vector2 po
             position = source.getPosition();
         }
 
-        // Calculate the volume multiplier for the sound
-        float globalVolumeMultiplier = game.getCmp().getOptions().sfxVolumeMultiplier;
-        if (globalVolumeMultiplier == 0) {
```
dave2s: Why did you leave out 
`        if (globalVolumeMultiplier == 0) {
            return;
        }`
from the getVolume method? It still might speed up the process when Global Sound volume is zero? Or is it so that when global volume multiplier for sounds is zero all incoming requests are ignored at a higher level?

Adrijaned: Conclusion for future reference: It might speed it up a little bit, but it would make nearly unnoticeable difference, and the method isn't even called often much. (Except for beacons/engines that spam it like a hell, but these should be handled elsewhere).

GooeyHub:

engine/src/main/java/org/destinationsol/game/sound/OggSoundManager.java
```diff
@@ -64,19 +110,38 @@ public OggSound getSound(String path, float basePitch) {
...
+     * {@code position} must be specified.
      *
-     * @param position         position of a sound. If null, source.getPosition() will be used
-     * @param source           bearer of a sound. Must not be null for looped sounds
-     * @param volumeMultiplier multiplier for sound volume
+     * @param game     Game to play the sound in.
+     * @param sound    The sound to play
+     * @param position Position to play the sound at. If null, source.getPosition() will be used.
+     * @param source   Bearer of a sound. Must not be null for looped sounds or when {@code position} is null.
+     */
```
dave2s: I find this javadoc redundant? It might be sufficient for overloaded methods like this to just reffer the other one with full javadoc.

engine/src/main/java/org/destinationsol/game/sound/OggSoundManager.java
```diff
@@ -34,27 +34,73 @@
 import java.util.HashMap;
 import java.util.Map;
 
+/**
+ * This class represents the sound manager used in  DestinationSol. It is responsible for handling and playing all
```
Cervator: This is great info with some extra nice usage of javadoc tags, but overall it is a really dense wall of text, and that comes from somebody who loves walls of text! :-)

Split into some paragraphs if needed and consider a less conversational tone in favor of a more reference-based one. See https://github.com/Terasology/ModuleTestingEnvironment/blob/master/src/main/java/org/terasology/moduletestingenvironment/ModuleTestingEnvironment.java#L68 for a really good way to do a big description with a lot of nice formatting

An example reference-style rewrite might sound something like this:

"
Responsible for handling all game sounds. 

Loads sounds required from {@link Assets} in {@code .ogg} format in one of the asset-scanned directories. 

Optionally place a same-name text file as the ogg with a {@code .soundinfo} extension, containing:

<pre>
volume=x
loopTime=x
</pre>

{@code volume} specifies the default volume multiplier for the sound, as a {@code float} indicating a percentage
{@code loopTime} unset or zero will allow playing the sound again, even concurrently with itself. Greater than zero will only accept a replay after {@code float} x seconds from the same originator. 
"

(Note: I assume volume to be a percentage and loopTime being seconds, this may be wrong)

I believe that roughly conveys the same information but in fewer words with better formatting :-)

engine/src/main/java/org/destinationsol/game/sound/OggSoundManager.java
Cervator: Yeah I might include just notes on what is unique in that method and that it then forwards to another method variant. But also does indeed not hurt - unless in the future one has its javadoc updated but not the other :-)

Tricky call.

engine/src/main/java/org/destinationsol/game/sound/OggSoundManager.java
```diff
@@ -114,66 +209,59 @@ public void play(SolGame game, PlayableSound playableSound, @Nullable Vector2 po
...
-        Sound gdxSound = sound.getSound();
-        gdxSound.play(volume, pitch, 0);
+        return sound.getBaseVolume() * volumeMultiplier * distanceMultiplier * globalVolumeMultiplier;
     }
 
     /**
-     * Plays a sound. Source must not be null.
+     * Returns true when sound should not be played because of loop, false otherwise.
+     * <p>
+     * Sound should not be played when it's {@code loopTime > 0} and {@code loopTime} milliseconds have not yet passed
```
Cervator: `it's` vs `its` :-)

engine/src/main/java/org/destinationsol/game/sound/OggSoundManager.java
Adrijaned: I'd say leave it in. By checking at  the libGdx code, I found out it's the same over there.

engine/src/main/java/org/destinationsol/game/sound/OggSoundManager.java
Adrijaned: Milliseconds, or even better, real-time undependent frame-based time units, not seconds :) But I get the point.

GooeyHub:

Cervator:
Thanks for the updates - merged :-)

PR #242, opened 2018-01-29
Added List Modules Feature to the script

digitalripperynr:
This adds a list-modules feature to the groovy script. It shows all the modules available for download and all the modules that are already installed.

list-modules uses GitHub API to check all the repositories under the DestinationSol org and removes the system repos (eg. Splash Site) from the list.

Syntax : `groovyw module list-modules`

Previews :

PR #272, opened 2018-04-10
Re-designed Groovy Setup for DS

digitalripperynr:
# Description
> This PR aims to port the new and improved modular groovyw setup from Terasology to DestinationSol. For complete information on what is changed please check the original Terasology PR : [#3263](https://github.com/MovingBlocks/Terasology/pull/3263)

# Testing
> You may see usage information using the command `groovyw usage` . This overhaul does not change much of the functionality and is only meant to be a backend re-design.

# Notes
> Although I've tested all the commands before commiting the work, it might be possible that some things would have gone wrong during the port. Please test the code atleast once before merging.

> This includes, improves, and closes #242 .

> Terasology PRs [#3283](https://github.com/MovingBlocks/Terasology/pull/3283) and [#3284](https://github.com/MovingBlocks/Terasology/pull/3284) are also included in this update.

# Tweaks
> I've changed the module info template extension from `module.info` to `module.json` because the modules now use json info files iirc.

# Credits
> Original Re-design - @digitalripperynr
> Concept - @mjuvekar7 and @Cervator 
> Improvements & Overhaul - @Cervator 
> Github API Fix - @mwesolowski


GooeyHub:

mjuvekar7:
It does actually make sense to fetch modules without recursing. An example scenario where this would be useful is if someone wants to make minor tweaks to a module (eg changing the README) but doesn't really want to use the module in-game. It would be really annoying if you had to download a while bunch of dependencies just to change a README

mjuvekar7:
As you said, when modules get more complex and start having a large number
of dependencies, forcing recurses is a bad idea, because it can cause a lot
of unnecessary downloads. Doesn't hurt to keep recurse and regular fetch
different.

digitalripperynr:
Should I add it now or keep it for later ?

Also, recurse command was not present in original groovy script . Although I think that was probably because dependencies in DS groovy were added only recently.

mjuvekar7:
Add it now if you can.

Cervator:
Haven't checked the details yet, but one bit of guidance: try to make the differences between the setup here and the one for Terasology as minimal as possible - even if `x` isn't needed for DS yet. That way it is less work over time to copy changes across from one repo to another. 

So don't leave something out just because it may not currently make sense for DS. That just sets us up for more of a headache keeping the two sets of utility in sync later :-)

Cervator:
Thanks for hooking this up! Tests out great. I tweaked a slight bit more to make it even closer to the Terasology files. One script had lost a layer of indentation and another was still more module centric than "item" generic. But I'm quite happy to see how near identical the two sets of scripts can be :-)

I left a couple DS specific notes, like how only the module logic is present, and not quite the same just yet. Facades etc are not hooked up yet but would be very close. For instance "desktop" is the DestSol "PC Facade", there is one for Android, and even a broken one for GWT. We might just need a little bit of tweaking and the Android setup could be added that way!

PR #277, opened 2018-04-24
Javadoc for SolMath

Adrijaned:
Some tiny improvements, too


GooeyHub:

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -15,49 +15,72 @@
...
 
 /**
- * A class with helpful mathematical functions
+ * This class presents a collection of handy mathematical as well as programmatical functions, for working with angles,
+ * distances, speeds, vectors.
  */
 public class SolMath {
-    private static Logger logger = LoggerFactory.getLogger(SolMath.class);
+    /**
+     * The pi number constant.
```
Cervator: It puzzles me a bit why this class even wraps these `MathUtils` constants and such. Which came up on Slack too. Fine to add javadoc to them but it seems silly as opposed to just replacing all the actual calls to use `MathUtils` directly.

However, that's existing code, so I'm fine leaving it that way until somebody wants to take the time to throw all the silly stuff away, carefully :-)

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -15,49 +15,72 @@
...
+     *
+     * @param b boolean to convert
+     * @return int representation of the boolean
+     */
     public static int toInt(boolean b) {
         return b ? 1 : -1;
     }
 
+    /**
+     * Returns absolute value of a float.
```
Cervator: Wat. This is literally a method that already exists. I don't know why this is here. But at least now it is documented! :-)

Same story as earlier, this is kinda progress until we throw it all away so 👍

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -67,146 +90,185 @@ public static float approach(float src, float dst, float speed) {
...
+     * Check that an object has returned all the vectors he has borrowed.
+     *
+     * @param o Object to blame upon the not returning of vectors.
+     * @throws AssertionError When not all of the vectors were returned
      */
-    public static float sin(float a) {
-        return MathUtils.sinDeg(a);
+    public static void checkVectorsTaken(Object o) {
+        if (SolMath.vectorsTaken != 0) {
+            throw new AssertionError("vectors " + SolMath.vectorsTaken + ", blame on " + o);
```
Cervator: I'm not sure that my thinking is correct here or there's some odd convention I'm unaware of, but I've seen this pattern in the code of using `AssertionError` to outright crash the game over some (usually) relatively tiny issue rather than simply log an angry error message then continue. That looks really strange to me and I've taken it out a few places where I could make sense of some better form of error handling to harden the game.

I would suggest we try to avoid them unless somebody can illustrate why they're a good thing :-)

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -259,67 +318,97 @@ public static Vector2 distVec(Vector2 from, Vector2 to) {
...
+     * @return Length of the projection
      */
     public static float project(Vector2 v, float angle) {
         float angleDiff = angle - SolMath.angle(v);
-        return v.len() * cos(angleDiff);
+        return v.len() * MathUtils.cosDeg(angleDiff);
     }
 
+    /**
+     * Computes a square root of number.
```
Cervator: More "wat" - why is this method a thing :/

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -259,67 +318,97 @@ public static Vector2 distVec(Vector2 from, Vector2 to) {
...
+     * @param radius   Radius of the arc
+     * @return Calculated angle, in degrees
+     */
     public static float arcToAngle(float hordeLen, float radius) {
-        return 180 * hordeLen / (PI * radius);
+        return radDeg * hordeLen / radius;
     }
 
+    /**
+     * Calculate the length hypotenuse from the two other sides in triangle.
```
Cervator: This also seems to already exist in `Math` :/

engine/src/main/java/org/destinationsol/common/SolMath.java
Adrijaned: This thing is here to avoid memory leaks of sort? If the vectors were not returned, they couldn't be used anymore, but nor would they be collected by garbage collector, since reference to them would still exist in the pool. Also, I was thinking about doing something similar to these vector pools to TS too, but there are several problems that can't be overcome easy enough, especially when the code hasn't been made with those... bound objects in mind. One of them is freeing the resources, as you need to be sure that no more refs to the object exist elsewhere before freeing them, and who knows what goes on in WorldGens and such?

Adrijaned: Also, error messages might go unnoticed, while crashes not. And even a tiny issue might grow into a big well-hidden bug if it goes unnoticed for a while

vampcat:
Actually, can you wait a bit before merging? I have a ton of unposted comments on this, I'll complete them today.

engine/src/main/java/org/destinationsol/common/SolMath.java
Cervator: Talked on Slack a bit, my comment was purely about the `AssertionError` causing a crash, the vector pool concept is definitely interesting :-)

I'd like to see us more toward using the CrashReporter and telemetry more to highlight errors, rather than risk full crashes over even small things. But yeah until we get there I don't mind leaving these in for better visibility. They do suck for hard to reproduce errors though (user keeps hitting mysterious crashes without obvious explanations) 

Will wait for @vampcat in any case on this PR since he commented about having pending stuff :)

engine/src/main/java/org/destinationsol/common/SolMath.java
vampcat: I'll agree that this extra layer of abstraction seems weird. Plus, the javadocs adds no real information either - everyone reading this code will know what `public static float PI` represents.

Either way, we can deal with this later.

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -15,49 +15,72 @@
...
 public class SolMath {
-    private static Logger logger = LoggerFactory.getLogger(SolMath.class);
+    /**
+     * The pi number constant.
+     */
     public static final float PI = MathUtils.PI;
+    /**
+     * Multiply radians by this number to convert them to degrees.
+     */
     public static float radDeg = MathUtils.radDeg;
```
vampcat: `Multiply radians by this number to convert them to degrees.` -> `Conversion factor to transform radians to degrees.`

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -15,49 +15,72 @@
...
+    /**
+     * Multiply degrees by this number to convert them to radians.
```
vampcat: `Conversion factor to transform degrees to radians.`

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -15,49 +15,72 @@
...
+    private static int vectorsTaken;
 
+    /**
+     * Converts boolean to integer, where {@code true} equals {@code 1} and {@code false} equals {@code -1}.
+     *
+     * @param b boolean to convert
+     * @return int representation of the boolean
+     */
     public static int toInt(boolean b) {
         return b ? 1 : -1;
```
vampcat: Not your work, but converting false to -1 seems weird.

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -67,146 +90,185 @@ public static float approach(float src, float dst, float speed) {
...
             return dst;
         }
-        return diff > 0 ? src + speed : src - speed;
+        float diff = norm(dst - src);
+        return norm(diff > 0 ? src + speed : src - speed);
     }
 
     /**
-     * Normalizes the angle
+     * Normalizes the angle, ie puts it in the range [-179.999, 180].
```
vampcat: Technically, it puts it in the range `(-180, 180]`.

engine/src/main/java/org/destinationsol/game/StarPort.java
```diff
@@ -73,7 +73,7 @@ private static void blip(SolGame game, SolShip ship) {
     @Bound
     public static Vector2 getDesiredPosition(Planet from, Planet to, boolean precise) {
         Vector2 fromPosition = from.getPosition();
-        float angle = SolMath.angle(fromPosition, to.getPosition(), precise);
+        float angle = SolMath.angle(fromPosition, to.getPosition());
```
vampcat: Why did we convert all non-precise calls to precise ones?

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -15,49 +15,72 @@
...
+    /**
+     * Multiply radians by this number to convert them to degrees.
+     */
     public static float radDeg = MathUtils.radDeg;
+    /**
+     * Multiply degrees by this number to convert them to radians.
+     */
     public static float degRad = MathUtils.degRad;
-    public static Pool<Vector2> vs = new Pool<Vector2>() {
+    /**
```
vampcat: Can we add newline before all javadocs?

engine/src/main/java/org/destinationsol/common/SolMath.java
```diff
@@ -15,49 +15,72 @@
...
 import com.badlogic.gdx.math.Vector2;
 import com.badlogic.gdx.utils.JsonValue;
 import com.badlogic.gdx.utils.Pool;
+import org.destinationsol.Const;
+
+import java.util.ArrayList;
 
 /**
- * A class with helpful mathematical functions
+ * This class presents a collection of handy mathematical as well as programmatical functions, for working with angles,
```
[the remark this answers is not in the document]
Adrijaned: `readV2()`, `readV2List()` and such. There is a TODO about removing these tho, IIRC

engine/src/main/java/org/destinationsol/common/SolMath.java
Adrijaned: Aren't javadoc for explaining what it is and how to use it? It might be just me, but given your version, I'd still have to look at the actual contents of the var to understand what it is and how to use it. Also, since the variable is basically just copy-pasted from libGDX, so is the javadoc of mine.

engine/src/main/java/org/destinationsol/common/SolMath.java
Adrijaned: It is used for some direction, I believe? Like `toInt(moveForward) * movementConst`?

engine/src/main/java/org/destinationsol/common/SolMath.java
Adrijaned: Technically, in czech at least, it puts it in range `(-180, 180>`. Didn't know the right notation, so I had to help myself in any way I could.

engine/src/main/java/org/destinationsol/game/StarPort.java
Adrijaned: There were two cases of precise/non-precise functions. In one case, the precise executed _exactly_ the same code, in the other, it was told in javadoc a several layers upper, that the precise version is quite slower, while the non-precise one has error of on average 0.01 degrees.

GooeyHub:

Adrijaned:
In case of eventually merging this, @vampcat , if you ever happen to be floating around, I'd first suggest squashing all the commits in one first, now that I'm looking at the commit history :smile:

Cervator:
Random PR merge day today - review one get one free!

Seems like this has hung around long enough. Glanced over the diff again, saw a few final tweaks, and tested out in-game a while. No particular issues noted, although the enemies in a maze seemed mildly directionally challenged when it came to acquiring targets. Bit odd but probably has been an issue for a while.

Not too worried about commits. Shipping it!

PR #279, opened 2018-05-01
Remove TutorialManager access from SolGame

oniatus:
Moves the TutorialManager to context or direct access. I also cleaned up some tutorial-mode checks.

Manual tested:
1. Tutorial mode still starts and tutorial works
2. Tutorial mode does not override existing save

### DIFF
--- a/engine/src/main/java/org/destinationsol/SolApplication.java
+++ b/engine/src/main/java/org/destinationsol/SolApplication.java
@@ -89,7 +89,7 @@ public class SolApplication implements ApplicationListener {
 
         musicManager = new OggMusicManager();
         soundManager = new OggSoundManager(context);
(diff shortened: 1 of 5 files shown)


GooeyHub:

NicholasBatesNZ:
This looks really good, however when I start the tutorial and exit immediately, the tutorial elements are still present on continuation of a regular game, and they don't go away when the instructions on them are followed. @Cervator has confirmed that this does not happen in develop.

msteiger:
Code changes look ok to me. Maybe there are more places in the code base that need to be changed? I guess that we should check again after the code conflicts are resolved.

Adrijaned:
Closing with #423 existing now

PR #281, opened 2018-05-01
Refactor system updates

oniatus:
Puts system updates under a common interface which can later on be used for an ES adaption but does also work on its own. I also started to remove the timeStep state from SolGame, a second batch in a follow-up PR would be to remove the access from SolObjects.


GooeyHub:

PR #284, opened 2018-05-08
Static context sketch

NicholasBatesNZ:
#### This is by no means a mergeable PR, it is simply the result of some experimentation.

By altering the new Context work by @Adrijaned to make the methods static, we can heavily reduce the amount of parameters in a great number of methods. Each class can simply grab the object that they require (providing it has been added to the Context Map) and use it accordingly. I only did this in a very small number of places (SolGame + the classes that already used the existing Context) but there is potentially allot more that can be done.


GooeyHub:

eviltak:
I completely understand the motivation behind this PR, but while making `Context` static and accessible anywhere does seem like a good idea at first glance, it defeats the whole purpose of using a "context".

First, a `Context` is not _meant_ to be accessible everywhere. The contents of a `Context` are typically meant to be read-only, that is they will not change throughout the lifetime of the program. The contents of a `Context` are meant to be set once only and shared everywhere. Making the `Context` accessible (and modifiable) at any time and any place defeats this purpose.

Secondly (and more importantly), a `Context` is, deriving from the literal meaning of "context", the _environment_ in which an object is _created_. This environment may *differ* from object to object, and is fully accessible only during _initialization_ of objects. Depending on the type of object, it may have to be created in a more secure and restricted `Context`. For instance, external code (like modules) may be given access to a select few systems/classes via a `ModuleContext`, whereas internal engine classes can be given full access to all systems present in a more complete `EngineContext`. During initialization, objects can take from the environment (context) and add to it -- the environment is immutable and inaccessible otherwise.

Adrijaned:
I have to agree with @eviltak , and just note, that the two contexts in talk will probably be `appContext`, which will stay the same during the whole runtime, and `gameContet`, which will be created new on every new game. Also, the `gameContext` would probably have `appContext` as _parent_, so that `appContext` classes can be requested from game-specific places. Also, having specific `moduleContext` seems useless to me, since we want the modules to be able to do the same as game, and security should be handled by module sandboxing from gestalt (given we can make it work).

Adrijaned:
Static access to the Context things can then be done through InjectionHelper.

Adrijaned:
Sorry, wrong button. Why do they have `Close` right next to `Comment`? :disappointed:

GooeyHub:

Cervator:
Good effort and thinking here, but yeah this is something we kinda-sorta did with Terasology long ago, learned from, and then the next step after that was what became `Context` and later yet `FlexibleConfig` and so on. It is a deep deep topic with a lot of background and architecture. 

Fun to explore, lots to learn, and don't feel bad for thinking in the same way we have in the past :-) See `CoreRegistry` vs `Context` over in the Terasology codebase (and in the forum) for some of the history.

@eviltak and @vampcat have both been heavily involved in the topic. It is still in progress but I think we're getting closer to the "next generation" and I hope that's something we can extract and use the same way in both Terasology and DestSol

NicholasBatesNZ:
Coolio, this was just some fun experimentation of mine, I thought I'd show it here in case we wanted something similar to it in the meanwhile before we jump into ECS

PR #286, opened 2018-05-11
Continuous pilot input

eviltak:
# Motivation
The current input system (served through `Pilot`s) is a set of boolean values -- `isUp` for acceleration, `isLeft` for turning left, and `isRight` for turning right. Due to the discrete nature of this boolean input system, it is quite difficult to bring about accuracy in controlling ships. This PR introduces a _continuous_ input system, i.e. an input system that tells the `ShipEngine` _exactly_ how much to accelerate and _exactly_ how much to turn. In essence, the previous system would tell the ship to simply "turn right"; the current system tells it to "turn right by 37.867 degrees".

The benefits of this input system may not be that noticeable during normal keyboard and mouse gameplay, but are clearly evident when using controllers and observing AI controlled characters (throttle support TODO). The throttle system will also play a big role in a near future PR-posal involving throttle velocity regulation in the engine.

# Changes
## Additions
- `Pilot.getThrottle` and `getOrientation` with corresponding implementations
- `Mover.getThrottle` and `getDesiredOrientation`
- `Shooter.isRotatingTowardsTarget` and `getDesiredOrientation`

## Modifications
- New angular acceleration based ship rotating system in `ShipEngine` which gradually rotates the ship towards the desired orientation as reported by the `Pilot`
- `Mover`, `Shooter` and implementations of `ShipUiControl` now return continuous throttle and orientation inputs instead of booleans

## Removals
- `Pilot.isUp`. `isLeft` and `isRight`
- `Mover.isUp`, `isLeft`, `isRight`
- `Shooter.isLeft`, `isRight`
- `MainScreen.isUp`, `isLeft`, `isRight`

# Usage
Use the up and down buttons/axis to increase and decrease the throttle, respectively. For a button based input system, throttle is increased by a hard-coded increment of 10%. For controllers, the throttle is calculated by calculating the extent to which the joystick is pushed. In other words, throttle for controllers is equal to the magnitude of the heading vector. Orientation controls are as they were previously, except for controllers. The up-down and left-right controller axes are combined to form the heading vector of the ship. If the two axes are on the same joystick, the ship moves where the joystick is pointed (unless the input is inverted).

# TODO
- [x] Use joystick directions to provide continuous throttle and orientation input
- [ ] Calculate intermediate throttle inputs (rather than just 0 and 1) in `Mover`
- [ ] Modulate engine sound and particle emitter intensity with changes in throttle
- [ ] Visualize target orientation for button/keyboard orientation input

# Future TODO
- [ ] Add screen slider (or something of the like) for easier throttle input
- [x] Change keyboard controls to increase ease of use
- [ ] Any UI side changes (potentially for the controller, since two axes will be used for orientation)

Thoughts, improvements, criticisms and suggestions welcome.


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

Great PR write-up, thanks for that :-)

I've tried it out locally and love the idea of it as an option, especially if it is tied to something player controllable, ideally via ship fitting: affix a throttle control module if you want this type of control

It might be a little ahead of its time in that sense since we haven't even broken ground on customizing ships in-game. Maybe it should be the only option for a given sub-set of ships? That might be one way to avoid having to teach the AI how to use this system before eventually merging it: use it solely for a player controlled ship.

One potential challenge: how to maintain multiple control schemes in parallel in a sane way with the codebase the way it is right now.

GooeyHub:

eviltak:
@Cervator you bring up a very interesting point regarding the uncertainty of future customization. While it may seem like some (future) ships might be better off without the system because we don't know how they are going to be designed, I feel that this control system would actually make it easier to generalize ship controls and bring a good amount of certainty to them.

The current system simply asks ships to move forward (accelerate), turn left or turn right. This looks very simple and certain at first glance, but it is actually very ambiguous: a ship (controlled by a `ShipEngine`, possibly interchangable/customizable via external modules in the future) may choose to interpret that information in a myriad of ways, like turning 10, 20, 45, or 90 degrees to the left or right; or simply adding a rotational impulse (momentum) to the ship in the specified direction so that it maintains its rotational velocity and keeps rotating in that direction (because space). However, when a ship is given a target orientation, it does not leave room for ambiguity -- no matter how, at the end of the maneuver, the ship _has_ to end up in the target orientation. The player (and the AI controller) is assured that the ship will point in the direction where it needs to be pointed. Also, the controls become highly accurate, the benefits of which have already been mentioned in the PR description.

Throttle is a slightly less useful option as compared to target orientation, but it should allow for future introductions with functionality like #288, basically giving more control and ease of use to the player/AI controller.

OvermindDL1:
Then of course a ship could be created, say the generic saucer-ufo ship, that does not rotate but instead only gets a thrust in the direction of input.

eviltak:
Great example, @OvermindDL1, which serves to show just how flexible this method really is without losing any unambiguity. A `Pilot` gives the `ShipEngine` input in the form of throttle and target orientation, but it's the `ShipEngine` that does all the work. It is up to the engine to decide whether to simply point the ship and exert thrust in the target direction; turn around to point in the diametrically opposite direction and exert thrust so that the ship _slows down_ when a button is pressed; or rotate the engine slip ring (or rotate nothing at all) _without_ rotating the ship and exert thrust in the target direction, like your UFO example.

NicholasBatesNZ:
@eviltak this has grown a bit old and stale unfortunately. What's the status? Do you plan to continue this (I see some TODO boxes have not been checked)? Or should we close it now

eviltak:
@NicholasBatesNZ I would definitely love to keep this going. I shall take a look at rebasing and updating the PR as soon as possible.

NicholasBatesNZ:
Sorry for the inconvenience.

PR #287, opened 2018-05-11
Vecpool

Adrijaned:
Vector pool liberal implementation, as discussed in #destination-sol as well as #architecture, on Slack, also contains bugfix.

### DIFF
--- a/engine/src/main/java/org/destinationsol/common/Vector2Pool.java
+++ b/engine/src/main/java/org/destinationsol/common/Vector2Pool.java
@@ -21,15 +21,16 @@ public class Vector2Pool {
     private static final int MAX_VECTORS = 100;
     private static final Vector2[] EBP = new Vector2[MAX_VECTORS];
     private static int esp = -1;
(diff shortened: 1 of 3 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/common/Vector2Pool.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+package org.destinationsol.common;
+
+import com.badlogic.gdx.math.Vector2;
+
+public class Vector2Pool {
+    private static final int MAX_VECTORS = 100;
+    private static final Vector2[] EBP = new Vector2[MAX_VECTORS];
+    private static int esp = -1;
```
eviltak: This isn't assembly, we can change these names to something better! :smiley:

Adrijaned:
Closing as unneeded and only a proof of concept, also there is now more-fledged PR of the same thing in https://github.com/MovingBlocks/Terasology/pull/3386

PR #288, opened 2018-05-11
Target velocity based engine with throttling

eviltak:
# Motivation
Inspired by the controls of the game Galaxy on Fire 2, this PR introduces a target velocity based `ShipEngine`, which constrains the ship speed instead of accelerating indefinitely (till the max velocity cap, rather) like the current approach. The ship speed constraint is determined by scaling the maximum possible speed with the throttle, making slowing down and fine maneuvering _much_ easier. With this system, stopping the ship is as simple as setting the throttle to zero.

This PR is based on the continuous input system introduced in #286, and should be merged after it. This PR is pretty small and changes `ShipEngine` and `Engine` only. The extraneous changes in this PR which have already introduced in #286 should disappear after #286 is merged.

# Changes
- The engine now accelerates only to maintain constant speed
- The acceleration property in `Engine` has been increased to make the controls feel more responsive, in line with the intentions of this PR.

# Usage
Increase or decrease the throttle to see a corresponding increase or decrease in ship speed.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/input/AiPilot.java
+++ b/engine/src/main/java/org/destinationsol/game/input/AiPilot.java
@@ -52,6 +52,9 @@ public class AiPilot implements Pilot {
     private PlanetBind myPlanetBind;
     private float myReEquipAwait;
 
+    private float orientation;
+    private float throttle;
+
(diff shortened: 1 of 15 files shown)


GooeyHub:

PR #295, opened 2018-05-26
Created scheduler and timeProvider

Adrijaned:
Improvement upon time handling in DS. should play more nicely with Context and stuff. TODO in some later PR - make something actually USE the timeProvider/scheduler.

# FAQ
> Why is there the scheduler?  

I thought I would need it for the SloMo. Learned I was wrong. Too lazy to remove it; might be still useful for something.

> Why did you make this PR?  

Although it may not seem so, this is an important step on the way towards ship interiors. Don't ask me for the reasoning tho. It probably made sense to me once, but now I just know it's important. That's it.


GooeyHub:

NicholasBatesNZ:
The time provider makes sense, but why can't we just use the [LibGDX Timer](https://libgdx.badlogicgames.com/ci/nightlies/docs/api/com/badlogic/gdx/utils/Timer.html) for scheduling tasks?

Adrijaned:
:flushed: Because I didn't know such class exists...

GooeyHub:

Cervator:
Looks reasonable, but the conflict is somewhat non-trivial and might need some prodding to resolve

Adrijaned:
Closing as old without any activity and with conflicts, will probably do it one more time sometimes in the future as still needed (but now we need to lower our PR amount a bit)

PR #297, opened 2018-06-02
In-game Console!

Adrijaned:
The code is ugly. Don't even look at the git history. On the bright side, it works!
How to test in-game: Press \` while game is in progress. Type. Only command currently supported is `echo`. Press ESC to close again


GooeyHub:

engine/src/main/java/org/destinationsol/game/Console.java
```diff
@@ -0,0 +1,198 @@
...
+import org.destinationsol.ui.SolInputManager;
+import org.destinationsol.ui.SolUiControl;
+import org.destinationsol.ui.SolUiScreen;
+import org.destinationsol.ui.UiDrawer;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class Console implements SolUiScreen {
```
Cervator: Javadoc pretty please :-)

engine/src/main/java/org/destinationsol/game/Console.java
```diff
@@ -0,0 +1,198 @@
...
+        font = Assets.getFont("engine:main").getBitmapFont();
+        exitControl = new SolUiControl(null, true, Input.Keys.ESCAPE);
+        controls = Collections.singletonList(exitControl);
+        linesOfOutput = new ArrayList<>();
+        inputLine = new StringBuilder();
+        println("Welcome to the world of Destination Sol! Your journey begins!");
+        setInputHandler(new ShellInputHandler());
+    }
+
+    public void println(String s) {
```
Cervator: Perhaps also javadoc here. Curiously have you checked out the Terasology console classes for some of the potential there, like auto-completion, documented parameters, etc?

engine/src/main/java/org/destinationsol/game/Console.java
```diff
@@ -0,0 +1,198 @@
...
+        drawInputLine(uiDrawer, textX);
+
+    }
+
+    private void drawInputLine(UiDrawer uiDrawer, float textX) {
+        StringBuilder stringBuilder = new StringBuilder();
+        int width = 0;
+        for (char c : inputLine.reverse().toString().toCharArray()) {
+            try {
+                width += (c == ' ' ? 3 : 1) * font.getData().getGlyph(c).width;
```
Cervator: This line is kinda scary. May be better to be less condensed but more readily understandable.

engine/src/main/java/org/destinationsol/game/Console.java
```diff
@@ -0,0 +1,198 @@
...
+                if (width > MAX_WIDTH_OF_LINE) {
+                    break;
+                }
+                stringBuilder.append(c);
+            } catch (NullPointerException e) {
+                inputLine.deleteCharAt(0); // when there is character that can't be handled by font, it'll always be the last (first in reversed), since if it was anywhere elsewhere, it'd already be removed earlier.
```
Cervator: This sort of exception handling looks like a poor practice (exception handling of runtime issues should not be part of the control flow or something). How about detecting the nulls and reacting to them in the method itself instead? Then hopefully there'll be a good place to put better comments there rather than as a sort of dangling long comment in a catch block that's easy to ignore

engine/src/main/java/org/destinationsol/game/Console.java
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+        uiDrawer.drawString(stringBuilder.reverse().toString(), textX, getLineY(20.666f), 0.5f, UiDrawer.TextAlignment.LEFT, false, Color.WHITE);
+        inputLine.reverse();
+    }
+
+    private float getLineY(float line) {
+        return TOP_LEFT.y + 2 * FRAME_WIDTH + line * UiDrawer.FONT_SIZE * 0.5f * 1.8f;
```
Cervator: Magic numbers. Use commented constants instead please :-)

engine/src/main/java/org/destinationsol/game/console/ConsoleInputHandler.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.game.console;
+
+import org.destinationsol.game.Console;
+
+public interface ConsoleInputHandler {
```
Cervator: Javadoc :-)

engine/src/main/java/org/destinationsol/game/screens/MainScreen.java
```diff
@@ -141,6 +143,9 @@ public MainScreen(float resolutionRatio, RightPaneLayout rightPaneLayout, Contex
         controls.add(pauseControl);
         cameraControl = new CameraKeyboardControl(gameOptions, controls);
 
+        consoleControl = new SolUiControl(null, true, Input.Keys.GRAVE);
```
Cervator: Keep in mind `GRAVE` can be a tricky hot key due to different keyboard formats. In Terasology we default to *both* `GRAVE` and `F1` to better catch a wider array of keyboard formats.

engine/src/main/java/org/destinationsol/ui/SolInputManager.java
```diff
@@ -147,7 +147,7 @@ public void addScreen(SolApplication solApplication, SolUiScreen screen) {
         screen.onAdd(solApplication);
     }
 
-    private void removeScreen(SolUiScreen screen, SolApplication solApplication) {
+    public void removeScreen(SolUiScreen screen, SolApplication solApplication) {
```
Cervator: Is this access modification potentially problematic at all? Could other things just start randomly removing screens, by chance or on purpose, or is it badly needed for something?

GooeyHub:

engine/src/main/java/org/destinationsol/ui/SolInputManager.java
Adrijaned: Not actually sure what it does in this PR. Probably was just testing some stuff and forgot to revert.

engine/src/main/java/org/destinationsol/game/screens/MainScreen.java
Sigma-One: This is true, I'm constantly annoyed by games having the console on it, as I do not have the key

GooeyHub:

engine/src/main/java/org/destinationsol/game/screens/MainScreen.java
```diff
@@ -144,7 +145,9 @@ public MainScreen(float resolutionRatio, RightPaneLayout rightPaneLayout, Contex
         cameraControl = new CameraKeyboardControl(gameOptions, controls);
 
         consoleControl = new SolUiControl(null, true, Input.Keys.GRAVE);
+        consoleControl2 = new SolUiControl(null, true, Input.Keys.F1);
```
NicholasBatesNZ: Can we call this `consoleControlF1` instead? And the other `consoleControlGrave`. Or something similar so we know what's going on down the line

GooeyHub:

engine/src/main/java/org/destinationsol/game/Console.java
```diff
@@ -0,0 +1,309 @@
...
+                if (width > MAX_WIDTH_OF_LINE) {
+                    break;
+                }
+                stringBuilder.append(c);
+            } catch (NullPointerException e) {
+                inputLine.deleteCharAt(0); // when there is character that can't be handled by font, it'll always be the last (first in reversed), since if it was anywhere elsewhere, it'd already be removed earlier.
+            }
+        }
+        // 20.666f - magic constant, change if console is ever resized.
+        uiDrawer.drawString(stringBuilder.reverse().toString(), textX, getLineY(20.666f), 0.5f, UiDrawer.TextAlignment.LEFT, false, Color.WHITE);
```
jellysnake: Magic constants should be in a field on the class ;)

engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -48,8 +50,6 @@
 import org.destinationsol.game.ship.ShipBuilder;
 import org.destinationsol.game.ship.SloMo;
 import org.destinationsol.game.ship.hulls.HullConfig;
-import org.destinationsol.assets.audio.OggSoundManager;
-import org.destinationsol.assets.audio.SpecialSounds;
```
jellysnake: The changes in this file seen irrelevant to those in the rest of the PR and should probably be removed

engine/src/main/java/org/destinationsol/ui/SolInputManager.java
```diff
@@ -26,11 +26,11 @@
 import org.destinationsol.GameOptions;
 import org.destinationsol.SolApplication;
 import org.destinationsol.assets.Assets;
+import org.destinationsol.assets.audio.OggSoundManager;
 import org.destinationsol.assets.audio.PlayableSound;
 import org.destinationsol.common.SolColor;
 import org.destinationsol.common.SolMath;
 import org.destinationsol.game.SolGame;
-import org.destinationsol.assets.audio.OggSoundManager;
```
jellysnake: Again seems to be the only change in this class.

engine/src/main/java/org/destinationsol/game/Console.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /**
+     * Draws the separator between the area for user input and the area to render console output into
+     *
+     * @param uiDrawer Drawer to draw to.
+     */
+    private void drawTextEntrySeparator(UiDrawer uiDrawer) {
+        // 20.333f - magic constant, change is console is ever resized
+        uiDrawer.drawLine(TOP_LEFT.x + 2 * FRAME_WIDTH, getLineY(20.333f), 0, (BOTTOM_RIGHT.x - TOP_LEFT.x) - 4 * FRAME_WIDTH, Color.WHITE);
```
jellysnake: Magic constants should be a field. Applies to all of them

engine/src/main/java/org/destinationsol/game/Console.java
Cervator: Bump - any resolution on this? :-)

GooeyHub:

GooeyHub:

Cervator:
Got poked on Slack, checked this out again, tested locally, looks and plays good 👍 

I'm sure we can improve the console further (and maybe share it with Terasology) but this is good to have. Nice job :-)

Issue #181, opened 2017-12-04
Creating distribution of the developer version; crash because resource not found


Avalancs:
### What you were trying to do
I wanted to package the latest development version with the 'dist' and 'distZip' Gradle targets to be able to distribute it

### What actually happened
When the game is started it will crash with the error message missing src/main/resources/icon.png.

### How to reproduce
- run the distZip gradle task
- open the <project directory>\desktop\build\distributions\app folder
- start sol.exe

My discoveries:
- I see that instead of the old version's res folder now the assets are packaged inside the jar file.
- The path src/main/resources/icon.png can mean the program is actually looking for the file instead of using the classloader?
- icon.png, debugOptions.ini, log4j.properties, module.info, prevShip.ini and settings.ini seems to be added twice to <project directory>\desktop\build\distributions\app\libs\sol.jar. I had a similar problem once where gradle's distribution task had two rules which packaged the same files. I will investigate further tomorrow.

### Log details and game version
commit 9efb5ac0cc985546ac2fc211c64dd600b4ca92bd

Cervator:
Thanks for the details! Yes this appeared after the GSOC project and we haven't had a chance yet to figure out what isn't going to the right place. Probably isn't too tricky a fix so if you can figure it out that'd be excellent :-)

Avalancs:
Found the reason for the duplicated files in the jar. It's engine/build.gradle
```
sourceSets {
    // Main is almost everything - the true engine
    main {
        java {
            output.classesDir 'build/classes'
        }
        // This makes the resources (assets, 3d wizardry includes, etc) also go to build/classes
        output.resourcesDir 'build/classes'
}
```

Since the project has the 'java' plugin enabled it automatically has build/classes as source set hence this declaration makes it compress the files twice. Removing it fixes the duplication issue, but does not fix the original problem when starting the game, so I will continue with the investigation

Avalancs:
Okay, so the problem is that when finding resources there are a lot of if-elses that add engine/src/main/resources to the beginning which work well in the development environment (since the files are physically there) but after packaging the files are not there but in the jar.

For engine resources which are included in the jar, the gfx FileHandles had to be set to FileType=classpath, and the *FileFormat classes had to be written that if they don't find a resource as a physical file then they should look inside the classpath. The problem is that when trying to load an asset the AbstractAssetFileFormat class does not know if the asset comes from a core module from inside a jar file, or a physical file path.

[...]

PR #259, opened 2018-03-29
classpathModule Fix

manas96:
# Description
You can now use methods like `ModuleEnvironment`'s `getTypesAnnotatedWith()`.

# Testing
1. Create SampleAnnotation.java:
```
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.TYPE)
@interface SampleAnnotation{

}
```
2. Pick any class, and annotate that class with above annotation. Example:
I picked SolApplication:
```
@SampleAnnotation
public class SolApplication implements ApplicationListener {
...
}
```
3. At the end of `ModuleManager.java` 's constructor , put this line:
`Iterable<Class<?>> list = environment.getTypesAnnotatedWith(SampleAnnotation.class);`
Also put a breakpoint on this line.

4. Debug program, and examine value of `list` . It will contain names of all classes annotated with `SampleAnnotation` . 
You will also get a bunch of Debug messages from `Reflections.java` (See below)
# Outstanding Work

However, this has one issue: ALL files in /out/production/engine  (.json, .ogg, .png etc.) are scanned  for annotations. This causes a bunch of scan error messages in the console because `Reflections.java` is trying to scan .json / .ogg / .png files for annotations.

The scanner needs to scan only .class files.

PR #300, opened 2018-06-17
Fix distribution creation

NicholasBatesNZ:
This PR contains the work that @Avalancs did in relation to issue #181 however due to changes that have since occurred in the codebase, it no longer functions fully :/

### Current Status
Crashing

### How to test
- run `gradlew build` (this includes the key task `distZip` and is necessary to update the game binaries)
- extract `desktop/build/distributions/DestinationSol.zip`
- run `sol.exe` (or `sol.sh`)

### Milestones
- [x] Fix `missing src/main/resources/icon.png` crash
- [ ] Fix `ModuleManager.printAvailableModules` NPE crash
- [ ] Fix `whatever else crops up` crash
- [ ] Do something with any `TODO:`s in the code


GooeyHub:

NicholasBatesNZ:
Concerning the ModuleManager NPE: by adding a whole bunch of `textfile.println()`s (I'm almost certain there's a better way but ¯\_(ツ)_/¯ ) I was able to identify that it fails at [`ModuleManager.loadEnvironment(...)`](https://github.com/NicholasBatesNZ/DestinationSol/blob/1cb6b3038e45d715fd71c4bf435b41c9e06d7dce/engine/src/main/java/org/destinationsol/ModuleManager.java#L74)
I also tried incorporating the changes from #259 but that didn't seem to help

Cervator:
Alright lots to sort through on this one :-)

IIRC the old goal was to simply let the zip work again. Yet since then we also have more module stuff going on. You think dealing with both at once is needed? It may well be.

In testing locally some notes:

* Loads of debug logging indicating first search for engine assets happens under `modules/engine` - should that be a special case somewhere so the engine path is checked first?
* On getting myself killed then respawning I hit an NPE `at org.destinationsol.game.PlayerCreator.grantPlayerMoney(PlayerCreator.java:152)` - didn't happen if I swapped back to `develop` although for some reason I respawned in the loaded imperial large while first I had picked the regular little wimpy imperial
* Looks like a unit test or two may be problematic when run via Gradle. Seemed OK in IntelliJ. Log snippet below. Looks like the print thing already mentioned. Same thing that crashes when run as `exe`
* A `.tar` file generates as part of the distribution plugin I figure (as part of `gradlew build distZip` but *not* from a `gradlew clean test` then `gradlew distZip` huh). Do we really need it? Seems like it double-includes something anyway, it is a bit more than twice the size of the zip

Log snippet for failing test:

```
org.destinationsol.game.ShardTest STANDARD_ERROR
    java.io.IOException: Could not resolve module metadata for module at C:\Dev\DestinationSol\Git\integrate\DestinationSol\engine\build\classes\main
        at org.terasology.module.ModuleFactory.createPathModule(ModuleFactory.java:189)
        at org.terasology.module.ModuleFactory.createModule(ModuleFactory.java:168)
        at org.destinationsol.ModuleManager.<init>(ModuleManager.java:52)
        at org.destinationsol.SolApplication.create(SolApplication.java:85)
        at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:143)
        at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:120)
    Exception in thread "LWJGL Application" java.lang.NullPointerException
        at org.destinationsol.ModuleManager.printAvailableModules(ModuleManager.java:84)
        at org.destinationsol.SolApplication.create(SolApplication.java:88)
        at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:143)
        at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:120)
```

More tinkering to come this is just my round one ...

GooeyHub:

BenjaminAmos:
You get a NullPointerException when you run the application with the embedded jre but using the system jre, e.g. by running the command `java -jar libs/solDesktop.jar`, does not throw an exception. The game then starts correctly and appears to function with any further issues.

The crash has a stacktrace relating to a missing java.nio.file provider. The missing provider appears to be the [Zip Filesystem Provider](https://docs.oracle.com/javase/8/docs/technotes/guides/io/fsp/zipfilesystemprovider.html).

The bundled jre folders, which are used by the launch scripts (sol.exe, sol.sh, sol64.sh and solOSX.sh), are missing the provider jar file needed to run the application. The jar file is called `zipfs.jar` and by copying it from another jre distribution (i.e. the one installed locally on the machine) into the `lwjre[OS]/lib/ext` folder the game should run.

This may be resolved by #413, as the jre distribution will be removed in that pull request. I have obtained a copy of the zip provider jar from the `com.github.marschall:zipfilesystem-standalone` maven package, which I have added as a dependency of the engine project (since the engine uses gestalt-module). The package appears to be copy of the original source, abeit with a changed namespace. You can find the source [here](https://github.com/marschall/zipfilesystem-standalone).

This pull request appears to run perfectly after the `zipfs` jar is added.

You can find the changes that I've made in my fork, on the [fix-distribution](https://github.com/BenjaminAmos/DestinationSol/tree/fix-distribution) branch. Should I make a pull request with these changes? The changes are alike to those provided by @Avalancs but they also add some code to package the modules with the distribution and do not utilise the gradle distribution plugin.

In terms of packaging the modules with the distribution, there appears to be some caveats. The LibGDX library only accepts FileHandle objects for assets, whilst Gestalt returns InputStream objects to access the file contents directly. To overcome this, the FileHandle class is sub-classed to provide access to a backing AssetDataFile, which is provided by Gestalt. This allows the files to be read from jar archives (zip archives do not work at the moment). The only asset type not using this method is the BitmapFont, which access the files directly and relies on a navigable file system being present.

Please let me know if there is a better solution to this, or if any more work needs to be done.

**Note**: this has not been tested on Android.

Adrijaned:
Hey @BenjaminAmos, nice to hear from you again! I'd say that if your code accomplishes something, open the PR. Someone who knows a bit more about our gradle stuff than me should look at it tho. One think I noticed from my quick scourge over your code is that unless there is something special happening for core, it should not have it's own `build.gradle` file in git, and all of the modules should have this file gitignored, even if only for security reasons (not implying that DestSol currently has any sort of security procedures in place :grin:). For the same reason on each `./gradlew build` or wherever that piece of code is handled the `build.gradle` should probably be just deleted and recopied - but that's for another time, and probably not of any importance.

Cervator:
Superseded by #415 - closing as complete/merged via inspiration :-)

PR #303, opened 2018-07-25
IRC chat

Adrijaned:
Console command `chat`, console is opened by `F1`, opens an IRC connection to freenode, room `##destination-sol-players` (might want to change that), don't you dare connect to that room with normal IRC client and sending non-standard ascii range characters or you will crash all the clients currently connected. Might be broken even in other ways.


Adrijaned:
/HELP prints help message, when in chat

GooeyHub:

Adrijaned:
Not to be merged into engine, more of a showcase of what can be done using console, this would also make for a nice module content

PR #307, opened 2018-08-13
Graphics and UI overhaul

vampcat:
Building upon Adri's work, this PR makes the ui and content drawn on screen resolution-agnostic.

PR #311, opened 2018-09-19
Graphics overhaul, better git history

Adrijaned:
Hopefully replaces #307

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/SolCam.java
+++ b/engine/src/main/java/org/destinationsol/game/SolCam.java
@@ -20,12 +20,15 @@ import com.badlogic.gdx.graphics.OrthographicCamera;
 import com.badlogic.gdx.math.Matrix4;
 import com.badlogic.gdx.math.Vector2;
 import com.badlogic.gdx.math.Vector3;
+import com.badlogic.gdx.utils.viewport.ScreenViewport;
 import org.destinationsol.Const;
+import org.destinationsol.SolApplication;
 import org.destinationsol.common.SolColor;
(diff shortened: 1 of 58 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/game/SolCam.java
```diff
@@ -110,6 +117,8 @@ public void updateMap(SolGame game) {
         myZoom = SolMath.approach(myZoom, desiredZoom, ZOOM_CHG_SPD * ts);
         applyZoom(game.getMapDrawer());
         myCam.update();
+        viewport.update(Gdx.graphics.getWidth(), -Gdx.graphics.getHeight());
```
vampcat: Use displayDimensions here (and in the next line) instead of using Gdx.graphics.whatever

engine/src/main/java/org/destinationsol/game/sound/OggMusicManager.java
```diff
@@ -0,0 +1,95 @@
+/*
```
vampcat: Why are we messing around with sound files? (There are 5 sounds files here that shouldn't be)

engine/src/main/java/org/destinationsol/game/SolCam.java
Cervator: Fixed, I think, with #321

Cervator:
Now hopefully replaced with #321 ! Conflicts scary. Apply tractor + plow to conflicts. Then sleep.

PR #315, opened 2018-10-01
Display helpfuller err msg on malformed ship JSON

Adrijaned:
During GCI, lots of students will probably make lots of new ships, which will be broken in lots of not-so-original ways. With this PR merged in, the error displayed then might at least hint some of them what's wrong with their ships.


GooeyHub:

PR #317, opened 2018-10-02
fix regression introduced by #281: red target crosshair now draws again

### LINKED ISSUE #306: Aiming cross not being displayed
When approaching enemy ships, your ship's weapon auto-targets the closest ship up to some distance. The auto-targeting works, but not showing the visual indicator of it ([here to be seen](https://github.com/MovingBlocks/DestinationSol/blob/develop/engine/src/main/resources/assets/textures/ui/uiBeaconAttack.png))

Avalancs:
fixes #306 

# Root cause
the order of the systems update() calls changed, and mountDetectDrawer needs to go before beaconHandler for the red target crosshair to work.

# How to test
Start the game, use the first laser, or any weapon that can rotate on the player ship, and approach an enemy. You should see a red crosshair behind the enemy station/ship

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/SolGame.java
+++ b/engine/src/main/java/org/destinationsol/game/SolGame.java
@@ -58,6 +58,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 public class SolGame {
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
Went ahead and merged while adding `drawableDebugger` back in, in case it was an oversight. If not just make a sound on chat @Avalancs and we can revisit :-) Thanks!

Issue #267, opened 2018-04-02
Ammo is not saved


oniatus:
When a game is loaded the player will always start with full ammo.
Steps to reproduce:
1. Start a new game with "Loaded Imperial Large" starting ship
2. Fire some rounds with left and right gun
3. (optional): Respawn: The player keeps the ammo
4. Exit to menu then continue the game, the player has full ammo again
Money on the other hand is saved.

Steampunkery:
I can fix that. I'll get right on it.

Steampunkery:
I stopped working on this because gestalt integration will serialize the object, thus saving the ammo. I can make a temporary fix though.

Avalancs:
Tested on newest commit of develop branch, it has been fixed by https://github.com/MovingBlocks/DestinationSol/pull/318 so this issue can be closed.

SimonC4:
Closing this issue as it is confirmed as fixed in version 2.0.0.

PR #318, opened 2018-10-03
Fix ammo saving while reloading

Adrijaned:
Before, if the game was exited while the guns were being reloaded, player would lose the ammo being loaded in the gun. This hs been fixed in this PR. This is not the same issue as #267


GooeyHub:

Issue #313, opened 2018-09-20
Lost equipped item on death


Avalancs:
There was an instance when in the game I died and the money went from 1000 to 15 and I lost the armor that was equipped. Based on past experience and some code reading you are not supposed to be able to drop equipped items, and money should go down by 25% only.

### Log details and game version
latest develop 8e371845af9b17cfbe1bc7ab8dd77f66d8e80577

Adrijaned:
Might be related/be the same thing as #319

Cervator:
Merge the auto-save PR. Good note from @Adrijaned there near the end that I think we should try to address via this issue:

> In one of oniatus' refactors of create player workflow, which accepts the new ship for player to create with items and such, or null if it is o be the same as previous, which is used on startup/respawn. However, when called with null, it loads the previous ship from savefile (y'know, startup), but for respawn, this means you respawn with ship you last used before saving.

So we want to come up with some sane behavior in that case to avoid surprises around death/respawn/reload.

PR #319, opened 2018-10-03
After dying, you won't respawn with your previous save.

Adrijaned:
Also, there goes autosaving, every 60 seconds!

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/SolGame.java
+++ b/engine/src/main/java/org/destinationsol/game/SolGame.java
@@ -17,6 +17,7 @@ package org.destinationsol.game;
 
 import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.math.Vector2;
+import com.badlogic.gdx.utils.Timer;
 import org.destinationsol.CommonDrawer;
 import org.destinationsol.Const;
 import org.destinationsol.GameOptions;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Cervator:
I think I need a bit more detail on this. I know #313 is a thing and Ererbe mentioned a similar case of losing stuff on Discord. But on very latest develop I can't reproduce it. Or at least I come back with the same fancy stuff I started with rather than in a blank newbie ship.

Auto-save is a fine thing and all but I'm not sure how it relates to the dying bit? Have you been able to reproduce or sort out the related code to explain the mysteries? :-)

In other words: this change looks fine when it comes to keeping a better save. But how does it differ from current state and what *should* actually happen? Is it that you respawn with the same setup you picked in new game, so if you start with a loaded imperial you *always* start with that if you die, but we want to make it so you respawn in the exact same state that you died in? And do we want that despite it removing a certain element of risk?

Adrijaned:
How to easily test this: on develop, start a new game ith loaded imperial
big, then exit to menu, then start a new game with imperial small,then drop
all your items, then ram into everything solid until you die,then respawn,
you'll be in the imperialBig again

čt 4. 10. 2018 v 5:04 odesílatel Rasmus Praestholm <notifications@github.com>
napsal:

Adrijaned:
In one of oniatus' refactors of create player workflow, which accepts the new ship for player to create with items and such, or null if it is o be the same as previous, which is used on startup/respawn. However, when called with null, it loads the previous ship from savefile (y'know, startup), but for respawn, this means you respawn with ship you last used before saving.

Cervator:
Ahh. Alright. Lets fix that part then! In addition to auto-save :-)

I'll merge this and submit a separate issue. Thanks!

Cervator:
Merged! I switched to 30 sec auto-save to match Terasology, hope that's OK :-)

rraallvv:
On Android Lollipop the game is not restored and the button "Previous Ship" is always disabled in the menu "New Game", does anyone know if this is a known issue?

PR #320, opened 2018-10-04
Add drawableDebugger to onPausedUpdateSystems

NicholasBatesNZ:
4feca26 forgot to add `drawableDebugger` to the `onPausedUpdateSystems`, this completes #317

PR #321, opened 2018-10-05
Merge strategy plow

Cervator:
This is simply #311 with rushed conflict resolution and an initial round of testing. Also one small review comment fix from @vampcat made to @Adrijaned that really might have been for himself since Adri just redid the PR? So naturally I tried to fix it!

Honestly I had only merged #311 locally as I had the minimal #320 merged in and wanted something more meaty - then saw 18 files with substantial conflicts and tried to merge through the power of lulz. I did *not* expect it to compile, let alone work - but it did, so I guess we might consider this a round 3 and maybe even merge it tomorrow?

I plan to run a round of formatting on top before merge as I'm sure I left some unused stuff in places from the rushed merge and some of the files affected were probably full of style issues even before the PR. Sure got Checkstyle mad at me trying to commit.

Will also actually test a bit more but I'm really surprised at the apparent absence of issues. What I did notice - and possibly only the first one is actually from the base PR here:

1) main menu reversed the button order, but just that one screen (huh) - I think?
2) main buttons on said main menu can be oddly sized, with their button/wrapper getting smaller vs the text as the resolution increases ...
3) some resolutions are off center meaning a bit of the right side gets cut off. Full screen seems fine,

I have to sometimes really mess with the resolution settings to provoke the 2nd and 3rd item so I'm not sure if it is actually this PR. Need to check without the PR, but needed to push the merge somewhere and do a little write-up so I can go clean and try fresh tomorrow.

Also includes #320

### DIFF
--- a/engine/src/main/java/org/destinationsol/CommonDrawer.java
+++ b/engine/src/main/java/org/destinationsol/CommonDrawer.java
@@ -15,9 +15,9 @@
  */
 package org.destinationsol;
 
-import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.GL20;
+import com.badlogic.gdx.graphics.OrthographicCamera;
 import com.badlogic.gdx.graphics.g2d.BitmapFont;
 import com.badlogic.gdx.graphics.g2d.GlyphLayout;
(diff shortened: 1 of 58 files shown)


GooeyHub:

PR #324, opened 2018-10-06
fix main menu button order and offset

Avalancs:
PR #311 introduced changes that have mixed up the order of the buttons on the main menu, which is fixed by this PR.


GooeyHub:

PR #328, opened 2018-10-07
mazesConfig can now be in modules.

Adrijaned:
We might want to do that even for other config files, but so far no demand, so no supply needed. @ererbe needs this so he can work on his module


GooeyHub:

PR #336, opened 2018-10-22
Animated menu screen!

vampcat:
Adds a basic implementation for animated menu. This PR only contains "dumb" asteroids for now, but based on feedback, I plan on making the asteroids collide with each other, add ships, and add animated stars.

Feedback would be appreciated.

(Also, I am hoping to get this done before GCI, so there's that.)

### DIFF
--- a/engine/src/main/java/org/destinationsol/menu/MainMenuScreen.java
+++ b/engine/src/main/java/org/destinationsol/menu/MainMenuScreen.java
@@ -18,14 +18,21 @@ package org.destinationsol.menu;
 
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.Input;
+import com.badlogic.gdx.graphics.Color;
 import com.badlogic.gdx.graphics.Texture;
 import com.badlogic.gdx.graphics.g2d.TextureAtlas;
+import com.badlogic.gdx.math.Vector2;
+import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


vampcat:
Comments by @ererbe :

```
I just tested out the animated main menu, and its very nice.
But sometimes there is popping up a aseroid in the "middle" of my screen, and this is very weird.
Whats about flying and sometimes slowly rotating but not moving space junk?
and increase the minimum speed of the asteriods a bit, I have a big one that is moving a pixel every 10 seconds or something, but its rotating quite fast.
the color variation is amazing btw, make it looks way more spacy.
is this also ingame? when not: we should do something like this. maybe different types of solar systems:
with a specific sun type (size, color, maybe black hole, white or black dwarf, neutron star... etc)
maybe a color type of asteroids or a "min" color and a "max" color.
And planet types that are generated, and the different enemys like normal.
currently there is just hard and normal...
also sometimes there are popping out in the middle of the screen, this looks very weird.
when there is added a colliding, it should be very rare.
maybe a bit random about the density in general.
```

GooeyHub:

bpas247:
Is this a duplicate of #387?

Adrijaned:
@bpas247 Yeah, basically. This was here only for reference purpose during GCI, so I'm going to close this now straightaway.

PR #337, opened 2018-10-25
Stop font size from resizing with window

NicholasBatesNZ:
The size of the font now stays constant even as the window resizes

### DIFF
--- a/engine/src/main/java/org/destinationsol/ui/UiDrawer.java
+++ b/engine/src/main/java/org/destinationsol/ui/UiDrawer.java
@@ -33,12 +33,12 @@ public class UiDrawer implements ResizeSubscriber {
         RIGHT
     }
 
-    public static final float FONT_SIZE = .02f;
-    private static final float FONT_SIZE_MULTIPLIER = 1000f;
+    public static final float FONT_SIZE_PX = 20f;
 
     private Matrix4 straightMtx;
     private final float uiLineWidth;
 
+    public static float fontSize;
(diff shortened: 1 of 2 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/ui/UiDrawer.java
```diff
@@ -87,7 +88,7 @@ public void drawString(String s, float x, float y, float scale, boolean centered
     }
 
     public void drawString(String s, float x, float y, float scale, TextAlignment align, boolean centered, Color tint) {
-        drawer.drawString(s, x, y, scale * FONT_SIZE, align, centered, tint);
+        drawer.drawString(s, x, y, scale * FONT_SIZE * FONT_SIZE_MULTIPLIER / displayDimensions.getHeight(), align, centered, tint);
```
vampcat: We are *definitely* not computing the true value of fontSize on every call to drawString.

Remove FONT_SIZE and FONT_SIZE_MULTIPLIER, and addd two new variables - `FONT_SIZE_IN_PX = 20` and `fontSize = someCalculation;`.

vampcat: We are *definitely* not computing the true value of fontSize on every call to drawString.

Remove FONT_SIZE and FONT_SIZE_MULTIPLIER, and addd two new variables - `FONT_SIZE_IN_PX = 20` and `fontSize = someCalculation;.`

PR #341, opened 2018-10-26
Add module.json to templates directory

Steampunkery:
# Description
This PR adds the file `module.json` to the `templates` directory. This file is used for module creation via `groovyw` and will cause an error if no present.

# Testing
Try to run `./groovyw module create test`. See the error? Now checkout my changes. No error. Look at the module.json in the new module and you'll see it has your module name where appropriate.


GooeyHub:

PR #346, opened 2018-10-27
Got rid of ResolutionScreen

Adrijaned:
Fixing some screens with UIrefactor

PR #349, opened 2018-10-28
UiTextButtons now expand to fit their contents

BenjaminAmos:
# Description
This pull request resizes buttons to fit their contents. This only occurs when assigning the text to a button, so it can be overrided afterwards. It also modifies the UiVerticalListLayout to resize its content width to match the largest width of its individual elements.

# Testing
- Start the game
- Open the Play Game -> New Game menu
- Click the "Starting Ship" button multiple times to cycle through the ships
- The size of the button should change as the ship name gets larger.

# Notes
If the size of any of the buttons changes, the UiVerticalListLayout will change the size of all of the buttons, if the new size is larger than the curent size of the buttons. As such, on the New Ship screen, all the buttons may expand when selecting a new ship.

### DIFF
--- a/engine/src/main/java/org/destinationsol/CommonDrawer.java
+++ b/engine/src/main/java/org/destinationsol/CommonDrawer.java
@@ -115,15 +115,11 @@ public class CommonDrawer implements ResizeSubscriber {
      * @return The visible length
      */
     public int getStringLength(String s, float fontSize) {
-        BitmapFont.BitmapFontData data = font.getData();
-        data.setScale(fontSize / originalFontHeight);
-        int stringLength = 0;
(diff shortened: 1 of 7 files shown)


GooeyHub:

Adrijaned:
@BenjaminAmos I'm afraid your text width getting code doesn't quite work (try putting in other scale than 1)

BenjaminAmos:
I think that I have fixed the issue. The buttons now resize correctly when the text has a different scale. There may be some inaccuracy due to the integer cast although it appears to have a neglible effect on the actual appearance.

vampcat:
Closing, as this PR has been covered by #346.

PR #350, opened 2018-10-29
Auto-Adjusting of UiTextButtons

AndyTechGuy:
# Description
This pull request makes it so `UiTextButton`s now have dynamic width, adjusting to the text inside of them. `UiVerticalListLayout` also will adjust the width of the entire list to fit to the largest width of the elements inside.

# Testing
- Start Game
- Go into UI's with text buttons (Main Menu, Ship Select, Options, etc.)
- Vertical lists should have all buttons as the same size
- The button with the largest string should have a margin of 10px on the left&right

# Notes
- This does not take into account changing text within text buttons, if the text is switched to be larger than the original, the bounding box will be broken.
- `BUTTON_WIDTH` in UiTextButton has been removed entirely, with all references to it replaced with getWidth() of the button or vertical list accordingly.

### DIFF
--- a/engine/src/main/java/org/destinationsol/CommonDrawer.java
+++ b/engine/src/main/java/org/destinationsol/CommonDrawer.java
@@ -82,11 +82,7 @@ public class CommonDrawer implements ResizeSubscriber {
             return;
         }
 
-        font.setColor(col);
-        font.getData().setScale(fontSize / originalFontHeight);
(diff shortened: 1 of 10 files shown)


GooeyHub:

Adrijaned:
Works flawlessly :smiley:

vampcat:
Closing, as this PR has been covered by #346.

PR #353, opened 2018-11-01
Fixed autosave crash when using star ports

BenjaminAmos:
# Description
This fixes a bug where the game crashes if it tries to autosave whilst the hero is in a transcendent state, such as when travelling between star ports.

# Testing
- Un-comment line 156 in [SolGame.java](https://github.com/MovingBlocks/DestinationSol/blob/develop/engine/src/main/java/org/destinationsol/game/SolGame.java#L156)
- Create a new game with 2 systems.
- Visit the nearest star port and enter it.
- Open the console window.
- If the game attempts to autosave, the "game saved" message should be shown. If this message is shown whilst the hero is in transit between two star ports, the game should not crash, Previously, it crashed just before the message was shown. It should not crash when the hero is in a normal state either.

## Notes
This pull request removes the transcendent check on the getMercs method, as it is not actually dependant on the hero ship.  Another approach could be to not autosave when the hero is in a transcendent state, although this could further delay saving for another 30 seconds.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/Hero.java
+++ b/engine/src/main/java/org/destinationsol/game/Hero.java
@@ -172,7 +172,6 @@ public class Hero {
     }
 
     public ItemContainer getMercs() {
-        assertNonTranscendent();
         return mercs;
     }
 


GooeyHub:

GooeyHub:

GooeyHub:

BenjaminAmos:
The game no longer saves when the hero is in a transcendent state. In fact, it should throw an exception if this happens (manually or automatically). I think that this should fix the issue more generally. The hero will be saved whenever he enters or exits a star port.

PR #358, opened 2018-11-10
Added a splash screen on start-up

BenjaminAmos:
# Description
This pull request implements a splash screen that is shown whilst the game is loading its modules on start-up.

# Testing
- Start the game
- There should be a background shown with the text "Loading, please wait..." faintly pulsating in transparency whilst the game loads on start-up, It should disappear once the game has loaded and should then show the main menu screen instead.

# Notes
- The modules are initialised in another thread, although no synchronisation statements are in place. I would not assume this to be a problem though, as the spawned thread only writes to variables that the main thread reads from.
- The loading screen is split into two parts: the background and the load message. As the load message is subtly changed in transparency (as a way of telling the user that the program is working, not frozen and not crashed), it is split into a separate texture. If a static splash-screen is desired, then the two textures can be re-combined and the ```SplashScreen``` class modified accordingly.
- The load message must be a static texture, as the font it requires is loaded separately through the Assets class, which is initialised whilst the screen is being shown.
- I also added some more verbose logging when loading assets, although this can be reverted if needed. The increased verbosity could help to identify asset loading issues more clearly.


GooeyHub:

BenjaminAmos:
This has been replaced by #363.

PR #359, opened 2018-11-11
Upgraded LibGDX to v1.9.8

BenjaminAmos:
# Description
This pull request upgrades LibGDX to the latest released version (1.9.8).

# Testing
This pull request should not introduce any noticable changes.

# Notes
- This partially addresses issue #312.
- This also upgrades RoboVM to v2.3.3 (to be compatible with the new LibGDX version), although I have not tested that it works (does the Android build even compile anymore?).
- Many thanks to Avalancs for the initial implementation.

PR #360, opened 2018-11-11
Upgraded the LibGDX backend from LWJGL2 to LWJGL3

### LINKED ISSUE #312: Upgrade libgdx version, possibly change backend from lwjgl2 to lwjgl3
This is a proposition to upgrade libgdx versions, and possibly the rendering backend in the future.
Destination Sol is currently using libgdx 1.7.0. The full changelog for libgdx can be seen here: https://github.com/libgdx/libgdx/blob/master/CHANGES

Main points for upgrading to libgdx 1.9.9:
Pros:
* reimplementation of freetype which makes fonts look much better on low resolution
* newer versions of bullet and robovm
Cons:
* Android api level 9 requirement. But based on https://developer.android.com/about/dashboards/ this should not be a problem in 2018

The proposal to change the backend is the multi-monitor support which could pontentially fix #40, and the access to vulkan and opencl apis for wizards.

Upgrading only libgdx can be done with minor modifications, which can already be tested on https://github.com/Avalancs/DestinationSol/tree/libgdxUpgrade

BenjaminAmos:
# Description
The version 2 of the LWJGL library is [no longer actively maintained](https://github.com/LWJGL/lwjgl/blob/master/README.md), as stated in the project's README:
> the original LWJGL, which is no longer actively maintained. Unless you have released a product that uses LWJGL 2.x, you should probably be looking at LWJGL 3.

This pull request upgrades the LibGDX backend from LWJGL2 to LWJGL3, which should provide many graphical fixes as well as more general stability in the future.

# Testing
- There should be no visible or functional changes. To test this, just check that there is no broken or changed functionality. 

# Notes
- This fixes #312.
- This depends on #359.
- I had to implement my own framerate-limiter, as there is no built-in one in the LWJGL3 backend, unlike the LWJGL2 one. This should be replaced with a proper solution as soon as possible.

## Known issues:
- Mouse locking has been disabled on Linux, as it does not appear to work anymore (the mouse co-ordinates returned are incorrect when the cursor is locked, so the mouse cursor is never visible when locked).
- This breaks the Android facade due to the SolApplication constructor being modified. I couldn't get the Android facade to build, so I do not have a way to test any fixes for it.
- This breaks the unit tests, due the the new backend not being able to run in environments without a running X server ("X11: The DISPLAY environment variable is missing"). This could be resolved by using the headless LibGDX backend, which I have implemented on my [lwjgl3-upgrade-tests-fix](https://github.com/BenjaminAmos/DestinationSol/tree/lwjgl3-upgrade-tests-fix) branch. The commit is https://github.com/BenjaminAmos/DestinationSol/commit/9626f352ddc13536419ea8a1ad6e4b9a1137b2d1.


GooeyHub:

GooeyHub:

BenjaminAmos:
This breaks the unit tests, due the the new backend not being able to run in environments without a running X server ("X11: The DISPLAY environment variable is missing").

This could be resolved by using the headless LibGDX backend, which I have implemented on my [lwjgl3-upgrade-tests-fix](https://github.com/BenjaminAmos/DestinationSol/tree/lwjgl3-upgrade-tests-fix) branch. The commit is https://github.com/BenjaminAmos/DestinationSol/commit/9626f352ddc13536419ea8a1ad6e4b9a1137b2d1.

I have not included it in this pull request as it is mostly a workaround rather than an actual solution, although I can do so if needed.

Cervator:
I've tested this out some and could not find any obvious issues. As notable by all the new linked issues I created follow-ups for the mentioned problems/hacks :-)

Thanks @BenjaminAmos !

PR #368, opened 2018-11-21
Feature dagger 2 upstream

pollend:
this just merges in the changes from upstream develop.

### DIFF
--- a/config/groovy/common.groovy
+++ b/config/groovy/common.groovy
@@ -108,20 +108,34 @@ class common {
         } else if (itemsRetrieved.contains(itemName)) {
             println "We already retrieved $itemName - skipping"
         } else {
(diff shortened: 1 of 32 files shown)


GooeyHub:

Cervator:
@GooeyHub: add to whitelist

Had not expected that one to be missing :-)

@pollend: I think it might make more sense to _rebase_ this branch on top of develop. With a merge like this we'll get a merge commit and history will get messier than it needs to be. So long as you're sure the rebase worked I'd just push direct without PR. You could still push first to a test branch and validate on GitHub that the compare looks right, just skip actually making the PR.

The more frequently we rebase the branch the better. Easier to deal with smaller batches of changes

GooeyHub:

Issue #373, opened 2018-11-26
problem in playing the game


sohil1234:
when i click tutorial or play game the screen closes giving exit code 0.
and gives this error message:
java.lang.NullPointerException
	at org.destinationsol.assets.audio.OggMusicManager.playMusic(OggMusicManager.java:99)
	at org.destinationsol.SolApplication.play(SolApplication.java:228)
	at org.destinationsol.menu.NewShipScreen.updateCustom(NewShipScreen.java:82)
	at org.destinationsol.ui.SolInputManager.update(SolInputManager.java:220)
	at org.destinationsol.SolApplication.update(SolApplication.java:175)
	at org.destinationsol.SolApplication.safeUpdate(SolApplication.java:153)
	at org.destinationsol.SolApplication.render(SolApplication.java:132)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication.mainLoop(LwjglApplication.java:215)
	at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:120)

sohil1234:
i am getting this message when i am typing command gradlew idea:
Deprecated Gradle features were used in this build, making it incompatible with Gradle 5.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/4.10.2/userguide/command_line_interface.html#sec:command_line_warnings

jellysnake:
Hey, in future please place log ouputs into hastebin.com or pastebin.com and link them instead, like how I edited your comment, because they are _really_ long.

As for the error, the errors about the gradle features being deprecated are benign and hence can be ignored.  
As for the main error, did you delete your copy of the game and then re-clone? Because the errors being thrown seem to indicate that there is some mismatch and you are not on the latest version for all of the code.

sohil1234:
yes. can you tell me how to fix it

arpitkamboj:
You just need to reclone the workspace. Check [this tutorial](https://github.com/MovingBlocks/DestinationSol/wiki/Full-Workspace-Setup) for more info.

sohil1234:
i recloned the workspace as per the tutorial but still facing the same problem

vampcat:
Hey sohil.

[...]

PR #374, opened 2018-11-26
Fixed random playlist bug and no registration bug

Torpedo99:
# Description
This PR fixes the random playlist bug where a random song played from music.
It also fixes other bugs such as no registration of music when game is started using "continue".


GooeyHub:

engine/src/main/java/org/destinationsol/assets/audio/OggMusicManager.java
```diff
@@ -206,5 +203,24 @@ public void registerAllMenuMusic() {
...
+        registerMusic(MENU_MUSIC_SET, "engine:dreadnaught");
+        //Skips to module music
+        playMusicTrack(musicMap.get(MENU_MUSIC_SET).get(0), options);
+    }
+
+    /**
+     * Unregisters all Module Music
+     */
+    public void unregisterModuleMusic() {
+        musicMap.get(GAME_MUSIC_SET).remove(0);
```
Adrijaned:
```suggestion
        musicMap.get(GAME_MUSIC_SET).clear();
```

GooeyHub:

engine/src/main/java/org/destinationsol/game/Hero.java
```diff
@@ -40,11 +41,11 @@
     private boolean isTranscendent;
     private boolean isDead;
 
-    public Hero(SolShip shipHero) {
+    public Hero(SolShip shipHero, SolGame solGame) {
```
Cervator: Watch out - this is a potentially _big_ change. You need to be careful about what code you might affect when you do things like add or subtract arguments on heavily used objects. It does seem like you found the other spots where `Hero` is constructed, but there are suddenly 27 unit tests that fail, indicating something went really wrong. This might relate

See http://jenkins.terasology.org/view/DestSol/job/DestinationSolPRs/697/testReport/

engine/src/main/java/org/destinationsol/assets/audio/OggMusicManager.java
```diff
@@ -145,53 +148,47 @@ public void changeVolume(GameOptions options) {
...
+            return;
         }
 
+        Json musicJson = Assets.getJson(moduleName + ":musicConfig");
 
-        registerMusic(GAME_MUSIC_SET, "engine:cimmerianDawn");
-        registerMusic(GAME_MUSIC_SET, "engine:intoTheDark");
-        registerMusic(GAME_MUSIC_SET, "engine:spaceTheatre");
+        MusicConfig musicConfig = MusicConfig.load(moduleName, musicJson.getJsonValue());
+        Map<String, List<String>> musicMap = musicConfig.getMusicMap();
```
Cervator: This local `musicMap` variable _hides_ the `musicMap` instance variable declared at the top of a class, a bad practice and common cause of bugs. Your IDE should be highlighting this as something to look into. You could simply rename the local variable here but watch out - is there a bug? _Which_ `musicMap` are you needing to interact with in this method?

engine/src/main/java/org/destinationsol/assets/audio/OggMusicManager.java
```diff
@@ -145,53 +148,47 @@ public void changeVolume(GameOptions options) {
...
+            } else {
+                unregisterModuleMusic();
+                currentlyRegisteredModule = null;
             }
+        }
 
-            json.dispose();
+        if (Assets.getAssetHelper().list(Json.class, moduleName + ":musicConfig").isEmpty()) {
+            logger.warn("Music config not found for module " + moduleName + ":musicConfig");
+            return;
```
Cervator: Be careful with `return` statements for error handling in large methods like this. By skipping the remainder of the code in the method you need to be sure none of it is needed to do cleanup or other steps you want to make sure happens regardless of errors.

See https://github.com/MovingBlocks/DestinationSol/issues/373#issuecomment-442121528 - you seem to have previously introduced an issue that's resolved by switching a `return` to a `break` 

While this PR seems to resolve that issue it is an indicator that the way this is written may not be quite right yet. If music config is optional for a module why are we logging a warning when not finding it? That's okay, that music just doesn't need it. 

Take a close look at your logic flow. Review that each `return` makes sense where it is as opposed to reorganizing the code in a method to where you don't need them (better if-then-else tends to help). Consider the different possibilities - is there any way to end up in a weird situation (a module that has music but no music config, or vice versa) that would run through your logic in an unexpected way?

At one point during testing (before this PR, with the prior music issue present) I somehow provoked a situation in which the game crashed, but didn't close gracefully. Usually that's an indicator that the logic flow around error handling went sideways and didn't manage to clean up after itself. I am not sure exactly how I provoked it and this PR would probably change that whole area anyway, but that's one of the sort of things to look out for :-)

GooeyHub:

GooeyHub:

engine/src/main/java/org/destinationsol/assets/audio/OggMusicManager.java
Torpedo99: Yes the return caused errors as the default music was registered only after the module music is loaded. If no module music was loaded then the default music would never be registered. However this was fixed in the new PR as default music is loaded separately. I've also added better log messages now.

GooeyHub:

GooeyHub:

Cervator:
Thanks a lot - this fixes the music crash bug so I'm grabbing it. Appreciate it :)

PR #381, opened 2018-12-09
Added initial support for code modules

BenjaminAmos:
# Description
This pull request fixes the loading of the engine module from the class path and fixes the warnings produced when loading in code modules via gestalt (when it tries to read assets as classes). It also provides a method of creation for code modules. It exposes the following functionality to modules:
- Creating and defining new abilities (```ShipAbility``` and ```ShipAbilityConfig```)
- Listening to the game update loop (```UpdateAwareSystem``` and ```@RegisterUpdateSystem```)
- Adding static command/console handlers for use with the built-in console (```ConsoleInputHandler```)
- Some classes's constructors have been made public
- The ```SolGame``` class now exposes its ```HullConfigManager``` instance.
- The ```MainGameScreen``` class now allows other classes to add ```WarnDrawer``` instances to it.
- The ```MainGameScreen``` class now allows adding overlay screens to it which do not disable the ship controls when active (see notes).

# Testing
### Code Modules
For testing, you can create some classes using the information in the notes below. There are two modules that I have created that do this: my [warp](https://github.com/BenjaminAmos/warp) and [codeModuleTest](https://github.com/BenjaminAmos/codeModuleTest) modules. Testing instructions for those modules can be found below.
### Abilities ([warp](https://github.com/BenjaminAmos/warp))
Start a new game with the "Warp Endeavour" ship. Then, use the ship's ability, which should move the hero to the nearest planet. This is an example of a custom ability (using the newly exposed ```StarPort.Transcendent``` class).
### Commands ([codeModuleTest](https://github.com/BenjaminAmos/codeModuleTest))
Start a new game with any ship. Open the console screen and enter the command "TestCodeModule" (case sensitive). You should receive the response "This is the response to the TestCodeModule command.".
### Application Update loop listeners ([codeModuleTest](https://github.com/BenjaminAmos/codeModuleTest))
Start a new game with any ship. In the top-left corner of the screen you should see a counter showing how many seconds you have spent playing the game in your current session (paused and menus do not count). It should be constantly incremented, except when the game is paused.

# Notes
### Abilities
The ability configuration classes must implement ```AbilityConfig``` and follow a certain naming convention to be detected for their configuration files. The configuration class must be named [AbilityName] + "Config". The ability classes must implement the ```ShipAbility``` interface, although they are only used by the config classes.

### Static Commands
A command is assumed to be any class that inherits from ConsoleInputHandler. It must have a parameterless constructor to be created automatically.

### Build Scripts
This pull request includes several files from Terasology, including some modified forms of the build scripts. As such, not everything in there may be applicable. There may still be a couple of references to "module.txt" and "Terasology" in the comments.

### Application Update listeners
The application update loop listeners are sequential, consisting of an inverse priority for each listener. This is to allow finer control over the execution sequence of listeners, for example when one system must be updated only after another has. All the default systems have been put at priority 0 (the default), with their order of execution kept entact. The lowest priorities will be executed first, with the highest being executed last. To listen to the game update loop, you need to implement the ```UpdateAwareSystem``` interface and annotate the system class with the ```@RegisterUpdateSystem``` annotation.

### Sandboxing
Modules are only allowed to access a limited subset of APIs, which is derived from Terasology's list. Certain system libraries, such as the ```java.date``` API, will not be useable in modules until they are added to the whitelist. All the assets classes in the engine, as well as the ```SaveManager```, have been given file system permissions. Modules are given the most restrictive permissions by default, with exceptions being made where needed (for serialisers, assets, save data etc.).

### ```MainGameScreen``` changes
The ```MainGameScreen``` only permits the ship controls to be functional if the ```MainGameScreen``` is the only screen present. As such, any other screens presenting non-interactable content would break the game when added on-top.
https://github.com/MovingBlocks/DestinationSol/blob/f5b74588fbe234229493abed7c59101ab4d492af/engine/src/main/java/org/destinationsol/game/screens/MainGameScreen.java#L242

The easiest solution to this seems to be making the ```MainGameScreen``` class manage these screens, referred to as "overlays", directly. This would prevent the need to manually add the screens via ```SolInputManager```, which is known to be problematic. This has been accomplished by the addition of the following methods:
```java
void addOverlayScreen(SolUiScreen screen);
void removeOverlayScreen(SolUiScreen screen);
boolean hasOverlay(SolUiScreen screen);
```

## Credits
This is inspired by #259, which provided the initial fixes needed to load the engine module classes correctly.

## Known Issues
Modules cannot be loaded from jars currently, as the asset classes rely on directly accessible filesystem paths.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ android
 gwt
 modules/*
 !modules/core
+!modules/subprojects.gradle
 
 ## GWT
 war/
--- a/engine/build.gradle
+++ b/engine/build.gradle
@@ -2,6 +2,26 @@ apply from: '../config/gradle/common.gradle'
 
 [compileJava, compileTestJava]*.options*.encoding = 'UTF-8'
 
+import org.reflections.Reflections;
+import org.reflections.scanners.SubTypesScanner;
+import org.reflections.scanners.TypeAnnotationsScanner;
+import org.reflections.util.ConfigurationBuilder;
+
+// Dependencies needed for what our Gradle scripts themselves use. It cannot be included via an external Gradle file :-(
+buildscript {
+    repositories {
+        // External libs - jcenter is Bintray and is supposed to be a superset of Maven Central, but do both just in case
+        jcenter()
+        mavenCentral()
+    }
+
+    dependencies {
+        // Needed for caching reflected data during builds
+        classpath 'org.reflections:reflections:0.9.10'
+        classpath 'dom4j:dom4j:1.6.1'
+    }
+}
+
 dependencies {
     compile "com.badlogicgames.gdx:gdx:$gdxVersion"
     compile "com.badlogicgames.gdx:gdx-box2d:$gdxVersion"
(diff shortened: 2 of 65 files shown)


GooeyHub:

GooeyHub:

GooeyHub:

GooeyHub:

Cervator:
I haven't even started to scratch the surface of this PR, but I can confirm it lets code work in modules with https://github.com/BenjaminAmos/warp a great first example! 

I've made https://github.com/DestinationSol/warp for you to push that module to so we can add it to the regular lineup. But yep it'll need this branch before we can use it. That's fine :-)

Well done!

Cervator:
I've made https://github.com/MovingBlocks/DestinationSol/wiki/Using-code-in-modules in the wiki as an initial placeholder for documenting this new support. Probably some of the doc here could be copy pasted in there and/or we could start linking to some details in the area covered by Terasology (although not entity system magic yet)

@BenjaminAmos thank you again for this! @manas96 thank you for PR #259 that helped get this rolling, as well as to others like @vampcat for poking at Gestalt in ES. The future is now brighter here in DS land and we're one huge step closer to finally releasing v2.0.0 :-)

PR #382, opened 2018-12-09
Max and min distances between planets/mazes and stations

RatMoleRat:
# Description
> Adds max and min distances between planets/mazes/systems and stations. This makes it so that stations, including the one that the player spawns at, stay within a certain distance away from those structures.

# Testing
> The default distances that I chose are several hundred units away. However, you can change these, as they are set by constants in the `GalaxyFiller` class. A good way to test would be to change the min and max distances in the code to something smaller (30 for the min and 50 for the max work well), so that you can see how the object is closer to the initial point. Or you can take a look on the map with varying values for the max and min.

# Notes
> It might be interesting to allow the player to select values for the maximum and minimum distances. I'm not sure how this would fit in to the currently fairly simplistic options available to the player during game creation, however.


GooeyHub:

Adrijaned:
@GooeyHub ok to test

GooeyHub:

engine/src/main/java/org/destinationsol/game/GalaxyFiller.java
```diff
@@ -290,4 +323,71 @@ public Vector2 getMainStationPosition() {
...
+                        stationPos.add(0, modifier);
+                    }
+                } else if (gapY * Math.sqrt(2) <= min) {
+                    if (stationPos.y > nearest.y) {
+                        stationPos.add(0, modifier);
+                    } else {
+                        stationPos.sub(0, modifier);
+                    }
+                }
+            }
```
bpas247: Hmm, there seems to be duplicate code here. Could you try refactoring it to reduce the duplicate code?

engine/src/main/java/org/destinationsol/game/GalaxyFiller.java
```diff
@@ -161,6 +180,7 @@ public void fill(SolGame game, HullConfigManager hullConfigManager, ItemManager
 
             for (ShipConfig shipConfig : sysConfig.constAllies) {
                 int count = (int) (shipConfig.density);
+
```
bpas247: unnecessary space

NicholasBatesNZ:
Hey there @RatMoleRat are you planning on cleaning this up for merge? Let me know :)

NicholasBatesNZ:
Sorry for the inconvenience.

PR #383, opened 2018-12-11
Menu

sohil1234:
the code is same as like of vampcat just added the ships and changed there positions


GooeyHub:

Adrijaned:
DON'T USE GITHUB WEB INTERFACE TO UPLOAD FILES! Also, you just basically copied vampcat's code, changing only from displaying asteroids to displaying ships moving with the same move patter. You even broke the asteroids during that.

Adrijaned:
There is a subtle, yet important line between "similar" and "same", and you crossed it by quite far.

PR #397, opened 2018-12-26
removed gamedrawer in favor of a single drawing interface

pollend:
# Description
This removes the gameDrawer in favor of a unified interface. there is no need to have a boolean check since it appears that spritebatch already does this operation. this should be a lot more maintainable across the board. the old implementation just rewraps the commondrawer without adding a lot of new functionality. 

I think gamedrawer is a bit redundant and removing it would make a lot more sense in this case.

code from spritebatch. so the blendfunction only flushes when the src and dst differ from the boolean values already set in the spritebtach. the extra check is just consuming extra operations.
```
@Override
	public void setBlendFunction (int srcFunc, int dstFunc) {
		if (blendSrcFunc == srcFunc && blendDstFunc == dstFunc) return;
		flush();
		blendSrcFunc = srcFunc;
		blendDstFunc = dstFunc;
```
This should be a small change to the engine and removes a lot of odd unnecessary boilerplate code associated with drawing. the gamedrawer at this point just wraps the commondrawer.


GooeyHub:

engine/src/main/java/org/destinationsol/CommonDrawer.java
```diff
@@ -147,11 +153,11 @@ private void setTint(Color tint) {
         spriteBatch.setColor(tint);
     }
 
-    public void draw(TextureRegion tex, Rectangle rect, Color tint) {
-        draw(tex, rect.width, rect.height, (float) 0, (float) 0, rect.x, rect.y, (float) 0, tint);
+    public void draw(TextureRegion tex, Rectangle rect, Color tint,boolean addative) {
```
Cervator: Typo, or is "addative" used on purpose here and in a few other places?

Also minor style issue with whitespace here and elsewhere

engine/src/main/java/org/destinationsol/game/MapDrawer.java
```diff
@@ -129,7 +130,7 @@ public void draw(GameDrawer drawer, SolGame game) {
         float camAngle = cam.getAngle();
         float heroDmgCap = hero.isTranscendent() ? Float.MAX_VALUE : HardnessCalc.getShipDmgCap(hero.getShip());
 
-        drawer.updateMatrix(game);
+        drawer.setMatrix(game.getCam().getMtx());
```
Cervator: What does this do? Seemingly not part of just switching drawer stuff around?

NicholasBatesNZ:
See #455

PR #406, opened 2019-01-17
Internationalization (i18n)

ZPendi:
# Description
This implements an internationalization system similar to Terasology's. Lang files can be defined for any language, and the text will be seamlessly translated depending on the user's locale. Currently some translations for German have been made.

# Testing
To create translations in other languages, simply follow the format within one of the already created menu translations, and create a way of changing the user's local to that particular language (a button in the new language selection screen works well). To access these translations use the translate function from the Translation class, and pass a string based off of this - "${module:filename#key}"

# Outstanding Work
- [ ] Only a handful of translations have actually been made as of yet. This PR is mainly on implementing the system itself as opposed to the creation of actual translations. Most of what has been done is strictly within the main menu.
- [ ] I'm not heavily familiar with any languages outside of English, therefore the German translations are largely Google Translated, so those probably aren't gonna be final.
- [ ] More languages of course need created, that way so the whole world can appreciate the beauty of Destination Sol
- [ ] Currently it is necessary to relaunch the game for the language changes to take effect. If there is a way to change the text on all screens that have already been loaded that would be ideal, however I have not found a method in doing so


GooeyHub:

GooeyHub:

ZPendi:
I'm not sure whether or not I should have added new lines at the end of the Lang files. I'll save y'all the trouble of marking each of them, and instead just let me know whether or not I should change it

engine/src/main/java/org/destinationsol/menu/MainMenuScreen.java
```diff
@@ -57,25 +58,25 @@
         displayDimensions = SolApplication.displayDimensions;
 
         tutorialControl = new SolUiControl(buttonWidth, buttonHeight, UiDrawer.positions.get("bottom"), 0, calculateButtonOffsetFromBottom(3), true, Input.Keys.T);
-        tutorialControl.setDisplayName("Tutorial");
+        tutorialControl.setDisplayName(Translation.translate("${core:titlemenu#tutorial}"));
```
NicholasBatesNZ: Instead of having to `Translation.translate("${GESTALT_ID}")` every time, can't we simply pass `GESTALT_ID` to `setDisplayName(GESTALT_ID)` which will then call the translation itself?
I think that it should be up to the function that actually does the text display to handle the translation thus the calls to it only need to provide where to look for said translation.
NOTE that this might mean changing the name of the `setDisplayName(...)` function to something more appropriate, like `setDisplayTranslationString()` or something

engine/src/main/java/org/destinationsol/game/i18n/TranslationData.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private String findString(String fileName, String key) {
+        Lang lang = Assets.getLang(fileName);
+        String entry = lang.getLangValue().getString(key);
+        Lang localeLang = Assets.getLang(fileName + "_" + Translation.getLocale());
+        return localeLang.getLangValue().getString(entry);
+    }
+
+    private void syntaxCheck(String data) {
+        if (!(data.contains("${") && data.contains(":") && data.contains("#") && data.contains("}"))) {
```
Adrijaned: Use regex please

engine/src/main/java/org/destinationsol/menu/LanguageScreen.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        if (germanControl.isJustOff()) {
+            locale = "de";
+        }
+        if (exitControl.isJustOff()) {
+            if (locale != null) {
+                solApplication.getOptions().setLocale(locale);
+            }
+            solApplication.getOptions().save();
+            Gdx.app.exit();
```
Adrijaned: Why exit the game?

engine/src/main/java/org/destinationsol/menu/MainMenuScreen.java
Adrijaned: This is the way tera does it. Also, said changes would need to happen in uiOverhaul branch.

modules/core/assets/i18n/titlemenu_en.lang
```diff
@@ -0,0 +1,7 @@
+{
+  "tutorial" : "Tutorial",
+  "play" : "Play Game",
+  "options" : "Options",
+  "exit" : "Exit",
+  "credits" : "Credits"
+}
```
Adrijaned: newlines at the end of file please

engine/src/main/java/org/destinationsol/assets/lang/Lang.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2018 MovingBlocks
```
Adrijaned: Year is 2019 now

engine/src/main/java/org/destinationsol/menu/LanguageScreen.java
ZPendi: The text on screens isn't updated upon the change in language. There's probably a way to do so but I didn't see it.

engine/src/main/java/org/destinationsol/menu/MainMenuScreen.java
ZPendi: As Adrijaned had said, according to the wiki https://metaterasology.github.io/docs/developing/translation.html that is the method Terasology has in translation. That definitely sounds like that would be more straightforward though.

GooeyHub:

engine/src/main/java/org/destinationsol/assets/Assets.java
```diff
@@ -163,6 +164,22 @@ public static Json getJson(String path) {
...
+    /**
+     * Loads a Lang (.lang) from the current environment. Throws an exception if the asset is not found.
+     *
+     * @param path A String specifying the desired asset.
+     * @return The loaded Lang.
+     */
+    public static Lang getLang(String path) {
+        Optional<Lang> langOptional = assetHelper.get(parsePath(path), Lang.class);
+
+        if (langOptional.isPresent()) {
```
jellysnake: There is actually a method, `Optional#orElseThrow(Throwable)` that will either return the contained item or, if it's not present, throw the given exception.

That would let you reduce this code down to:
```java
return assetHelper.get(parsePath(path), Lang.class)
            .orElseThrow(new RuntimeException("Lang file " + path + " not found");
```

engine/src/main/java/org/destinationsol/game/i18n/TranslationData.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private String findString(String fileName, String key) {
+        Lang lang = Assets.getLang(fileName);
+        String entry = lang.getLangValue().getString(key);
+        Lang localeLang = Assets.getLang(fileName + "_" + Translation.getLocale());
+        return localeLang.getLangValue().getString(entry);
+    }
+
+    private void syntaxCheck(String data) {
+        Pattern pattern = Pattern.compile("(^\\$\\{)(\\w+)(:)(\\w+)(#)(\\w+)(})");
```
jellysnake: Nitpick:  
This should be in a static field to avoid needlessly compiling the regex each time the method is called.

jellysnake: Also, you don't need those groups and you forgot to specify the end of the string. Try this for a simpler but equivalent regex.

```suggestion
        Pattern pattern = Pattern.compile("^\\${\\w+:\\w+#\\w+}$");
```

engine/src/main/java/org/destinationsol/menu/LanguageScreen.java
jellysnake: I'd _much_ rather it if this instead displayed a popup to the user saying something like:
_"Language will not change until the game is restated.
Do you wish to restart now?"_
And then have _"Yes"_ & _"No"_ options. Ideally in the selected language of course, but even the current one will do if you keep the buttons to just Yes and No

Forcing restart when they push exit seems a bit much. Additionally, if you just open up the language change and then close it without actually editing anything it will exit.

Is there a method to restart instead of just exiting? That seems like it could be preferable as well.

modules/core/assets/i18n/languagemenu_en.lang
```diff
@@ -0,0 +1,5 @@
+{
+  "english" : "English",
+  "german" : "German",
```
jellysnake: It's fairly standard and logical to display the name of the language, in it's own language.
That way German speakers don't need to know the name of their language in English, just in German.
```suggestion
  "german" : "Deutsch",
```

modules/core/assets/i18n/languagemenu_de.lang
```diff
@@ -0,0 +1,5 @@
+{
+  "english" : "Englisch",
```
jellysnake: See below for explanation.

```suggestion
  "english" : "English",
```

modules/core/assets/i18n/languagemenu_en.lang
ZPendi: Yeah I had thought about doing that.

engine/src/main/java/org/destinationsol/assets/Assets.java
ZPendi: For whatever reason it's stating that that method can't be applied to an Optional. I tried to shuffle it around a bit but I was unsuccessful.  The rest of the methods are structured similarly to what I originally had, but if you'd like I could do some more research and fix it.

GooeyHub:

engine/src/main/java/org/destinationsol/menu/LoadingScreen.java
```diff
@@ -49,7 +50,7 @@ public void updateCustom(SolApplication solApplication, SolInputManager.InputPoi
 
     @Override
     public void drawText(UiDrawer uiDrawer, SolApplication solApplication) {
-        uiDrawer.drawString("Loading...", displayDimensions.getRatio() / 2, .5f, FontSize.MENU, true, SolColor.WHITE);
+        uiDrawer.drawString(Translation.translate("${core:loading#loading}") + "...", displayDimensions.getRatio() / 2, .5f, FontSize.MENU, true, SolColor.WHITE);
```
Cervator: I wonder if the `...` here should be made part of the translation string and not hard coded. The way to indicate a pending state may vary between languages. For instance imagine a language like Japanese.

modules/core/assets/i18n/english/languagemenu_en.lang
```diff
@@ -0,0 +1,5 @@
+{
+  "english" : "English",
+  "german" : "Deutsch",
```
Cervator: In Terasology we actually have _two_ fields for languages - their name in English and their name in their origin language. Maybe we should have something like that in DS as well?

Cervator:
@Adrijaned: what do we do about this one and the uiOverhaul? Which goes first to cause the fewest headaches?

engine/src/main/java/org/destinationsol/GameOptions.java
```diff
@@ -250,12 +252,16 @@ public void advanceMusicVolMul() {
...
+        this.locale = locale;
+    }
+
     /**
      * Save the configuration settings to file.
      */
     public void save() {
         IniReader.write(FILE_NAME, "x", x, "y", y, "fullscreen", fullscreen, "controlType", controlType,
-                "sfxVolume", sfxVolume, "musicVolume", musicVolume, "canSellEquippedItems", canSellEquippedItems,
+                "sfxVolume", sfxVolume, "musicVolume", musicVolume, "canSellEquippedItems", canSellEquippedItems, "language", locale,
```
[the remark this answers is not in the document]
ZPendi: I could definitely do that to prevent any potential issues

modules/core/assets/i18n/english/languagemenu_en.lang
ZPendi: Sure, which variation should appear in the language menu though?

GooeyHub:

GooeyHub:

GooeyHub:

modules/core/assets/i18n/english/languagemenu_en.lang
Adrijaned: What about both? `"Deutsch (German)"`

namjisu:
I am a Korean who speaks English well. I will translate it into Korean. Can you also reflect the Korean version? How is it? Isn't it great?

Adrijaned:
Hi @namjisu! For now, we aren't looking for translators yet, as the game isn't ready yet to be translated, and probably won't be for a while still. However, thanks for your interest!

yoonbincho:
If the game is not finished, can I help you with the writing work in Korean?

Adrijaned:
Hi @yoonbincho, as I meantioned earlier - the game is not yet ready to be translated, so even if you wrote the translated texts, we would have no way to actually use them for anything in-game.

yoonbincho:
I'm sorry. There was a misunderstanding in my words. What I wanted to do was ask if I could translate the contributer document or the readme document into Korean.

Adrijaned:
I think I would personally refrain from that. When somebody wants to
contribute to the game, we require all the code in english nevertheless,
and such translated document could also easily go out of date.

engine/src/main/java/org/destinationsol/game/i18n/Translation.java
```diff
@@ -0,0 +1,41 @@
...
+ * Translates information using included lang files
+ */
+public class Translation {
+
+    public static String getLocale() {
+       return GameOptions.getLocale();
+    }
+
+    /**
+     * Syntax should contain a $ at the start and enclosed within braces
```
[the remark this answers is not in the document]
ZPendi: Hey, it's been quite a while but I'm finally ready to start contributing a bit again. Currently I already have the syntax being checked through a regex pattern in TranslationData,

NicholasBatesNZ:
@ZPendi do you plan to continue working on this? By the looks of things we are pretty close.

ZPendi:
Yeah, I will eventually. Sorry, I've just been dealing with a lack of
motivation due to some mental health funstuff, but I will work on it at
some point.

NicholasBatesNZ:
No problem - I'll leave it open for now then :)

GooeyHub:

NicholasBatesNZ:
Not much action - plus I think I just broke it. Oops! Let me know if you want to open it again :+1:

PR #411, opened 2019-04-14
Fixed bug where mercs are removed by star ports

### LINKED ISSUE #354: Mercenaries are removed from the hero when travelling through star ports
I entered a star port after buying some mercenaries, assuming that they would go to the destination once I had arrived there.

### What actually happened
The mercenary UI showed that all the mercenaries were gone when the hero arrived at the destination star port. They still existed in the game world though and would still follow the player. When the game was restarted, the mercenaries were gone and their save file (mercenaries.json) only contained an empty JSON array. This file previously contained the mercenaries I had hired.

### How did it happen
This is what I think caused this issue:
The "[mercs](https://github.com/MovingBlocks/DestinationSol/blob/develop/engine/src/main/java/org/destinationsol/game/Hero.java#L38)" container on the hero [is emptied](https://github.com/MovingBlocks/DestinationSol/blob/develop/engine/src/main/java/org/destinationsol/game/Hero.java#L66-L68) if the hero is assigned a new ship. When the hero traverses through a star port, his ship is converted [into a FarShip](https://github.com/MovingBlocks/DestinationSol/blob/develop/engine/src/main/java/org/destinationsol/game/StarPort.java#L357). When he arrives again, the ship is converted [back into a SolShip](https://github.com/MovingBlocks/DestinationSol/blob/develop/engine/src/main/java/org/destinationsol/game/StarPort.java#L397-L400), which creates a new SolShip object, which is a duplicate of the actual ship. This duplicate is then set as the hero's ship. This recreation is detected as a new ship though, when being set, rather than just a relocated version, so all the mercenaries are lost. This empty data is then saved at the next autosave, leading to a complete loss of mecenaries when you next start the game.

### How to reproduce
* Start a new game.
* Go to a planet and hire some mercenaries.
* Open the mercenaries UI screen to check that they are all present.
* Go through a star port.
* Open the mercenaries UI screen again. The mercenaries previously shown should be gone from the UI screen, although they will still be visible and functional in the game world.
* Quit the game after it has saved.
* Restart the game and continue from the save you just made.
* You should not be able to see any mercenaries in the mercenary UI screen or in the game world.

- Game Version: current "develop" branch (https://github.com/MovingBlocks/DestinationSol/commit/f5b74588fbe234229493abed7c59101ab4d492af)

### Notes
I believe that this bug may have been introduced in commit https://github.com/MovingBlocks/DestinationSol/commit/b15c89b966a922b7a3e7bf3e791dc82ef7599989, although I have not tested it.

IsaacLic:
# Description
Bug #354 was causing the player to lose their mercenaries when they went through a star port. This was because the setSolShip method didn't check for transcendence before creating a new mercenary list. This PR adds a check to ensure that the hero can transcend space and time with impunity :)

# Testing
Hire some mercenaries, go through a star port, then check the mercenaries tab. The mercenaries that were hired should still be there.

fixes #354

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/Hero.java
+++ b/engine/src/main/java/org/destinationsol/game/Hero.java
@@ -66,7 +66,7 @@ public class Hero {
 
     public void setSolShip(SolShip hero, SolGame solGame) {
         isDead = false;
-        if (hero != shipHero) {
+        if (hero != shipHero && !isTranscendent) {
             mercs = new ItemContainer();
         }
         this.shipHero = hero;


GooeyHub:

Adrijaned:
@GooeyHub add to whitelist

GooeyHub:

Adrijaned:
Fixes #354

PR #413, opened 2019-05-11
Remove jre redists

Adrijaned:
Upon discussion on chat, removing jre redists from the source tree as well as git history (for that reason too there are 880 commits in this PR). Major change. Proceed cautiously


GooeyHub:

Cervator:
I was wondering why Jenkins crashed trying to load the DS PR job - found clues in its log that this was the culprit, hehe.

While we should get this done it is probably the rare thing where we'd _not_ want to use a PR at all, but instead do the necessary Git surgery locally then force push it to overwrite the target branch. Which is scary even on a good day.

This will naturally also break every outstanding PR so we should try to get as many of them merged or closed first :-)

NicholasBatesNZ:
Done. Thanks for the idea :)

Issue #414, opened 2019-05-27
DS (or LibGDX) not happy on later Java versions? 10 on Mac seems to cause a failure


Cervator:
### What you were trying to do

Update and run DS on my Macbook while out of town. Would be nice to know what happens on other OSes

### What actually happened

Angry red text (see below). Some of it seems to suggest a JVM flag may be able to help, or to avoid certain other libs?

### How to reproduce

* Be on a later Java (10 in this case - unsure about 9, 11, or 12)
* Possibly be on a Mac
* Simply run the game

### Log details and game version

```
DEBUG [Thread-0] (PermissionSet.java:186) - Added API classes 'org.destinationsol.files.*'
DEBUG [Thread-0] (PermissionSet.java:186) - Added API classes 'org.destinationsol.game.input.*'
DEBUG [Thread-0] (ModuleClassLoader.java:83) - Module path: /Users/cervator/dev/terasology/git_ws/DestinationSol/out/production/engine
 WARN [Thread-0] (Reflections.java:181) - given scan urls are empty. set urls in the configuration
2019-05-27 19:08:49.931 java[27529:613056] pid(27529)/euid(502) is calling TIS/TSM in non-main thread environment, ERROR : This is NOT allowed. Please call TIS/TSM in main thread!!!
2019-05-27 19:08:50.029 java[27529:613056] WARNING: NSWindow drag regions should only be invalidated on the Main Thread! This will throw an exception in the future. Called from (
	0   AppKit                              0x00007fff468667eb -[NSWindow(NSWindow_Theme) _postWindowNeedsToResetDragMarginsUnlessPostingDisabled] + 378
	1   AppKit                              0x00007fff46863bdb -[NSWindow _initContent:styleMask:backing:defer:contentView:] + 1479
	2   AppKit                              0x00007fff4686360e -[NSWindow initWithContentRect:styleMask:backing:defer:] + 45
	3   libglfw.dylib                       0x00000001307d90c9 libglfw.dylib + 57545
	4   libglfw.dylib                       0x00000001307d3a5b libglfw.dylib + 35419
	5   ???                                 0x000000011ad3ab50 0x0 + 4745046864
	6   ???                                 0x000000011ad34a50 0x0 + 4745022032
)
Exception in thread "main" java.lang.ExceptionInInitializerError
	at org.lwjgl.glfw.GLFW.glfwShowWindow(GLFW.java:2149)
[...]

PR #415, opened 2019-05-31
Fix distribution creation and add modules to exported distribution

### LINKED ISSUE #181: Creating distribution of the developer version; crash because resource not found

BenjaminAmos:
# Description
This pull request fixes the `gradlew dist` and `gradlew distZip` commands. It also adds support for including all the modules from the current workspace in the distribution,

The modules will be compiled into jars and copied into the `modules` directory when the distribution is created. Both the code and the assets should be contained within a single jar for each module, named after the module and its version.

Directory-based modules, such as those used for development, should still work when copied into the output modules directory. Zip archives do not work currently.

# Testing
## Testing distribution creation
- Run the command `gradlew dist`.
- A folder, named `app`, should be created within the `desktop/build/distributions/` directory. Enter this directory.
- Execute the relevant launcher (`sol.exe` on Windows, `solOSX.sh` on MacOS and `sol[64].sh` on Linux). The game should start and should be able to be played normally, without any additional crashes.

## Testing distribution zip creation
- Run the command `gradlew distZip`
- A zip archive should be created, called `DestinationSol.zip`, within the `desktop/build/distributions` directory.
- Extract `DestinationSol.zip` into a directory and enter it,
- Execute the relevant launcher (`sol.exe` on Windows, `solOSX.sh` on MacOS and `sol[64].sh` on Linux). The game should start and should be able to be played normally, without any additional crashes.

### Testing code-only modules
- Run the command `groovyw module get codeModuleTest -remote BenjaminAmos` to get my codeModuleTest module.
- Follow the steps contained within the "Testing distribution creation" section.
- Check that in the top-left corner of the screen there is some text labelled "Game Duration: ", followed by the quantity of seconds that the current game session has run for.

### Testing asset-only modules
- Follow the steps contained within the "Testing distribution creation" section.
- Check that the game does not crash when starting a new game. If the core module is not present or loaded then the game will crash.

### Testing modules containing both assets and code
- Run the command `groovyw module get warp` to get the warp module.
- Follow the steps contained within the "Testing distribution creation" section.
- Start a new game with the "Warp Scout" ship.
- Check that the game does not crash.
- Check that there is a button on the right side of the screen labelled "Research".

# Notes
- A dependency on the package `com.github.marschall:zipfilesystem-standalone` has been added to the engine module in order to fix the built-in jre distribution (see https://github.com/MovingBlocks/DestinationSol/pull/300#issuecomment-497719782 for more information).
- This has not been tested on Android.
- I have tested this on Windows and Linux (Ubuntu 19.04) but not on MacOS.
- Modules are not loaded successfully when provided as zip archives (.zip).
- On Linux, the executable permissions are broken, so you will need to add execute rights to the following files:
  - `lwjreLinux/bin/java`
  - `sol.sh`
  - `sol64.sh`
- If merged, this can be considered to supersede #300.
- This fixes #181.

# Credits
- To **Avalancs** for providing the intial changes that partially resolved the issue (see #181)
- To **NicholasBatesNZ** for providing additional improvements to **Avalancs**'s code (see #300)
- To **Adrijaned** for making clear that build.gradle should be ignored in modules (see https://github.com/MovingBlocks/DestinationSol/pull/300#issuecomment-497729991)
- I referenced Terasology's code for this and certain methods used may be similar, if not identical, to code in the following files:
  - [build.gradle](https://github.com/MovingBlocks/Terasology/blob/dbc32f0da97e46c447ae2293a00e5babaf92c85b/build.gradle#L145-L148)
  - [facades/PC/build.gradle](https://github.com/MovingBlocks/Terasology/blob/dbc32f0da97e46c447ae2293a00e5babaf92c85b/facades/PC/build.gradle#L225-L277)
  - [config/groovy/module.groovy](https://github.com/MovingBlocks/Terasology/blob/dbc32f0da97e46c447ae2293a00e5babaf92c85b/config/groovy/module.groovy#L46-L63)


GooeyHub:

engine/src/main/java/org/destinationsol/assets/AssetDataFileHandle.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright 2018 MovingBlocks
```
NicholasBatesNZ: It's 2019 :P

GooeyHub:

BenjaminAmos:
When I copied the zip archive produced on Windows into Linux the permissions were missing. They seem to be correct if compiling on Linux directly.

I have added some commits to reduce the logging level in distribution builds (or more precisely to increase it in IDE and gradle builds). You may need to run the commands `gradlew cleanIdea` and `gradlew idea` in order to use those changes.

I also applied the fix suggested by **Avalancs** in https://github.com/MovingBlocks/DestinationSol/issues/181#issuecomment-349786851, which appears to have fixed the duplicate files in sol.jar.

Cervator:
Added commit for Macs is https://github.com/MovingBlocks/DestinationSol/commit/6ab488d1b0b075c59c622d0a67978944e6ed93e4 - included also making the Mac script executable. Although we might want a `.command` instead - also breaks embedded JRE for Mac which needs a refix along with #414

Cervator:
Oh, and thanks for the great testing instructions by the way. I happened to have both your content module and test module in my workspace still and tested with both :-)

(the mega-weapon on one ship in "warp" is semi-broken by some faction changes though, i think the projectiles insta-collide with each other now)

PR #426, opened 2019-09-24
drop-in framerate synchronization code from LWJGL forums

### LINKED ISSUE #389: Research and implement a better frame limiter if appropriate
Follow-up to #360 which upgraded DS to use LWJGL3 on the backend. This broke the old framerate-limiter and @BenjaminAmos provided a quick effort re-implementation so we could move forward.

The goal of this issue will be to research whether a better implementation could be chosen, or if we should stick with what we have. I imagine plenty of LibGDX-LWJGL3 projects have implemented framerate-limiters, maybe one is hiding in an extension or side project somewhere?

See `SolApplication.java` and search for `//HACK: A crude and primitive frame-limiter...` to find the current replacement

Avalancs:
# Description
Fix #389 by adding in code from LWJGL forums to synchronize framerate.
Source: http://forum.lwjgl.org/index.php?topic=5653.0

# Testing
In the graphic card's driver disable vsync, and start the game. You should see the game run on 100 frames per second. It might take a few seconds for this algorithm to converge to 100.

### DIFF
--- a/engine/src/main/java/org/destinationsol/SolApplication.java
+++ b/engine/src/main/java/org/destinationsol/SolApplication.java
@@ -41,6 +41,7 @@ import org.destinationsol.ui.ResizeSubscriber;
 import org.destinationsol.ui.SolInputManager;
 import org.destinationsol.ui.SolLayouts;
 import org.destinationsol.ui.UiDrawer;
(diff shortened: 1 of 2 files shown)


GooeyHub:

PR #428, opened 2019-09-24
#427 Fix too much money items on death: new limit parameter, and HUGE…

### LINKED ISSUE #427: Game freeze when dying with too much money/items
On the latest develop commit 4a5142c5a35c5efaa18e2dda76dd1c400786a9c9 I was testing the game with 120.000 money, and died on a planet, stuck between a wall and a ground based turret.
The last image I saw was a lot of money coming out of the ship, then the rendering in the game froze. I opened up JVisualVM to check what the game was doing, and it was doing Box2D collision checking.

My theory is that the game spawned too many money/items on death and their collision checking got stuck in an infinite loop. We should:
[ ] Check the number of items emitted on death
[ ] See if collision checking can be skipped/disabled on a large number of collisions

Avalancs:
Fix #427 by limiting the number of money items dropped on death to 20, and added huge money with placeholder icon for dropping 1000 money


GooeyHub:

engine/src/main/java/org/destinationsol/game/item/ItemManager.java
```diff
@@ -231,22 +233,40 @@ public void addAllGuns(ItemContainer ic) {
...
-                amt -= MoneyItem.BIG_AMT;
-            } else if (amt > MoneyItem.MED_AMT) {
-                example = moneyItem(MoneyItem.MED_AMT);
-                amt -= MoneyItem.MED_AMT;
+    /**
+     * Turn money into item drops
+     * @param amount the amount of money to turn into drops
+     */
+    public List<MoneyItem> moneyToItems(float amount) {
+        return moneyToItems(amount, -1);
```
Adrijaned: I'm unsure about the -1. IMO, either set a sane default limit, or mark this method as  as `@Deprecated`, or even straightforward remove it

engine/src/main/java/org/destinationsol/game/item/ItemManager.java
```diff
@@ -231,22 +233,40 @@ public void addAllGuns(ItemContainer ic) {
...
+        ArrayList<MoneyItem> items = new ArrayList<>();
+        int moneyAmount;
+        while (amount > MoneyItem.SMALL_AMOUNTT) {
+            if (amount > MoneyItem.HUGE_AMOUNT) {
+                moneyAmount = MoneyItem.HUGE_AMOUNT;
+            } else if (amount > MoneyItem.BIG_AMOUNT) {
+                moneyAmount = MoneyItem.BIG_AMOUNT;
+            } else if (amount > MoneyItem.MEDIUM_AMOUNT) {
+                moneyAmount = MoneyItem.MEDIUM_AMOUNT;
             } else {
```
Adrijaned: Maybe add `if` there too, so if you die with 15 money for example, you don't end up dropping 20?

engine/src/main/java/org/destinationsol/game/item/ItemManager.java
Avalancs: I didn't want to bake-in a default value, because there might be enemies in the future who might need to drop more, even though this method is just used for the player right now. I also thought about checking out the physics, maybe items should not collide with each other (or maybe if there are too many items on screen), and then the limit could be removed later.

engine/src/main/java/org/destinationsol/game/item/ItemManager.java
Avalancs: Thanks for noticing! Just a sec...

Avalancs: I checked the code, and right now enemies and the player only drop the exact listed amount of money, so 10, 30 and 100 (before I added 1000 in this branch). the remainder is discarded, so money under 10 is not considered. I'll open an issue for it, but I think we should handle it separately from this PR

engine/src/main/java/org/destinationsol/game/item/ItemManager.java
Adrijaned: If you need to drop more items, use the full function instead of this override IMO. The default behaviour should be sane when possible, esp. with the new code in modules stuff.

Adrijaned:
Also, you can get passable "Huge Money" icon by filling in the center of the "Big Money" - could you provide at least that until better option instead of hackish-y reusing the big money?

engine/src/main/java/org/destinationsol/game/item/ItemManager.java
Avalancs: Okay, set default behavior to 60 money items dropped, that should be okay for all needs.

Avalancs:
@Adrijaned  I did put in some extra pixels for the huge money, but I'm bad at art. If you can make one that you think would be suitable feel free to replace it.

GooeyHub:

GooeyHub:

Adrijaned:
I was thinking something like this for the money icon (it's just below this text, although invisible due to white background). What you submitted could be well invisible due to the resolution it is displayed at
iconHugeMoney

Avalancs:
Right, it really is not visible in github's "View file" mode... It looks like this (same white background rule applies):
iconHugeMoney

Adrijaned:
Yeah, I know how what you submitted looks like (hit the download button there to the right) - my comment about it being well invisible was concerning more the thickness of the individual lines composing the icon, especially since it may be even downscaled ingame from the original 32x32px

Avalancs:
Ohh right, it is not very noticeable... the two vertical lines coming out from the center piece are distinct, maybe I could color the center like you said, and add those vertical lines as well? I think that should be visible enough.

Adrijaned:
Cool :+1:

GooeyHub:

PR #431, opened 2019-09-27
Add 'die' console command

Avalancs:
# Description
First commit fixes enter not detected (for me the enter keypress generated '\n' instead of '\r' on windows, weird)
Second commit adds 'die' console command, which kills the player ship by damaging the hull with it's maximum health. Very handy for debugging current issues around dying/respawning.

# Testing
Start the game, bring down the console with the backtick character (`), then type in 'die' and press enter. The player's ship should explode, and items should be thrown out just like if the player died normally during gameplay.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/Console.java
+++ b/engine/src/main/java/org/destinationsol/game/Console.java
@@ -208,7 +208,7 @@ public class Console implements SolUiScreen {
      */
     public void registerCharEntered(char c) {
         if (isActive) {
-            if (c == '\r') {
+            if (c == '\r' || c == '\n') {
                 inputHandler.handle(inputLine.toString(), this);
                 inputLine = new StringBuilder();
                 return;
--- a/engine/src/main/java/org/destinationsol/game/console/ShellInputHandler.java
+++ b/engine/src/main/java/org/destinationsol/game/console/ShellInputHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 MovingBlocks
+ * Copyright 2019 MovingBlocks
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/game/Console.java
```diff
@@ -208,7 +208,7 @@ public static Console getInstance() {
      */
     public void registerCharEntered(char c) {
         if (isActive) {
-            if (c == '\r') {
+            if (c == '\r' || c == '\n') {
```
Adrijaned: How did this even work beforehand? (Don't ask me, I'm the original author)

engine/src/main/java/org/destinationsol/game/console/ShellInputHandler.java
```diff
@@ -40,7 +43,7 @@ public ShellInputHandler() {
     }
 
     /**
-     * Registers a command with this input handler.
+     * Registers a command with this input handler. If the command is already registered throws a {@link SolException}
```
Adrijaned: What about adding a `@throws` javadoc magic word? Not sure how that is called, it's late for me to think any much more now

engine/src/main/java/org/destinationsol/game/console/commands/DieCommandHandler.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2018 MovingBlocks
```
Adrijaned: Not anymore!

engine/src/main/java/org/destinationsol/game/console/ShellInputHandler.java
Avalancs: Thank you, I changed it in the next commit like you asked.

engine/src/main/java/org/destinationsol/game/console/commands/DieCommandHandler.java
Avalancs: Fixed in next commit

GooeyHub:

engine/src/main/java/org/destinationsol/game/Console.java
Avalancs: I don't know... Maybe the merge of lwjgl3 changed the behaviour?

PR #418, opened 2019-07-12
Fixed the Android build

BenjaminAmos:
# Description:
This Pull Request provides the engine changes needed to successfully compile and run Destination Sol on Android.

Due to the changes made, this Pull Request, if merged, may cause merge conflicts with other Pull Requests.

This depends on https://github.com/MovingBlocks/DestSolAndroid/pull/1.

# Testing:
#### Modifying build.gradle to use a custom remote:
- In order to fetch the Android Facade from my fork (that this Pull Request depends on for now), rather than the outdated main repository,
  you should change line 114 of `build.gradle`

from this:
```groovy
    def githubHome = 'MovingBlocks'
```
to this:
```groovy
    def githubHome = 'BenjaminAmos'
```
and you should change line 124 of `build.gradle`

from this:
```groovy
    //println "Fetching $repo from $uri"
```
to this:
```groovy
    branch = 'fix-2.0-build'
```

This will not be needed if https://github.com/MovingBlocks/DestSolAndroid/pull/1 is merged.
#### Building and running the facade
- You can build the android facade by either using the command line tools or using IntelliJ IDEA / Android Studio.
#### Using the command line:
- Download the [Android SDK](https://developer.android.com/studio#downloads) and use the `sdkmanager` tool to install the API 28 build tools.
- In the root project directory, run the command `echo sdk.dir=<SDK_ROOT> > local.properties`, where `<SDK_ROOT>` is the root directory in which
  you installed the Android SDK, to create a new local properties file.
- To fetch the Android facade, in the same root project directory run the command `gradlew fetchAndroid`.
  The command should complete successfully.
- Afterwards, run the command `gradlew android:assembleDebug` to build the APK. This command should also complete successfully.
- Once the command has completed, use the Android emulator or connect a suitable device and ensure that it can be detected by the `adb` tool.
  This can be verified by running the command `adb devices`, which should show the device name in its output.
- Then, run the command `gradlew android:installDebug` to install the APK to the first connected device.
- A new "Destination Sol" app should now be visible on the test device. Run it and play the game.
  It should not crash with any errors or exceptions.
#### Using IntelliJ IDEA (or Android Studio):
- Use the Android plugin (if not installed, then install it) to install the API 28 SDK build tools and platform tools.
- In the root project directory, run the command `gradlew fetchAndroid`. The command should complete successfully.
- Create a new build configuration based on the "Android App" template. You should set the start-up module to the "android" module.
[...]

PR #433, opened 2019-10-12
Draft version of #418 merged with workarounds removed

Cervator:
_Edit: Issues listed below are resolved! Android should work now, just need cleanup and formal rebuilds in Jenkins. I hit one issue trying to fully `gradlew build` Gestalt after some recent changes but that may also be fine now and the snapshots in Artifactory may be enough for now to get this running for any further testing_

Follows up on #418 by replacing the embedded source workarounds with binaries from our Artifactory, compiled once locally by me as test snapshops.

Main critical part was that @BenjaminAmos amazingly made it work with _two_ versions of Gestalt live at the same time - the embedded Gestalt v7 _for Android_ with the binary Gestalt v6 _for the engine_ - somehow this didn't seem to cause any clashes, which is both cool and frightening 😁 

With this (messy draft version!) I've switched over to pure v7 for everything, however that exposed a few parts of the engine that still targets v6 logic, which needs to be tweaked to work with v7.

After fixing that and officially merging the associated https://github.com/MovingBlocks/gestalt/pull/69 (which itself is a draft and needs cleanup) along with another quick Android tweak in https://github.com/MovingBlocks/DestSolAndroid/pull/2

Remaining issues in brief (really, just two classes!):

2019-10-11 22_13_19-DestinationSol  D__Dev_DestSol_git_ws_DestSolAndroidPR_DestinationSol  - engine


GooeyHub:

GooeyHub:

GooeyHub:

NicholasBatesNZ:
# Outstanding work
- [ ] Cleanup @Cervator's Gradle stuff
- [x] Fix engine issues with Gestalt v7. Thanks @BenjaminAmos!
- [ ] Cleanup and merge https://github.com/MovingBlocks/gestalt/pull/69
- [ ] Merge https://github.com/MovingBlocks/DestSolAndroid/pull/2
- [ ] Further Jenkins testing

GooeyHub:

GooeyHub:

GooeyHub:

NicholasBatesNZ:
While doing some gradle cleaning I noticed that having a code module present with the android facade causes problems. The game won't compile or run on either the desktop or android. Eek. Running `gradlew android:assembleDebug` with a code module fails, however it does manage to copy over the assets into the android/assets/modules directory. Thus after deleting the module from the root directory, another run of `gradlew android:assembleDebug` succeeds and includes the assets from the code module. Ships etc. work fine but the code doesn't run. It is evident from this that one must also delete the module from the android folder to properly get rid of it.

BenjaminAmos:
Were you testing using the warp module? The current android implementation dexes (translates java bytecode into android bytecode) module code using "dx" compiler, which does not support de-sugaring Java 8 bytecode, which new syntax features such as lambdas use. Without lambda de-sugaring, the minimum API level supported would be API 26 (rather than the current API 24).

The error encountered should appear like this:
```
com.android.dx.cf.code.SimException: ERROR in org.destinationsol.warp.research.uiScreens.ResearchUiScreen.updateCustom:(Lorg/destinationsol/SolApplication;[Lorg/destinationsol/ui/SolInputManager$InputPointer;Z)V: invalid opcode ba - invokedynamic requires --min-sdk-version >= 26 (currently 13)
```

Directly referenced libraries (such as gestalt and the engine library) are already dexed automatically using the newer "d8" compiler, which does support de-sugaring. I have modified the code to use the d8 compiler, although it does emit some warnings regarding missing interfaces.

The relevant changes needed to use the d8 compiler can be found here:
https://github.com/BenjaminAmos/DestSolAndroid/commit/0f924883e8fbca7231702cfb7f58995c9f34820e#diff-c197962302397baf3a4cc36463dce5eaL210-R269 .

Regarding the module not being removed, this occurs in two instances:
- The first is that the modules directory is not being deleted from the android assets folder. This was unintentional, as at the time I did not realise that `File.delete()` only deletes empty directories or individual files.
- The second was due to the caching of modules locally on the device. The cached modules were not being deleted upon re-installation of the application, so the removed modules were still being loaded from the local cache. Previously, the cache would only be updated when a new application version was detected, although removed files would not be deleted from the cache. The modules are cached on the local filesystem as gestalt does not appear to support loading the modules root directly from an archive (e.g. directly from the apk).

I have updated the module code to delete the modules sub-directory when a new build is performed. The changes I've made should cause the cache to always be re-created on start-up for debug builds and for other build types to only be re-created when a new version is detected (e.g. when the version changes from "2.0.0" to "2.0.1"). The cache is not always re-created by-default at it causes the start-up time to be slightly greater.

The commit will probably not merge cleanly, as it is based on several other un-related commits regarding support for older android versions (lowering the minimum to API 16), which I was previously planning to submit as a pull-request. The changes made should still be applicable to the current codebase, assuming that they are manually ported.

The relevant commit that should fix the module deletion issues can be found here: https://github.com/BenjaminAmos/DestSolAndroid/commit/e31541cca2d4f563f92386e36494b02579e4da44 .

The warp module does not appear to run with the android JSON parser though, due to an additional trailing comma that caused an incorrect array length to be inferred. I have pushed a commit fixing the error to my copy of the repository at https://github.com/BenjaminAmos/warp , which you may need to use instead to test the warp ships.

NicholasBatesNZ:
I tried cloning https://github.com/BenjaminAmos/codeModuleTest into modules and while it did assemble, it wouldn't install:
```
unknown output extension: "/home/nicholas/Documents/DestinationSol/android/assets/modules/codeModuleTest-code-0.1.0.jar"
```
[Log is here](https://pastebin.com/jMFqGM8m)
Now old warp does the same thing, as with your new warp so I guess I went crazy thinking it didn't assemble `¯\_(ツ)_/¯`

I'll see what I can do about bringing those two changes over

BenjaminAmos:
Firstly, there is a bug where the modules are not dexed until the second time gradle is run. The changes made in https://github.com/BenjaminAmos/DestSolAndroid/commit/e31541cca2d4f563f92386e36494b02579e4da44?w=1 should fix that issue, which was causing the module's assets to be copied but not module's code.

The "dx" tool appears to interpret the surrounding quotes as part of the path on Linux. This does not occur on Windows. To fix this, you can change line 238 of `android/build.gradle`
from this:
```groovy
commandLine (["$dex", '--dex', '--output', /*'--keep-classes',*/ "\"$jarOutputDir\""] + moduleClasses)
```
to this:
```groovy
commandLine (["$dex", '--dex', '--output', /*'--keep-classes',*/ "$jarOutputDir"] + moduleClasses)
```

That fixed the issue for me, although the same issue will affect the d8 compiler as well. Removing the quotes does not appear to affect paths containing spaces on both Windows and Linux. The changes can be found in https://github.com/BenjaminAmos/DestSolAndroid/commit/285fa8d54cd8dec3e88f86e3dd59f08c864f36f7, which can be applied directly to the android subdirectory.

If applying the d8-relevant commits, then you will need to change lines 287-290 of `android/build.gradle`
from this:
```groovy
 commandLine (["$dex"] + moduleClasses + ['--classpath', "\"$rootDir/engine/build/classes\"",
                         '--lib', "$path/platforms/android-$compileSdk/android.jar",
                         '--min-api', "$minSdk",
                         '--output', "\"$jarOutputDir\""])
```
to this:
```groovy
commandLine (["$dex"] + moduleClasses + ['--classpath', "$rootDir/engine/build/classes",
                       '--lib', "$path/platforms/android-$compileSdk/android.jar",
                       '--min-api', "$minSdk",
                       '--output', "$jarOutputDir"])
```

I have applied the d8 changes in https://github.com/BenjaminAmos/DestSolAndroid/commit/f74e7c379a7ad9a78bc76dd81f3fbf72398d4b0c.

NicholasBatesNZ:
Thank you @BenjaminAmos that dx fix did the trick! We are getting *so* close to merging this. https://github.com/MovingBlocks/gestalt/pull/71 needs to be merged so that we can tweak and merge https://github.com/MovingBlocks/DestSolAndroid/pull/2 and then I can add these changes to DestSolAndroid. After that I believe it's just a matter of me cherry-picking over those other two commits (UI issue and rebase crash) and merging!

GooeyHub:

GooeyHub:

NicholasBatesNZ:
Closing this in favour of #439 which has everything here plus those rebase fixes

PR #434, opened 2019-10-13
Drop small money

### LINKED ISSUE #430: Cannot drop arbitrary money, only increments of pre-defined values
As the discussion from https://github.com/MovingBlocks/DestinationSol/pull/428 Right now money under 10 (the smalles recognized increment) is discarded when a ship dies and throws out money.

I think the icon for the smallest money should handle 1-10. Since the amount is already stored next to the icon in the MoneyItem class, and the exact amount is saved on the player, this should require minimal code changes.

Malanius:
## Description

Allows dropping any arbitrary amount of money between 0-30 with a small icon instead of discarding amounts below 10.

## Testing

I didn't find a way to directly test this (dropped money on player death != money the player loses), only tested with additional logging added after the counting block.

## Outstanding Work

- [ ] I have several questions regarding this in #430 which might lead to other changes depending of outcome of the answers

# Notes

Fixes #430


GooeyHub:

PR #435, opened 2019-10-13
Add `respawn` command

Malanius:
## Description

When I was testing the drop of a low amount of money, I've found the `die` console command very useful, but soon found out that it's a bit impractical as it has been missing quick way to respawn.
This PR adds `respawn` command to respawn the player's ship after it died.

## Testing

1. Start a new or even existing game
1. Open console
1. Type `respawn`, should show a message that you can't respawn when not dead.
1. Type `die` to kill yourself.
1. Type `respawn` to respawn your ship.

## Notes

Along with the `respawn` command, there are few other commits dealing with few other things in `org.destinationsol.game.console` package:
  - adds unit test for `die` command
  - fixes behaviour of `die` command (was still applying damage no matter of the outcome of alive/transcendent checks)
  - fixes various SonarLint issues in `ShellInputHandler`
  - reformats code inside the package
  - updates the copyright year to 2019 in package files as well as gradle `idea` task

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/console/commands/RespawnCommandHandler.java
+++ b/engine/src/main/java/org/destinationsol/game/console/commands/RespawnCommandHandler.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2019 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/game/console/commands/RespawnCommandHandler.java
```diff
@@ -0,0 +1,49 @@
...
+        this.game = game;
+    }
+
+    private static Logger logger = LoggerFactory.getLogger(RespawnCommandHandler.class);
+
+    @Override
+    public void handle(String input, Console console) {
+        if (hero.isAlive()) {
+            logger.warn("Cannot respawn hero when not dead!");
+            console.println("Cannot respawn hero when not dead!");
```
NicholasBatesNZ: I'd much rather have `console.println()` handle logging so that we don't need to do this every time.

Malanius: Yeah, would be better, I could re-work this in another PR, as this is a bit more unrelated and I have few other ideas that would add to this.

PR #436, opened 2019-10-15
Console enhancements

Malanius:
## Description

This PR contains a few console enhancements:
- adds different levels for messages printed into the console - info, warning, error
- adds different colours to log levels - white, yellow, red
- unifies console print and logging on the same level (previously was required to call `console.println()` and then log)

messages

## Testing

1. Start the game
1. Open console
1. Type command `position` - shows position info, uses original method resulting in info message without logging
1. Type command `respawn` when alive, results in warning message in yellow colour and wan level log
1. Type command `die` twice, when calling `die` when dead, results in an error message in red colour and error level log

## Outstanding Work

- [x] slightly adjust the message levels for current messages (i.e. `die` when dead shows is now an error but should be a warning)
- [x] determine which levels to use in `position` command handler (currently uses `console.println()`)

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/Console.java
+++ b/engine/src/main/java/org/destinationsol/game/Console.java
@@ -92,7 +92,7 @@ public class Console implements SolUiScreen {
     /**
      * Stores all the lines of output printed to console, each line shorter in rendered text width than {@link #MAX_WIDTH_OF_LINE}.
      */
-    private final List<String> linesOfOutput;
+    private final List<ConsoleLine> linesOfOutput;
     /**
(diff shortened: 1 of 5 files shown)


GooeyHub:

Malanius:
@Avalancs couldn't find it myself, would appreciate if someone can point me in the right direction :)

Anyway, if it worked before it shouldn't break, as the `position` command still uses the original method call for printing output, but I'd rather check as well.

NicholasBatesNZ:
Code looks good and testing checks out :+1: 
I think that position should have info level logging and yes, die should warn instead of error.
Also I think you may have forgotten to remove those logger imports ;)
Thanks for your work

GooeyHub:

PR #438, opened 2019-10-20
Position command refactor

Malanius:
## Description

I've been messing around console commands lately and realized that the `position` command required some love to be more readable and was missing a test.
This PR provides both :)

## Testing

Apart from unit test created for this command, you can test it directly in-game:

1. Start a game
1. Open the console
1. Write `position` command + required format - `terse`, `verbose`, `bold`, `internal` or none (defaults to `terse`, anything else will print out a warning with a list of available ones

## Notes

- during the writing of the test for `position` command, I've switched to using static import for Mockito as well as annotated mocks, as it started to be a bit messy, I've done the same for remaining commands test
- the change mentioned above helped to discover that the test for `die` command was completely wrong, testing different command or nothing at all, it's fixed within this PR


GooeyHub:

PR #444, opened 2019-12-03
fix mercs crash

NicholasBatesNZ:
testing: open mercs screen while transcendent


GooeyHub:

PR #445, opened 2019-12-04
audio/menu: Don't reset main menu music when exiting options menu

FearlessTobi:
This resolves the bug that the game music of the title screen would reset when exiting the options menu.

I don't really like having to add extra state tracking here, so if there's a better solution, feel free to tell me.

### DIFF
--- a/engine/src/main/java/org/destinationsol/assets/audio/OggMusicManager.java
+++ b/engine/src/main/java/org/destinationsol/assets/audio/OggMusicManager.java
@@ -37,12 +37,13 @@ import java.util.Set;
  * until another is chosen. By default, music does not play concurrently.
  */
 public class OggMusicManager {
+    public static final String NO_MUSIC = "";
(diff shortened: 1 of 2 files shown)


GooeyHub:

Adrijaned:
@GooeyHub ok to test.

NicholasBatesNZ:
Can we use a constant `NO_MUSIC = ""` to make the code clearer?

GooeyHub:

FearlessTobi:
@Adrijaned While looking at the code before opening this PR, it looked like the code in the MusicManager was actually intended to skip the current track, because there's an extra case [here](https://github.com/MovingBlocks/DestinationSol/blob/b2cd0c2d771d957da68f2d80084b968739bbb264/engine/src/main/java/org/destinationsol/assets/audio/OggMusicManager.java#L95).

GooeyHub:

PR #446, opened 2019-12-05
Address various Checkstyle suggestions

FearlessTobi:
This PR should be fairly self-explanatory.
Please review it commit-by-commit, in order to make it easier to see what changes were done.

If you have any questions why a particular change was done, don't hesitate to ask.


GooeyHub:

engine/src/main/java/org/destinationsol/game/ship/SolShip.java
```diff
@@ -189,7 +189,7 @@ private boolean tryCollectLoot(SolObject obj, SolGame game) {
         boolean canAdd = c.canAdd(i);
         if (canAdd) {
             c.add(i);
-            if (c == myItemContainer && myPilot.getMapHint() == "Merc") {
+            if (c == myItemContainer && myPilot.getMapHint().equals("Merc")) {
```
BenjaminAmos: I noticed that this change could cause a `NullPointerException` crash, as `Pilot.getMapHint` can and does often return `null`.
Examples of where the map hint is set to `null`:
https://github.com/MovingBlocks/DestinationSol/blob/ad2644c4a9dfd142f62ebd28ae0cddc5d50f0ac3/engine/src/main/java/org/destinationsol/game/planet/PlanetObjectsBuilder.java#L353
https://github.com/MovingBlocks/DestinationSol/blob/ad2644c4a9dfd142f62ebd28ae0cddc5d50f0ac3/engine/src/main/java/org/destinationsol/game/maze/MazeBuilder.java#L165
https://github.com/MovingBlocks/DestinationSol/blob/ad2644c4a9dfd142f62ebd28ae0cddc5d50f0ac3/engine/src/main/java/org/destinationsol/game/GalaxyFiller.java#L228
https://github.com/MovingBlocks/DestinationSol/blob/ad2644c4a9dfd142f62ebd28ae0cddc5d50f0ac3/engine/src/main/java/org/destinationsol/game/chunk/ChunkFiller.java#L195

This issue is handled also in the `BeaconHandler` class:
https://github.com/MovingBlocks/DestinationSol/blob/ad2644c4a9dfd142f62ebd28ae0cddc5d50f0ac3/engine/src/main/java/org/destinationsol/game/BeaconHandler.java#L267

Personally, I would suggest changing the above line to be similar to the following (untested code):
```java
String mapHint = myPilot.getMapHint();
if (c == myItemContainer && mapHint != null && mapHint.equals("Merc"))
```

The previous code did not crash because it did not access the value of the strings directly.

I hope this helps, you should be able to trigger the crash by destroying a ship in-game and collecting its "loot".

skaldarnar: Another way to solve problems with NPEs like this is to compare the other way around:

```java
// do
"Merc".equals(myPilot.getMapHint())
// instead of
myPilot.getMapHint().equals("Merc")
```

This, however, only solves the problem of a `null` value from `getMapHint()` - it would still throw a NPE if `myPilot` is `null`. 😉

FearlessTobi: There's also `Objects.equals(x, y)`, which is null-safe by default. Maybe we should use this?

skaldarnar: one has to be aware that [`Objects.equals(...)`](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-) treats to `null` referenes as equal 🤔

Might be a valid alternative as one argument will surely be non-`null`.

FearlessTobi:
Sorry this has gotten so messy now after the rebase. I was fighting some problems with my IDE. I'd really advise squashing all the commits when merging this.

Also, I changed the String comparison for now to use Object.equals.

FearlessTobi:
Rebased. Hope CI compiles now.

GooeyHub:

FearlessTobi:
Closing this for now as it keeps conflicting without being merged.
If someone is willing to get this done, feel free to reopen.

NicholasBatesNZ:
I'm happy to merge this now if you can rebase it before any major PR comes rolling in

GooeyHub:

NicholasBatesNZ:
Scratch that - I've done it :) thanks for your patience
And I'm sorry, but I must have done something wrong since this hasn't closed itself, yet the commits are there in `develop`

FearlessTobi:
Thanks!

PR #447, opened 2019-12-06
Fixed various checkstyle issues

BenjaminAmos:
# Description
This pull request fixes various errors reported through the use of the checkstyle tool. It focuses primarily on errors, although a couple of warnings may have been fixed as well.

# Testing
Test the game generally and it should behave exactly as before. Generating a checkstyle report should produce less errors though. If the game crashes, then it is probably a newly-introduced bug from this pull request, which I will need to fix.

# Notes
I have not tested this extensively, so I cannot be certain if it has any adverse effects on gameplay. So far though, it does not appear to introduce any major bugs that I could find. This may also cause conflicts with other open pull requests, due to the extensive minor changes made.

Some of the issues fixed here may also have been fixed in #446, although I have not thoroughly compared the two pull-requests yet.

A list containing most of the issues fixed by this pull request can be found in the file below:
[CheckStyleIssuesFixed.txt](https://github.com/MovingBlocks/DestinationSol/files/3931569/CheckStyleIssuesFixed.txt)


GooeyHub:

engine/src/main/java/org/destinationsol/IniReader.java
```diff
@@ -67,7 +67,8 @@ public static void write(String fileName, Object... keysVals) {
     }
 
     private void initValueMap(List<String> lines) {
-        for (String line : lines) {
+        for (int lineIndex = 0; lineIndex < lines.size(); lineIndex++) {
```
skaldarnar: Instead of going back from a for-each iterator to a plain for-statement the in-place modification of `line` should be fixed.

Maybe the following works (the important part is to introduce a new local variable `lineWithoutComment` instead of modifying `line` in-place):

```java
for (String line: lines) {
	String lineWithoutComment = line.split("#", 2)[0];
	String[] sides = lineWithoutComment.split("=");
	...
}
```

engine/src/main/java/org/destinationsol/assets/audio/OggSoundManager.java
```diff
@@ -163,11 +163,13 @@ public void play(SolGame game, PlayableSound playableSound, @Nullable Vector2 po
         if (source == null && sound.getLoopTime() > 0) {
             throw new AssertionError("Attempted to loop a sound without a parent object: " + sound.getUrn());
         }
+
+        Vector2 soundPosition = position;
         if (position == null) {
-            position = source.getPosition();
+            soundPosition = source.getPosition();
```
skaldarnar: A good way to ensure that this kind of modification does not happen to parameters it is good practice to declare them as `final`. 😉

engine/src/main/java/org/destinationsol/game/ColorSpan.java
```diff
@@ -54,11 +54,11 @@ public RgbSpan(Color start, Color end) {
 
         @Override
         public void set(float perc, Color col) {
-            perc = MathUtils.clamp(perc, (float) 0, (float) 1);
-            col.r = midVal(myStart.r, myEnd.r, perc);
-            col.g = midVal(myStart.g, myEnd.g, perc);
-            col.b = midVal(myStart.b, myEnd.b, perc);
-            col.a = midVal(myStart.a, myEnd.a, perc);
+            float percentage = MathUtils.clamp(perc, (float) 0, (float) 1);
```
skaldarnar: As stated above, to make it clear to the caller and to guide against bugs `perc` can be annotated as `final float perc` in the parameter list.
This is in addition to introduce a local variable.

engine/src/main/java/org/destinationsol/game/HsbSpan.java
```diff
@@ -30,11 +30,11 @@
 
     @Override
     public void set(float perc, Color col) {
-        perc = MathUtils.clamp(perc, (float) 0, (float) 1);
-        float hue = midVal(0, perc);
-        float sat = midVal(1, perc);
-        float br = midVal(2, perc);
-        float a = midVal(3, perc);
+        float percentage = MathUtils.clamp(perc, (float) 0, (float) 1);
```
skaldarnar: see above 😉

engine/src/main/java/org/destinationsol/game/chunk/ChunkManager.java
```diff
@@ -38,8 +38,8 @@
     private final RemoveController backgroundRemoveController;
     private final ChunkFiller filler;
 
-    private int X;
-    private int Y;
+    private int x;
```
skaldarnar: I don't know this code very well, but why does the chunk manager have coordinates? o.O Maybe these internal members can be renamed to something more specific, which would free the names `x` and `y` as method parameters.

engine/src/main/java/org/destinationsol/game/drawables/DrawableManager.java
```diff
@@ -64,12 +64,12 @@ public static float radiusFromDrawables(List<Drawable> drawables) {
...
     public void removeObject(SolObject o) {
-        List<Drawable> drawables = o.getDrawables();
-        removeAll(drawables);
+        List<Drawable> allDrawables = o.getDrawables();
+        removeAll(allDrawables);
     }
 
-    public void removeAll(List<Drawable> drawables) {
-        for (Drawable drawable : drawables) {
+    public void removeAll(List<Drawable> allDrawables) {
```
skaldarnar: I would call this `toRemove` or even `drawablesToRemove`

engine/src/main/java/org/destinationsol/ui/Position.java
```diff
@@ -29,9 +29,9 @@ public Position(float xNormalized, float yNormalized) {
         set(xNormalized, yNormalized);
     }
 
-    public void set(float xNormalized, float yNormalized) {
```
skaldarnar: There is a setting in Checkstyle to allow for name shadowing in setters and constructors. Maybe this should be enabled for DestSol to avoid cases like this (I think it is completely fine to shadow names in theses cases, but I don't decide this ^^)

engine/src/main/java/org/destinationsol/ui/DisplayDimensions.java
```diff
@@ -29,10 +29,10 @@ public DisplayDimensions(int width, int height) {
         set(width, height);
     }
 
-    public void set(int width, int height) {
-        this.width = width;
-        this.height = height;
-        ratio = width / (float)height;
+    public void set(int newWidth, int newHeight) {
```
skaldarnar: same here

engine/src/main/java/org/destinationsol/game/chunk/ChunkManager.java
BenjaminAmos: The chunk manager appears to use the co-ordinates as the origin point from which it fills chunks, or at least that is what the source code indicates. I have re-named the variables to `originX` and `originY`.

GooeyHub:

engine/src/main/java/org/destinationsol/ui/Position.java
BenjaminAmos: Personally, I would preferred to have done the same, however the checkstyle rules appear to dictate that it is invalid, rather than just warning the user.

The setting is `HiddenField.ignoreSetter` and it is off by default. The configuration is shared between both Destination Sol and Terasology though, so any changes made would affect both projects.
##### (source: https://checkstyle.sourceforge.io/config_coding.html#HiddenField_Properties)
It does seem that this option is already enabled (see [here](https://github.com/MovingBlocks/TeraConfig/blob/44c5632fff7ac493fc4cbc082731dd92b94c01ad/checkstyle/checkstyle.xml#L229)), so should I revert the parameter name changes? Would it be acceptable to add a `SuppressWarnings` annotation on these methods instead?

I think that the reason why they are not caught by the setter exception rule is because they both have more than one parameter and are called `set`, rather than `set<name>`.

BenjaminAmos:
The for-each loop should have been restored and most of the method parameters that were modified inside the method have been marked as final. The variable names in `DrawableManager.java` should now be more descriptive. The only changes that have not been made are regarding the parameter re-naming in setter methods, which may be reverted later-on and replaced with `SuppressWarning` annotations.

engine/src/main/java/org/destinationsol/ui/Position.java
skaldarnar: > I think that the reason why they are not caught by the setter exception rule is because they both have more than one parameter and are called set, rather than set<name>.

That makes sense. In this case, suppressing the warning might be okay here with a remark that it is basically the same as parameter shadowing in setters (which we allow).

BenjaminAmos: I have reverted the variable name changes and added a `SuppressWarning` annotation. The warnings may be visible again as the checkstyle rules do not appear to detect such annotations with the current configuration. This may require a `SuppressWarningsFilter` and a `SupressWarningsHolder` to be added to the configuration file (which is stored in the [TeraConfig](https://github.com/MovingBlocks/TeraConfig) repository).
##### (source: https://checkstyle.sourceforge.io/config_filters.html#SuppressWarningsFilter)

BenjaminAmos:
The setter method parameter name changes have been reverted and replaced with `SuppressWarnings` annotations.

GooeyHub:

NicholasBatesNZ:
Sorry for the inconvenience.

PR #451, opened 2019-12-11
ChangeShip command and tests

Mpcs:
# Description
Adds a new console command "changeship".
All it does is change actual ship. It accepts one argument which is an ID of the target ship.


GooeyHub:

Adrijaned:
@GooeyHub add to whitelist.

engine/src/main/java/org/destinationsol/game/console/commands/ChangeShipCommandHandler.java
```diff
@@ -0,0 +1,75 @@
...
+        console.warn("Usage: \"changeship module:shipName\"");
+    }
+
+    private SolShip cloneAndModifyShip(SolShip originalShip, String newShipID) {
+        HullConfig newHullConfig;
+        try {
+            newHullConfig = game.getHullConfigManager().getConfig(newShipID);
+        }
+        catch (RuntimeException e) {
+            return null;
```
Adrijaned: Could you use `Optional<SolShip>` instead?

GooeyHub:

engine/src/main/java/org/destinationsol/game/console/commands/ChangeShipCommandHandler.java
Mpcs: Yes, changed it in the new commit

GooeyHub:

GooeyHub:

engine/src/main/java/org/destinationsol/game/console/commands/ChangeShipCommandHandler.java
```diff
@@ -0,0 +1,86 @@
...
+    public void handle(String input, Console console) {
+        String[] args = input.split(" ", 2);
+
+        if(args.length != 2) {
+            printHelp(console);
+            return;
+        }
+
+        Optional<SolShip> newShip = cloneAndModifyShip(hero.getShip(), args[1]);
+        if(!newShip.isPresent()) {
```
NicholasBatesNZ:
```suggestion
        if (!newShip.isPresent()) {
```

engine/src/main/java/org/destinationsol/game/console/commands/ChangeShipCommandHandler.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private Optional<SolShip> cloneAndModifyShip(SolShip originalShip, String newShipID) {
+        HullConfig newHullConfig;
+        try {
+            HullConfigManager hcm = game.getHullConfigManager();
+            newHullConfig = hcm.getConfig(newShipID);
+        }
+        catch (RuntimeException e) {
+            System.out.println(e.getMessage());
```
NicholasBatesNZ: Please use the logger instead of System.out

Adrijaned: You could also go around this whole try/catch by using `Assets.getDefaultAssetHelper().listJSONByRegex()` to ensure the existence of the ship, and `Validater.getValidatedJSON()` wit hthe proper schema should ensure you get really a ship. If it excepts even afterward, it's almost definitely a broken ship then, and you should crash in that case.

Adrijaned: Please note - I have no idea how are the functions and classes mentioned in the above comment actually named.  The names I wrote just look righty-ish. You should be able to find them though if you can look around a bit.

GooeyHub:

GooeyHub:

GooeyHub:

PR #452, opened 2019-12-12
camera shake bugfix

Mpcs:
# Description
Fix for the bug causing a camera shake when changing the ship with one with smaller max HP.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/SolCam.java
+++ b/engine/src/main/java/org/destinationsol/game/SolCam.java
@@ -76,9 +76,13 @@ public class SolCam implements UpdateAwareSystem {
             return;
         }
 
+
         Hero hero = game.getHero();
         float life = hero.getLife();
-        Hull hull = hero.getHull();
+        Hull hull = null;
+        if(hero.isNonTranscendent())
+            hull = hero.getHull();
+
         if (hero.isDead() || DIRECT_CAM_CONTROL) {
             applyInput(game);
         } else {
--- a/engine/src/main/java/org/destinationsol/game/SolCam.java
+++ b/engine/src/main/java/org/destinationsol/game/SolCam.java
@@ -48,7 +48,7 @@ public class SolCam implements UpdateAwareSystem {
     private final Vector3 myTmpVec;
 
     private float myPrevHeroLife;
-    private Hull myPrevHeroHull;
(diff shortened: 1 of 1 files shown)


GooeyHub:

Mpcs:
Turns out there was a bug when the ship was transcended. Fixed.

GooeyHub:

engine/src/main/java/org/destinationsol/game/SolCam.java
```diff
@@ -47,6 +48,7 @@
     private final Vector3 myTmpVec;
 
     private float myPrevHeroLife;
+    private Hull myPrevHeroHull;
```
NicholasBatesNZ:
```suggestion
    private Hull previousHeroHull;
```
We are trying to get rid of all these `myX` variable names as well as the abbreviations.

engine/src/main/java/org/destinationsol/game/SolCam.java
```diff
@@ -74,8 +76,13 @@ public void update(SolGame game, float timeStep) {
             return;
         }
 
+
         Hero hero = game.getHero();
         float life = hero.getLife();
+        Hull hull = null;
+        if(hero.isNonTranscendent())
```
NicholasBatesNZ:
```suggestion
        if (hero.isNonTranscendent())
```
Also please use curly braces even if the block is a single line

engine/src/main/java/org/destinationsol/game/SolCam.java
```diff
@@ -94,14 +101,14 @@ public void update(SolGame game, float timeStep) {
             }
         }
 
-        if (life < myPrevHeroLife) {
+        if (life < myPrevHeroLife && hull == myPrevHeroHull) {
```
NicholasBatesNZ: Looking just at the code, it is not immediately obvious why this comparison fixes the bug. Could you please add a comment here

engine/src/main/java/org/destinationsol/game/SolCam.java
Mpcs: Okay, I thought I'll stay consistent with the naming

Mpcs:
Should be okay now.

GooeyHub:

PR #453, opened 2019-12-14
make console case insensitive

NicholasBatesNZ:
Note that this does not make the command arguments case insensitive


GooeyHub:

PR #456, opened 2019-12-21
Module Cleanup

NicholasBatesNZ:
Just trying to spring clean our Gestalt Module integration. Seems to me that we have a bunch of code that is not needed. Relies on Gestalt 7.0.4 though which has not yet been released. Publish it to mavenLocal to test.
Goes with https://github.com/MovingBlocks/DestSolAndroid/pull/7
Any module build.gradle must be deleted. But they shouldn't even have one in the first place...


GooeyHub:

engine/src/main/java/org/destinationsol/assets/textures/package-info.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@API package org.destinationsol.assets.textures;
+
+import org.terasology.gestalt.module.sandbox.API;
```
Avalancs: Do we build Destination Sol with the new JDK now? Does Android support it?

NicholasBatesNZ:
Bigger and better coming soon :tm:

PR #458, opened 2019-12-23
console: Implement help command

FearlessTobi:
This PR implements a help command for DestSol as part of the accompanying [GCI task](https://codein.withgoogle.com/tasks/6055961278545920/?sp-organization=5699684471930880&sp-is_beginner=False).

When providing no argument to the command, a list of all commands will be printed.
If an argument is supplied, a detailed description of that specific command will be printed.

The command descriptions are taken from the wiki and slightly altered, but the final versions are still tbd.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/console/ConsoleInputHandler.java
+++ b/engine/src/main/java/org/destinationsol/game/console/ConsoleInputHandler.java
@@ -28,4 +28,6 @@ public interface ConsoleInputHandler {
      * @param console Console from where the input originates
      */
     void handle(String input, Console console);
+
+    void printHelp(Console console);
 }
--- a/engine/src/main/java/org/destinationsol/game/console/commands/HelpCommandHandler.java
+++ b/engine/src/main/java/org/destinationsol/game/console/commands/HelpCommandHandler.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2019 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.game.console.commands;
+
+import org.destinationsol.common.SolException;
+import org.destinationsol.game.Console;
+import org.destinationsol.game.console.ConsoleInputHandler;
+import org.destinationsol.game.console.ShellInputHandler;
+
+/**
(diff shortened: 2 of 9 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/game/console/commands/HelpCommandHandler.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public void handle(String input, Console console) {
+        String[] args = input.split(" ");
+
+        if (args.length == 1) {
+            console.info("--- Showing all commands ---");
+
+            for (String command : defaultInputHandler.getRegisteredCommands().keySet()) {
+                if (!command.equals("ShellInputHandler")) {
```
FearlessTobi: I don't like hardcoding this at all, but I don't understand why this is even registered as a command.

Adrijaned: I was building that based on the assumption that `/bin/sh` is just a program too.

Adrijaned:
@GooeyHub add to whitelist

GooeyHub:

engine/src/main/java/org/destinationsol/game/console/ConsoleInputHandler.java
```diff
@@ -28,4 +28,6 @@
      * @param console Console from where the input originates
      */
     void handle(String input, Console console);
+
+    void printHelp(Console console);
```
NicholasBatesNZ: Javadoc please

engine/src/main/java/org/destinationsol/game/console/commands/HelpCommandHandler.java
```diff
@@ -0,0 +1,71 @@
...
+
+            return;
+        }
+
+        if (args.length == 2) {
+            try {
+                ConsoleInputHandler handler = defaultInputHandler.getRegisteredCommand(args[1]);
+                handler.printHelp(console);
+            } catch (SolException solException) {
+                console.warn("Unknown command: " + args[1]);
```
NicholasBatesNZ: Shouldn't this be using `console.error()` since it's catching an exception?

engine/src/main/java/org/destinationsol/game/console/commands/InvincibleCommandHandler.java
```diff
@@ -48,4 +48,10 @@ public void handle(String input, Console console) {
             hero.setInvincible(false);
         }
     }
+
+    @Override
+    public void printHelp(Console console) {
+        console.info("Makes your ship invincible or not invincible if it is already.");
```
NicholasBatesNZ: "Toggles ship invincibility" is much clearer imo

engine/src/main/java/org/destinationsol/game/console/commands/HelpCommandHandler.java
Adrijaned: I'm for warn, but this exception catching could be handled more gracefully - e.g. `getRegisteredCommands().keySet().contains()`

NicholasBatesNZ:
This is good work, but unfortunately has been made redundant by #467

PR #459, opened 2019-12-24
Console: setHealthCommandHandler

Mpcs:
# Description
 New command - `setHealth value`. Sets hero's life to the specified value. Has to be between 0 and max Health

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/console/commands/SetHealthCommandHandler.java
+++ b/engine/src/main/java/org/destinationsol/game/console/commands/SetHealthCommandHandler.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2019 MovingBlocks
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/game/console/commands/SetHealthCommandHandler.java
```diff
@@ -0,0 +1,68 @@
...
+        if (newHealthValue < 0f) {
+            console.warn("Cannot set health to less than zero");
+            return;
+        }
+        if (hero.isDead()) {
+            console.warn("Cannot set health when hero is dead");
+            return;
+        }
+
+        hero.getShip().setLife(newHealthValue);
```
Avalancs: Does this work when the hero is in transcendent mode?

Avalancs: Just tested, it throws an exception. please add a check for transcendence

NicholasBatesNZ:
If you have some time, would you mind making a new PR for this command? Your new console work is in `develop` now :)

PR #475, opened 2020-02-12
Updated steam release script and added documentation

SimonC4:
# Description
Updated the steam release script for version 2 of Destination Sol. Specifically the "res" folder has been replaced with "modules". I also added the documentation on how to manually release to Steam that was previously on the [forum](https://forum.terasology.org/threads/steam-build-instructions.1329/). 

# Testing
This can be tested by performing a steam release by following the instructions included in the documentation in this PR. 

# Notes
I have tested this by publishing a version 2.0.0 build to steam on a closed beta branch. 

### Pre Pull Request Checklist:
N/A - no code changes.

PR #478, opened 2020-02-13
Simplify steam upload process

NicholasBatesNZ:
Building on #475 
You now don't need to move things around unless you're updating the JRE with this. Only tested on Linux

### DIFF
--- a/engine/src/main/java/org/destinationsol/GameOptions.java
+++ b/engine/src/main/java/org/destinationsol/GameOptions.java
@@ -17,6 +17,7 @@ package org.destinationsol;
 
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.Input;
+import com.google.common.base.Enums;
 import org.destinationsol.menu.Resolution;
 import org.destinationsol.menu.ResolutionProvider;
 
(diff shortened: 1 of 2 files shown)


GooeyHub:

GooeyHub:

GooeyHub:

steam/scripts/app_build_342980.vdf
```diff
@@ -2,8 +2,8 @@
 {
 	"appid"	"342980"
 	"desc" "Your build description here" // description for this build
-	"buildoutput" "..\output\" // build output folder for .log, .csm & .csd files, relative to location of this file
-	"contentroot" "..\content\DestinationSol\" // root content folder, relative to location of this file
+	"buildoutput" "output\" // build output folder for .log, .csm & .csd files, relative to location of this file
+	"contentroot" "..\..\desktop\build\distributions\DestinationSol\" // root content folder, relative to location of this file
```
SimonC4: If you change the "DestinationSol" directory here to "app" then you shouldn't need to change the output the directory path in the gradle scripts.

NicholasBatesNZ: Yeah, the reason for that though was so you could download a Jenkins build and just extract it right in. The folder is ‘DestinationSol’ in Jenkins.

SimonC4: Right, I misunderstood that. 

I don't think it's a good idea to unzip the file to the same directory path as where you would build a distribution locally as it introduces a risk that you may upload a local build rather than the build from Jenkins.

GooeyHub:

Issue #477, opened 2020-02-13
32-bit Linux Isn't Supported in LWJGL3


SimonC4:
## Overview
Credit goes to @NicholasBatesNZ for finding this.
See below for the detailed bug report.

### Summary
PR #360 upgraded the LibGDX backend from LWJGL2 to LWJGL3. The side effect of this is that 32-bit Linux is [no longer supported](https://github.com/LWJGL/lwjgl3/issues/162). As far as I'm concerned, it is fine that we don't support 32-bit Linux, but there are a few things that we need to do for the version 2 release to ensure 64-bit Linux is supported.

### Tasks
- [x] Update the launch options in Steam to use the `sol64.sh` script instead of `sol.sh`. This should resolve the initial bug report below. (Or change `sol.sh` to use the 64-bit JRE).
- [x] Update the system requirements on the Steam Store page to state that only 64-bit Linux is supported. 
- [x] Do we also update the graphics card requirements to state OpenGL 3.0 compatible? Is LWJGL3 using OpenGL 3.0?
- [x] Remove 32-bit Linux references from the repo. This includes the `sol.sh` script and the associated JRE / script to download 32-bit Linux JRE.

### What you were trying to do
Run the Steam Private Beta of Destination Sol on Linux

### What actually happened
The game wouldn't start with error:
`Failed to locate library: liblwjgl32.so`

### How to reproduce

* Obtain a distribution build of version 2.0 of Destination Sol. Either from the Steam Beta, or [Jenkins](http://jenkins.terasology.org/view/DestSol/job/DestinationSol/), or build it locally. 
* Make sure you are on a Linux distro
* Start the game using the `sol.sh` script

### Log details and game version
![Linux32](https://user-images.githubusercontent.com/7613950/74487301-fcaf0680-4f23-11ea-9c76-c0c5bfd1a79c.png)

### Computer details
Ubuntu 18.04

SimonC4:
LWJGL3 isn't the same as OpenGL 3.0. We are still using OpenGL 2.0 so that task isn't relevant.

SimonC4:
Updated the minimum requirements on Steam to state it "Requires a 64-bit processor and operating system".

PR #479, opened 2020-02-14
Download JRE During Build Distribution Task

SimonC4:
# Description
This adds a new gradle task called `distZipBundleJREs` that is intended to replace the existing `distZip` task. 

The existing `distZip` task builds a distribution of Destination Sol and is used by Jenkins. The problem with this task is that it relies on having the JREs in the repo which we want to remove as part of #413.

The new task `distZipBundleJREs` is functionally similar in that it creates the distribution of Destination Sol. The difference is that the required JREs are downloaded as part of this task so they no longer need to be stored in the repo. 

There are also additional tasks that have been added, which may be useful on their own:
* `distUnbundledJRE` - creates a distribution without any bundled JREs. 
* `distZipUnbundledJRE` same as `distUnbundledJRE` and also creates a zip file.
* `distZipBundleJREs` - same as `distZipBundleJREs` mentioned above, but doesn't create the zip file.

# Testing
From the project root directory run `gradlew distZipBundleJREs` this will create the distribution in the builds/distribution folder. The output of this task should be compared with the output of running the current dist task `gradlew distZip`. The contents of the JRE folders will be different as the bundled JRE uses Liberica 8u212 whereas the JREs in the repo are a very old Oracle version. 

The JRE has changed from an old Oracle version to Liberica 8u212, so ideally the game should be tested to make sure it runs on all supported operating systems. 

# Outstanding Work
There isn't any outstanding work for this PR, but there are some additional tasks that can be done which are dependent on other work
- [ ] Once #413 has removed the JREs from the repo, the gradle scripts can be updated to remove the `dist` and `distZip` tasks.
- [ ] #477 has identified that 32-bit Linux is no longer supported so the downloaded JRE can be removed as part of that task.
- [ ] Once this is merged, Jenkins can be updated to use the `distZipBundleJREs` task.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,6 @@ engine/src/main/resources/settings.ini
 engine/src/main/resources/mercenaries.json
 engine/src/main/resources/world.ini
 engine/src/main/resources/world.json
+
+# Ignore donwloaded JREs
+jre/**
--- a/config/gradle/jre.gradle
+++ b/config/gradle/jre.gradle
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2020 The Terasology Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


GooeyHub:

SimonC4:
> Would it be feasible to have another set of set tasks `distBundleJRE$os` where it only downloads the JRE for the specified OS? Currently I can do this by running `distUnbundledJRE` followed by `downloadJRE$os`

That's probably a good idea. If we decide to create distributions that are OS specific we could create those specific tasks in gradle then.

PR #481, opened 2020-02-16
Remove 32-bit Linux Support.

SimonC4:
Changed the Linux start script to use the 64-bit version of Linux.

# Description
Removes support for 32-bit Linux as it's no longer supported by LWJGL3 as described in #477.
* Changed the Linux start script to use the 64-bit of the JRE.
* Left the existing 64-bit start script for backwards compatibility.
* Removed the 32-bit JRE from the download script.
* Updated the Steam script to no longer upload the 32-bit JRE. Also doesn't upload the 64-bit start script as that has never been used on Steam.

# Testing
Test the distribution build still works and is able to run on Linux from the `sol.sh` start script.


GooeyHub:

PR #469, opened 2020-01-18
Animated menu

Torpedo99:
# Description
Adds an animated menu to the game.
# Features
### Asteroids
Random asteroids (varying in size, texture and color) are spawned randomly from any side of the screen and move at a random velocity, colliding with each other. At any time, a fixed number of asteroids (currently 10), can be found on the screen.
### Ships
Ships can be seen flying from the right top corner of the screen to the bottom left. The ship is chosen at random from a module containing a [menuBackgroundShipConfig.json](https://github.com/Torpedo99/DestinationSol/blob/animatedMenu/modules/core/assets/configs/menuBackgroundShipConfig.json)

PR #483, opened 2020-02-18
Fix android menuBG

NicholasBatesNZ:
I didn't test #469 on android and `JSONArray.iterator` breaks things.

### Testing
Step 1: Open app
Before: crash
After: no crash


GooeyHub:

Issue #105, opened 2016-08-01
Crash while resizing window


gusbitencourt:
http://pastebin.com/9FajR4du

Cervator:
Hi @GBitencourt and thanks for the report!

What system are you running the game on and where did you get the game from? Latest version on Steam? Is there anything unusual on your system that might impact window sizing? Like some sort of window position/size saving utility program, or anything of the kind?

The log statement is very useful and I see where the crash happen, I think we can avoid it at the very least by just canceling the resize or something like that, but am curious what caused that condition to occur in the first place.

GBitencourt:
I'm using windows 10. Downloaded the game on Steam. No, nothing.

Cancel the resizing would fix. You also could put prompt window to choose
the video configuration in the first launch...

I launched the game and started the tutorial, so it started in 800x600
windowed, so instead in change the resolution, I first maximized the
window, I didnt like it because it stretched the graphics, so I put it back
to normal and tried to resize it to keep the ratio and touch the screen
borders, than it crashed. I later launched, changed the screen resolution
and avoided to modify the window size through the window border.

Thanks,
Gustavo

Cervator:
We've been able to reproduce this issue now. Thanks again for reporting it @GBitencourt :-)

GBitencourt:
No problem! Good luck and thanks to keep me in touch!

On Thu, Aug 4, 2016 at 12:28 AM, Rasmus Praestholm <notifications@github.com

crazywolf132:
Anyone still got the paste? I fixed this issue for our version... but i don't remember what we changed.

Cervator:
Just got a new report of what is *probably* the same issue, and I think we can probably still figure out reproducing it. New pastebin also expires but here it is copy pasted in (probably another of the weird crash errors instead of just picking a safe number and logging a warning):

```
USER-GIVEN INFO:
Tried to change the size of the window using the cursor. I was able to grab the corner, but then the mouse was captured again and it brought the window down to 0 size and crashed.
 
ERROR STACK TRACE:
com.badlogic.gdx.utils.GdxRuntimeException: java.lang.AssertionError: normalizing NaN angle
    at com.badlogic.gdx.backends.lwjgl.LwjglApplication$1.run(LwjglApplication.java:127)
Caused by: java.lang.AssertionError: normalizing NaN angle
    at org.destinationsol.common.SolMath.norm(SolMath.java:80)
[...]

PR #499, opened 2020-03-08
Fix alt-tab during full screen crash

Avalancs:
On pc when the game is in full screen, and you press alt-tab, then switch back to the game it would crash. libGDX needs a second when coming back to set the screen size, and immediately when it comes back the window has a resolution of 0*0, and the camera matrix will have NaN entries because of the division with zero.

So I added some checks so if the matrix or mouse coordinates are bad then the code will simply not turn the ship instead of crashing. Also I believe that instead of transforming the mouse screen coordinates to world coordinates, transforming the player ship to screen coordinates is a cheaper (and less error-prone) method for getting the turn angle

# Testing
Start the game, go into full screen, alt+tab, switch back to the game and it should not crash

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/SolCam.java
+++ b/engine/src/main/java/org/destinationsol/game/SolCam.java
@@ -323,4 +323,15 @@ public class SolCam implements UpdateAwareSystem {
         distanceDifference.y = .5f - distanceDifference.y;
         return distanceDifference;
     }
+
+    /** @return true if the camera matrix does not have Nan or infinite values */
+    public boolean isMatrixValid() {
+        for(float i : myCam.combined.val) {
+            if(!Float.isFinite(i)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
 }
--- a/engine/src/main/java/org/destinationsol/game/SolCam.java
+++ b/engine/src/main/java/org/destinationsol/game/SolCam.java
@@ -326,8 +326,8 @@ public class SolCam implements UpdateAwareSystem {
 
     /** @return true if the camera matrix does not have Nan or infinite values */
     public boolean isMatrixValid() {
-        for(float i : myCam.combined.val) {
-            if(!Float.isFinite(i)) {
+        for (float i : myCam.combined.val) {
+            if (!Float.isFinite(i)) {
                 return false;
             }
         }
--- a/engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
+++ b/engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
@@ -30,6 +30,8 @@ import org.destinationsol.game.input.Mover;
 import org.destinationsol.game.input.Shooter;
 import org.destinationsol.ui.SolInputManager;
 import org.destinationsol.ui.SolUiControl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.List;
 
@@ -39,15 +41,16 @@ public class ShipMixedControl implements ShipUiControl {
     public final SolUiControl shoot2Ctrl;
     public final SolUiControl abilityCtrl;
     private final SolUiControl myDownCtrl;
-    private final Vector2 myMouseScreenPos;
+    private final Vector2 mouseScreenPos;
     private final TextureAtlas.AtlasRegion myCursor;
-    private boolean myRight;
-    private boolean myLeft;
+    private boolean turnRight;
+    private boolean turnLeft;
+    private Logger logger = LoggerFactory.getLogger(ShipMixedControl.class);
 
     ShipMixedControl(SolApplication solApplication, List<SolUiControl> controls) {
         GameOptions gameOptions = solApplication.getOptions();
         myCursor = Assets.getAtlasRegion("engine:uiCursorTarget");
-        myMouseScreenPos = new Vector2();
+        mouseScreenPos = new Vector2();
         upCtrl = new SolUiControl(null, false, gameOptions.getKeyUpMouse());
         controls.add(upCtrl);
         myDownCtrl = new SolUiControl(null, false, gameOptions.getKeyDownMouse());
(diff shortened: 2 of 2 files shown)


GooeyHub:

engine/src/main/java/org/destinationsol/game/SolCam.java
```diff
@@ -323,4 +323,15 @@ public Vector2 worldToScreen(SolShip ship) {
         distanceDifference.y = .5f - distanceDifference.y;
         return distanceDifference;
     }
+
+    /** @return true if the camera matrix does not have Nan or infinite values */
+    public boolean isMatrixValid() {
+        for(float i : myCam.combined.val) {
+            if(!Float.isFinite(i)) {
```
NicholasBatesNZ:
```suggestion
        for (float i : myCam.combined.val) {
            if (!Float.isFinite(i)) {
```

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
```diff
@@ -38,15 +39,15 @@
     public final SolUiControl shoot2Ctrl;
     public final SolUiControl abilityCtrl;
     private final SolUiControl myDownCtrl;
-    private final Vector2 myMouseWorldPos;
+    private final Vector2 myMouseScreenPos;
```
NicholasBatesNZ: Please no `my` on any new/changed variables. We're slowly trying to get rid of that prefix

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
```diff
@@ -95,6 +99,33 @@ public void update(SolApplication solApplication, boolean enabled) {
...
+                camera.getViewWidth() <= 0.f || camera.getViewHeight() <= 0.f || !camera.isMatrixValid();
+    }
+
+    /**
+     * Assert that the following vectors do not contain NaN or infinite values
+     * @param mouseCoords mouse coordinates in [0;1] screen space
+     * @param heroCoords hero coordinates projected to [0;1] screen space
+     */
+    private void assertHeroAndMouseCoords(Vector2 mouseCoords, Vector2 heroCoords) {
+        if(Double.isNaN(mouseCoords.x) || Double.isNaN(mouseCoords.y)) {
```
NicholasBatesNZ:
```suggestion
        if (Double.isNaN(mouseCoords.x) || Double.isNaN(mouseCoords.y)) {
```

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
```diff
@@ -95,6 +99,33 @@ public void update(SolApplication solApplication, boolean enabled) {
...
+            System.err.println("Screen size: " + Gdx.graphics.getWidth() + " " + Gdx.graphics.getHeight());
```
NicholasBatesNZ: Use the logger instead of System

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
```diff
@@ -95,6 +99,33 @@ public void update(SolApplication solApplication, boolean enabled) {
...
+            throw new RuntimeException("Mouse coordinates are not valid: " + mouseCoords.x + " " + mouseCoords.y);
+        }
+
+        if(Double.isNaN(heroCoords.x) || Double.isNaN(heroCoords.y)) {
```
NicholasBatesNZ:
```suggestion
        if (Double.isNaN(heroCoords.x) || Double.isNaN(heroCoords.y)) {
```

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
```diff
@@ -95,6 +99,33 @@ public void update(SolApplication solApplication, boolean enabled) {
...
+            System.err.println("Screen size: " + Gdx.graphics.getWidth() + " " + Gdx.graphics.getHeight());
```
NicholasBatesNZ: Logger

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
Avalancs: Fixed in next commit

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
Avalancs: Fixed in next commit

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
Avalancs: That was a debug print which is not longer needed, so I removed it in the next commit

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
Avalancs: Fixed in next commit

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
Avalancs: That was a debug print which is not longer needed, so I removed it in the next commit

Avalancs:
Thanks for pointing out the style issues, I fixed them in the next commit. And I also forgot to mention in the description that this is a Windows libGDX issue, it does not affect Linux

GooeyHub:

Avalancs:
This seems to be the fix for #105

engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
```diff
@@ -30,6 +30,8 @@
 import org.destinationsol.game.input.Shooter;
 import org.destinationsol.ui.SolInputManager;
 import org.destinationsol.ui.SolUiControl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
NicholasBatesNZ: Can remove this since you're not logging anything :)

PR #500, opened 2020-03-09
Persistence

NicholasBatesNZ:
It's not at all pretty - more meant as a temporary solution until Gestalt-Serialization is a thing.
I've added support for some common data types, more will have to be added as needed.

Test with my module: `groovyw module get notepad -remote NicholasBatesNZ`
https://github.com/NicholasBatesNZ/notepad

Breaks on Android but I think that's from a problem on the base `gestaltCleanup` branch


GooeyHub:

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -125,6 +127,11 @@ public void create() {
 
         context.put(ComponentSystemManager.class, new ComponentSystemManager(moduleManager.getEnvironment(), context));
 
+        SerialisationManager serialisationManager = new SerialisationManager(
+                SaveManager.getResourcePath("entity_store.dat"), entitySystemManager.getEntityManager(),
+                moduleManager.getEnvironment().getSubtypesOf(Component.class).iterator().next().getClassLoader());
```
NicholasBatesNZ, line 132: Remove this classloader hack

NicholasBatesNZ: Just made it even more hackier. :P

IsaacLic:
I've tested this, and everything checks out, other than the known issue with module prefabs

Issue #416, opened 2019-06-02
Improve logging setup - Logback/SLF4J instead of log4j?


Cervator:
Currently with Terasology we use Logback/SLF4J for logging and control it and any "bridged" logging frameworks from a single logback config file. I noticed while reviewing #415 that we're still doing log4j things in DS. For consistency's sake (and the added power of the newer frameworks) we should probably look into switching over. 

See Terasology's setup for further details and https://logback.qos.ch/

Extra consideration here: Does Android require any additional considerations?

Avalancs:
Not sure what extras they provide, but there are 2 repos I found:
sl4j-android: https://github.com/nomis/slf4j-android
logback-android: https://github.com/tony19/logback-android

The first repo is just the SL4J abstraction: https://www.slf4j.org/android/
The second repo is sl4j + the modified logback library.

I'll try out the second one, make a PR and ask someone to test it on Android.

PR #501, opened 2020-03-12
Change desktop logger to Logback

Avalancs:
# Description
Work in progress, do not merge!
This is a prototype for #416. I got the Engine tests and desktop to work with SLF4J and Logback, but I do not know how to download the android part.

# Testing
1. Run JUnit tests from the Engine directory. They do not log anything, but you should not receive an exception from SLF4J about missing logger
2. Start the game with the "run" task, and see if messages are printed to console

# Outstanding Work
- [ ] Download android, copy logback.xml to src/main/resource/assets/logback.xml
- [ ] Change build.gradle to use 
```
runtime group: 'com.github.tony19', name: 'logback-android', version: '2.0.0'
runtime group: 'com.github.tony19', name: 'logback-android-classic', version: '1.1.1-6'
```
Instead of ch.qos.logback
- [] Test if logging works when the game is started (it might output an error, or just simply doesn't paste any text. See http://logback.qos.ch/manual/configuration.html#automaticStatusPrinting for help with debugging the logging configuration

### DIFF
--- a/config/gradle/ide.gradle
+++ b/config/gradle/ide.gradle
@@ -106,7 +106,7 @@ ext {
                 </pattern>
               </extension>
               <option name="MAIN_CLASS_NAME" value="org.destinationsol.desktop.SolDesktop"/>
-              <option name="VM_PARAMETERS" value="-splash:engine/src/main/resources/assets/textures/mainMenu/mainMenuLogo.png %s -Xms256m -Xmx1024m -Dlog4j.configuration=log4j-debug.properties"/>
+              <option name="VM_PARAMETERS" value="-splash:engine/src/main/resources/assets/textures/mainMenu/mainMenuLogo.png %s -Xms256m -Xmx1024m/>
               <option name="PROGRAM_PARAMETERS" value="-noCrashReport %s"/>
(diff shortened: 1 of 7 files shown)


GooeyHub:

NicholasBatesNZ:
Hi @Avalancs — thanks for this, and apologies it sat unreviewed for so long; that's on us, not you. Six years on, the underlying goal is still valid: develop is still on `slf4j-log4j12:1.7.25` with the `log4j*.properties` files, and #416 is still open, so nobody has beaten you to it.

Unfortunately the patch itself can't be carried forward — the build has since moved to Gradle 9.6.1 / Java 17 with `build-logic` convention plugins, so the `compile`/`runtime`/`testRuntime` configurations this uses no longer exist and `config/gradle/ide.gradle` has been deleted outright. The Android facade is also still very much alive (DestSolAndroid was updated this week), so the blocker you flagged is unchanged rather than moot.

I'm going to close this as stale to keep the queue honest. That's not a comment on the work — the Android appender/dependency checklist in your description above is still the clearest plan anyone has written for that half of #416, and it's worth anyone picking this up starting from it. If you ever fancy coming back to it, a fresh branch off develop would be very welcome.

PR #502, opened 2020-03-24
Basic gestalt implementation

IsaacLic:
# Description
This is a basic implementation of ECS which uses an event to send updates about the location of an entity. As of now, it crashes after the first time the event is sent.


GooeyHub:

IsaacLic:
I made this to the wrong branch, so I closed it

PR #509, opened 2020-04-07
Do clean builds just in case

NicholasBatesNZ:
Partly opening this to test the discord webhook and partly because it's also wanted :+1:


NicholasBatesNZ:
When a PR is opened from an origin branch Jenkins needs to be manually told to scan the repo since it technically has already done the build on the commit. Ought to be looked into

Cervator:
> Or did the Jenkins webhook break? Seems to be giving a connection error

That's likely when the GitHub -> Jenkins connection falls prey to that DNS failure, like it does to us at times. I wouldn't worry about it :-)

PR #511, opened 2020-05-04
Fixes #510

### LINKED ISSUE #510: beta v2.0.0 - Ship not pointing at a crosshair when on a planet
Greetings.
So, there's a weird bug I found while playing on a new v2.0.0 beta of Destination Sol:
When I land at a planet, my ship stops pointing where crosshair at and starts pointing some other way, and where it's pointing changes as I fly around the planet, for example: when I'm at one side of the planet - my ship points towards the left of a crosshair, and when I move to the other side, it points to the right. As I move around the planet, it shifts all around, including one spot where it points correctly at the crosshair like it should. I only have this bug while I'm on the planet; it fixes as soon as I'm in space. It's kinda annoying, but still playable.

I use Linux Mint. 
Here's also few screenshots on different sides of the planet, where ship SHOULD be pointing at the crosshair, but it isn't.

And normal position in space:

NicholasBatesNZ:
From #499:
```
Also I believe that instead of transforming the mouse screen coordinates to world coordinates, transforming the player ship to screen coordinates is a cheaper (and less error-prone) method for getting the turn angle
```
Evidently not... not sure why but reverting that changes does the trick :thinking: 

Some mild refactoring

Fixes #510

### DIFF
--- a/engine/src/main/java/org/destinationsol/common/SolMath.java
+++ b/engine/src/main/java/org/destinationsol/common/SolMath.java
@@ -309,6 +309,18 @@ public class SolMath {
     }
 
     /**
+     * Assert that no vector contains NaN or infinite values
+     * @param vectors to check
+     */
+    public static void assetReal(Vector2 ... vectors) {
+        for (Vector2 vector : vectors) {
+            if (Double.isNaN(vector.x) || Double.isNaN(vector.y)) {
+                throw new AssertionError("Vector is not valid: " + vector.x + " " + vector.y);
+            }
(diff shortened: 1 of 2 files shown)


Cervator:
Can also confirm, including that I can now alt-tab away from DestSol in fullscreen on Windows, that crashed the game before. It _still_ freaks out Microsoft Explorer to the point where I have to forcibly restart it for some reason, but that might be more about my setup than the game, so not overly worried about that. Merging!

PR #512, opened 2020-05-12
Game creation rework

NicholasBatesNZ:
Refactored some stuff to make a potential multiplayer effort easier, most notably `WorldConfig` usages.
Also puts every mode through the loading screen, not just continue game. Improves the android ux as it previously looked frozen when loading a new game.

# Testing
Play the game. Tutorial, new game and continue game should work as normal.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/SaveManager.java
+++ b/engine/src/main/java/org/destinationsol/game/SaveManager.java
@@ -46,10 +46,11 @@ import java.io.PrintWriter;
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 
 public class SaveManager {
 
-    private static Logger logger = LoggerFactory.getLogger(SaveManager.class);
+    private static final Logger logger = LoggerFactory.getLogger(SaveManager.class);
 
     protected SaveManager() { }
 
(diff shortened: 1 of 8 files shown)


engine/src/main/java/org/destinationsol/game/SaveManager.java
```diff
@@ -46,10 +46,11 @@
 import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 
 public class SaveManager {
 
-    private static Logger logger = LoggerFactory.getLogger(SaveManager.class);
+    private static final Logger logger = LoggerFactory.getLogger(SaveManager.class);
```
Adrijaned, line 53: AllCaps? Or are we not running that convention? Been a while...

engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -107,6 +107,9 @@
 
     public SolGame(String shipName, boolean isTutorial, boolean isNewGame, CommonDrawer commonDrawer, Context context,
                    WorldConfig worldConfig) {
+        // TODO: make this non-static
+        FactionInfo staticIgnore = new FactionInfo();
```
Adrijaned, line 111: Couldn't you, like, at least make a static `init()` method to call instead? This is ugly.

engine/src/main/java/org/destinationsol/game/SaveManager.java
NicholasBatesNZ: Good question... I think we are in some places but not others...

NicholasBatesNZ:
#514

PR #496, opened 2020-02-24
Gestalt cleanup

NicholasBatesNZ:
Context injection via `@In`
ES foundations
Commands now work in modules
Add `build.gradle` to module .gitignore
Less android specific code: https://github.com/MovingBlocks/DestSolAndroid/pull/12
No es-perf for now as android seems to break

`groovyw module get notepad -remote NicholasBatesNZ`
https://github.com/NicholasBatesNZ/notepad

- [ ] Refactor engine to `org.destinationsol.engine` so module code isn't put in the sandbox

PR #515, opened 2020-06-03
Health/damage system with unit tests

IsaacLic:
# Description
This PR adds the first ECS system, which manages health and damage. My unit test currently doesn't work, so something is problematic.
What should happen is that a player should be able to create an entity with a Health component which is lowered when the entity receives a Damage event. In my unit test, the entity doesn't receive the damage. I think the issue is that the system that handles the event isn't getting registered with the event system, even though it is annotated with `@RegisterEventReceivers`.

### DIFF
--- a/engine/src/main/java/org/destinationsol/components/HealthComponent.java
+++ b/engine/src/main/java/org/destinationsol/components/HealthComponent.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2020 The Terasology Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.components;
+
+import org.terasology.gestalt.entitysystem.component.Component;
+
+public final class HealthComponent implements Component<HealthComponent> {
+
+    private int maxHealth = 30;
+    private int currentHealth = 30;
+
+    public int getCurrentHealth() {
+        return currentHealth;
+    }
+
+    public void setCurrentHealth(int health) {
+        this.currentHealth = health;
+    }
+
+    public int getMaxHealth() {
+        return maxHealth;
+    }
+
+    public void setMaxHealth(int maxHealth) {
+        this.maxHealth = maxHealth;
+    }
+
+    @Override
+    public void copy(HealthComponent other) {
+        this.maxHealth = other.maxHealth;
+        this.currentHealth = other.currentHealth;
+    }
+}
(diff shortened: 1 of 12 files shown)


engine/src/main/java/org/destinationsol/components/HealthComponent.java
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.components;
+
+import org.terasology.gestalt.entitysystem.component.Component;
+
+public final class HealthComponent implements Component<HealthComponent> {
```
NicholasBatesNZ, line 20: Gestalt 7 doesn't require component class names to end in `Component` so I would just call this `Health`

IsaacLic: I'm working on the docs now. I can't believe I forgot XD
I'd rather use `isPresent()`, but I'm ok with either. Is either one preferable for any reason?
In terms of naming, should Events also not be called `ExampleEvent`?

NicholasBatesNZ: `isPresent()` is fine by me. `ifPresent()` is nice in that you don't need to `.get()` the component but `isPresent()` is slightly more readable.
You raise a very good point about the naming. I was thinking that I don't really want have to put `HealthComponent` into the prefabs. Not really a huge deal so I'll let you decide :)

IsaacLic: I think that it should be `Health` and `DamageEvent`. With components, the essence is really the thing itself, e.g. health, and `component` is a comment about the software architecture. With events, the main thing is that an event is happening. The "`Damage`" is just data about the event.

engine/src/main/java/org/destinationsol/systems/DamageSystem.java
```diff
@@ -0,0 +1,57 @@
...
+import org.terasology.gestalt.entitysystem.event.EventResult;
+import org.terasology.gestalt.entitysystem.event.ReceiveEvent;
+
+/**
+ * When a damage event happens to an entity with a health component, this system reads the damage from that event and
+ * lowers its health by that amount. If it would lower the health to less than zero, it's reduced to zero instead. If
+ * the damage is a negative amount, nothing happens.
+ */
+@RegisterEventReceivers
+public class DamageSystem extends ComponentSystem {
```
NicholasBatesNZ, line 32: Why are you extending `ComponentSystem`?

IsaacLic: @Cervator suggested it, because he said that annotations shouldn't fundamentally change anything

IsaacLic: (And therefore it should extend ComponentSystem too)

NicholasBatesNZ: The only reasons you would want to extend `ComponentSystem` are if you want access to the `context` or you want to tap into the methods it contains. What we were talking about the other day was that the annotation to register the event receiver can be *replaced entirely* with an interface to implement ie. `implements EventReceiver` see [here](https://github.com/NicholasBatesNZ/notepad/commit/4f4445da10f6d2fe29d3daf79dfcd9d5e531ba83).
You've got to undo the changes in [this commit](https://github.com/MovingBlocks/DestinationSol/pull/496/commits/006da6e2778bb4b0bded428b5f953387f42666ff) to do that.
Ping me on discord if you'd like more clarification :+1:

IsaacLic: Thanks 👍
I made the change

PR #516, opened 2020-06-05
force/contact handling

IsaacLic:
This handles the contact between two entities and the application of force to any entity. When two objects collide, a `ContactEvent` is created which, depending on how it is handled, can then create a `ForceEvent`.

### DIFF
--- a/engine/src/main/java/org/destinationsol/events/ContactEvent.java
+++ b/engine/src/main/java/org/destinationsol/events/ContactEvent.java
@@ -16,6 +16,7 @@
 package org.destinationsol.events;
 
 import com.badlogic.gdx.math.Vector2;
+import com.badlogic.gdx.physics.box2d.Contact;
 import org.destinationsol.game.item.Loot;
 import org.destinationsol.game.ship.SolShip;
 import org.terasology.gestalt.entitysystem.entity.EntityRef;
@@ -23,22 +24,19 @@ import org.terasology.gestalt.entitysystem.event.Event;
 
 /**
  * Event that represents the contact between two entities. Both entities involved in the contact will be sent separate
- * contact events. If an entity should be moved by the contact, its contact handling system should create an
- * {@link ImpulseEvent}. In some cases, the contact should handled without generating an impulse event, such as a
- * {@link SolShip} coming in contact with {@link Loot}.
+ * contact events. If the contact should be modified in any way, the {@link Contact} should be changed. For example, if
+ * a {@link SolShip} comes in contact with {@link Loot}, the contact should handled without generating an impulse.
  * <p>
  * Long-term forces, such as gravity, should be handled by a {@link ForceEvent}.
  */
 public class ContactEvent implements Event {
 
     private EntityRef otherEntity;
-    private Vector2 contactPosition;
-    private float absoluteImpulse;
+    private Contact contact;
 
-    public ContactEvent(EntityRef otherEntity, Vector2 contactPosition, float absoluteImpulse) {
+    public ContactEvent(EntityRef otherEntity, Contact contact) {
         this.otherEntity = otherEntity;
-        this.contactPosition = contactPosition;
-        this.absoluteImpulse = absoluteImpulse;
+        this.contact = contact;
     }
 
     /**
@@ -49,16 +47,10 @@ public class ContactEvent implements Event {
     }
 
     /**
-     * The position where the contact happened.
+     * Returns the {@link Contact} from the physics engine, which contains the information about the contact. This
+     * should be modified if any aspect of the contact should be changed before it is processed by the physics engine.
      */
-    public Vector2 getContactPosition() {
-        return contactPosition;
-    }
-
-    /**
-     * The impulse applied to the entity.
-     */
-    public float getAbsoluteImpulse() {
-        return absoluteImpulse;
+    public Contact getContact() {
+        return contact;
     }
 }
--- a/engine/src/main/java/org/destinationsol/events/ContactEvent.java
+++ b/engine/src/main/java/org/destinationsol/events/ContactEvent.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2020 The Terasology Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.events;
+
+import com.badlogic.gdx.math.Vector2;
+import org.terasology.gestalt.entitysystem.entity.EntityRef;
+import org.terasology.gestalt.entitysystem.event.Event;
+
+/**
+ * Event that represents the contact between two entities. This event can be used to create a force event, if force
+ * should be applied because of the contact.
+ */
+public class ContactEvent implements Event {
+
+    /**
+     * The entity causing the contact event.
+     */
+    private EntityRef triggeringEntity;
+
+    /**
+     * The position that the contact happened at.
+     */
+    private Vector2 contactPosition;
+
+
+    /**
+     * The force applied to the other entity.
+     */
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/destinationsol/events/ForceEvent.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.events;
+
+import com.badlogic.gdx.math.Vector2;
+import org.terasology.gestalt.entitysystem.event.Event;
+
+/**
+ * Event that represents a prolonged force applied to an entity, such as gravity. This does NOT represent anything
```
Adrijaned, line 22:
```suggestion
 * Event that represents a continuous force applied to an entity, like gravity. This does NOT represent anything
```

engine/src/main/java/org/destinationsol/events/ForceEvent.java
```diff
@@ -0,0 +1,49 @@
...
+ * that is a sudden, short force, such as a collision. That type of occurrence is handled by {@link ImpulseEvent}.
```
Adrijaned, line 23:
```suggestion
 * that is a sudden, immediate force, like collision. That type of occurrence is handled by {@link ImpulseEvent}.
```

engine/src/main/java/org/destinationsol/events/ForceEvent.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * The force applied to the entity.
+     */
+    private Vector2 force;
+
+    /**
+     * Whether the force applied causes acceleration.
+     */
+    private boolean causesAcceleration;
```
NicholasBatesNZ, line 35: More javadocs here please. Why is this a thing? F=ma so ideally a force will always cause acceleration.

engine/src/main/java/org/destinationsol/events/ForceEvent.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Event that represents a prolonged force applied to an entity, such as gravity. This does NOT represent anything
+ * that is a sudden, short force, such as a collision. That type of occurrence is handled by {@link ImpulseEvent}.
+ */
+public class ForceEvent implements Event {
+
+    /**
+     * The force applied to the entity.
+     */
```
Adrijaned, line 29: Could you javadoc the getters instead of the private fields? Applies to all events.

engine/src/main/java/org/destinationsol/events/ForceEvent.java
```diff
@@ -0,0 +1,49 @@
...
+    private Vector2 force;
+
+    /**
+     * Whether the force applied causes acceleration.
```
Adrijaned, line 33: This doesn't really tell me anything about what that means.

engine/src/main/java/org/destinationsol/events/ContactEvent.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    private EntityRef triggeringEntity;
+
+    /**
+     * The position where the contact happened.
+     */
+    private Vector2 contactPosition;
+
+    /**
+     * The impulse applied to the other entity.
```
Adrijaned, line 40: This seems wrong, maybe just needs more docs. But shouldn't impulse by definition be vector?

engine/src/main/java/org/destinationsol/events/ImpulseEvent.java
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+package org.destinationsol.events;
+
+import com.badlogic.gdx.math.Vector2;
+import org.terasology.gestalt.entitysystem.entity.EntityRef;
+
+/**
+ * Event that represents a sudden, short force applied to an entity, such as the impact of a collision. This does NOT
+ * represent anything that is a prolonged force, such as gravity. That type of occurrence is handled by
```
Adrijaned, line 23:
```suggestion
 * represent anything that is an application of continuous force, like gravity. That type of occurrence is handled by
```

engine/src/main/java/org/destinationsol/events/ImpulseEvent.java
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.events;
+
+import com.badlogic.gdx.math.Vector2;
+import org.terasology.gestalt.entitysystem.entity.EntityRef;
+
+/**
+ * Event that represents a sudden, short force applied to an entity, such as the impact of a collision. This does NOT
```
Adrijaned, line 22:
```suggestion
 * Event that represents a sudden, immediate force applied to an entity, like impact of a collision. This does NOT
```

engine/src/main/java/org/destinationsol/events/ImpulseEvent.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public class ImpulseEvent {
+
+    /**
+     * The position where the contact happened.
+     */
+    private Vector2 contactPosition;
+
+    /**
+     * The impulse applied to the entity.
```
Adrijaned, line 34: Shouldn't this be vector? See comment above.

engine/src/main/java/org/destinationsol/events/ForceEvent.java
NicholasBatesNZ: how about `instantaneous` rather than `immediate`?

engine/src/main/java/org/destinationsol/events/ImpulseEvent.java
NicholasBatesNZ: `instantaneous` again?

engine/src/main/java/org/destinationsol/events/ContactEvent.java
```diff
@@ -0,0 +1,59 @@
...
+    public ContactEvent(EntityRef triggeringEntity, Vector2 contactPosition, float absoluteImpulse) {
+        this.triggeringEntity = triggeringEntity;
+        this.contactPosition = contactPosition;
+        this.absoluteImpulse = absoluteImpulse;
+    }
+
+    /**
+     * The entity causing the contact event.
+     */
+    public EntityRef getTriggeringEntity() {
```
jellysnake, line 42: A contact would involve two entities would it not?
How is the entity causing the contact determined? Is it just arbitrary?

Does the entity that this event is sent against have any relevance? Is it the other entity in the contact?

engine/src/main/java/org/destinationsol/events/ContactEvent.java
```diff
@@ -0,0 +1,59 @@
...
+     * The position where the contact happened.
+     */
+    public Vector2 getContactPosition() {
+        return contactPosition;
+    }
+
+    /**
+     * The impulse applied to both entities.
+     */
+    public float getAbsoluteImpulse() {
```
jellysnake, line 56: It seems a little bit odd to me that this event has impulse stuff in it? Is this needed to be in this event or can it be removed and just use the impulse event below?

engine/src/main/java/org/destinationsol/events/ForceEvent.java
```diff
@@ -0,0 +1,48 @@
...
+package org.destinationsol.events;
+
+import com.badlogic.gdx.math.Vector2;
+import org.terasology.gestalt.entitysystem.event.Event;
+
+/**
+ * Event that represents a continuous force applied to an entity, like gravity. This does NOT represent anything
+ * that is a sudden, instantaneous force, like a collision. That type of occurrence is handled by {@link ImpulseEvent}.
+ */
+public class ForceEvent implements Event {
```
jellysnake, line 25: Because a force is applied to an object over a period of time, the lack of anything related to time is a tad confusing?

How would one use this event? Would you send it and the force is just applied forever? Do you send the event constantly until you want the force to end?

engine/src/main/java/org/destinationsol/events/ContactEvent.java
IsaacLic: I think that the contact handling can sometimes prevent an impulse event from happening. For example, if a SolShip comes into contact with a Loot object, it absorbs the loot without handling the impulse.

engine/src/main/java/org/destinationsol/events/ForceEvent.java
IsaacLic: Force events are sent repeatedly each timestep, for the duration of the time that the entity is being affected by the force.

PR #518, opened 2020-06-09
Stasis component for distant entities

IsaacLic:
New component for when an entity becomes too far from the player and should be put into stasis. For systems that shouldn't operate on an entity in stasis, there should be a method annotated with "@Before" that consumes that event if the entity has a stasis component.

### DIFF
--- a/engine/src/main/java/org/destinationsol/components/Stasis.java
+++ b/engine/src/main/java/org/destinationsol/components/Stasis.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2020 The Terasology Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/destinationsol/components/Stasis.java
```diff
@@ -0,0 +1,34 @@
...
+ * component is removed.
+ *
+ * For systems that shouldn't operate on an entity in stasis, there should be a method annotated with "@Before" that
+ * consumes that event if the entity has a stasis component.
+ */
+public class Stasis implements Component<Stasis> {
+
+    @Override
+    public void copy(Stasis other) {
+    }
```
NicholasBatesNZ, line 33: Does this need to be here?

IsaacLic: Yes - in order to implement Component, Java requires that all its methods be implemented, even if they're not doing anything

PR #519, opened 2020-06-12
Removal framework

IsaacLic:
This adds events to represent destruction of an entity and removal for optimization, as well as a default system to handle those events. This also adds a method to `EventSystemManager` to send an event directly to an entity.


engine/src/test/java/org/destinationsol/systems/DamageSystemTests/NonNegativeHealthTest.java
```diff
@@ -37,23 +39,23 @@
...
+        health.maxHealth = 50;
+        health.currentHealth = 50;
+        entity.setComponent(health);
+
         DamageEvent event = new DamageEvent(60);
 
         entitySystemManager.sendEvent(event, new Health());
 
-        assertEquals(0, entity.getComponent(Health.class).get().currentHealth);
+        assertEquals(0, health.currentHealth);
```
Adrijaned, line 58: Isn't this effectively an unspecified behavior? Since at this point the entity doesn't exist anymore?

IsaacLic: That's a good point

IsaacLic: Might there be a situation where an entity's health could drop to zero, but not be removed? For example, if someone designs a component that lets a player live for a few seconds after their health drops to zero, so that if they have a repair kit, they can still recover.

I know that that is niche, but I think that the removal is actually separate from the health handling. If a system stops the entity from dying, then it should still have zero health.

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -252,7 +251,7 @@ private void draw() {
             factionDisplay.drawFactionNames(solGame, uiDrawer, inputManager, solGame.getObjectManager());
         }
         if (fatalErrorMsg != null) {
-            uiDrawer.draw(uiDrawer.whiteTexture, displayDimensions.getRatio(), .5f, 0, 0, 0, .25f, 0, SolColor.UI_BG);
+            uiDrawer.draw(UiDrawer.whiteTexture, displayDimensions.getRatio(), .5f, 0, 0, 0, .25f, 0, SolColor.UI_BG);
```
NicholasBatesNZ, line 254: Did we figure out why this is here? Is this change useful?

engine/src/main/java/org/destinationsol/systems/DamageSystem.java
```diff
@@ -50,6 +55,10 @@ public EventResult onDamage(DamageEvent event, EntityRef entity) {
             }
             health.currentHealth = newHealthAmount;
             entity.setComponent(health);
+
+            if (health.currentHealth <= 0) {
```
NicholasBatesNZ, line 59: This whole block can be simplified, no? You're effectively checking for negative health twice now.

engine/src/main/java/org/destinationsol/systems/DefaultDestructionSystem.java
```diff
@@ -0,0 +1,48 @@
...
+        entity.delete();
+        return EventResult.COMPLETE;
+    }
+
+    /**
+     * Deletes an entity that no longer needs to exist.
+     */
+    @ReceiveEvent
+    public EventResult onRemovalForOptimization(RemovalForOptimizationEvent event, EntityRef entity) {
+        System.out.println("called!");
```
NicholasBatesNZ, line 44: Should this still be here?

engine/src/main/java/org/destinationsol/events/DestroyEvent.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Event that indicates that an entity should be removed.
+ * <p>
+ * An entity should usually be removed when its health reaches zero, or, for instance, if the object is a projectile,
+ * when it reaches its target. This event should not be used for entities being removed for optimization purposes when
+ * they get too far away. Removal for optimization purposes is handled by a {@link RemovalForOptimizationEvent}.
+ */
+
+public class DestroyEvent implements Event {
+
```
NicholasBatesNZ, line 29: Let's be consistent with newlines in empty classes please:)

engine/src/main/java/org/destinationsol/SolApplication.java
IsaacLic: Is seems that it's a static field, so the change effectively doesn't do anything. @Adrijaned might know, though - it was in his patch.

engine/src/main/java/org/destinationsol/systems/DamageSystem.java
IsaacLic: See my earlier reply to Adrijaned. I think that there are two different things that happen when an entity with 30 health takes 40 damage. 
1) The entity's health drops to 0 (so it shouldn't display -10 for the hero, for example)
2) a DestroyEvent is created
I think that the two things aren't intrinsically linked. If an entity doesn't get removed by the DestroyEvent, its health still should not be negative. Does that make sense? If so, I'll add some Javadoc explaining that.

engine/src/main/java/org/destinationsol/systems/DefaultDestructionSystem.java
IsaacLic: Nope! That was for debugging purposes XD

engine/src/main/java/org/destinationsol/SolApplication.java
Adrijaned: Probably just IDEA did something automatic

engine/src/main/java/org/destinationsol/systems/DamageSystem.java
Adrijaned: Nope, this block can't be simplified, since you need to set the Health component on the entity before sending the event

NicholasBatesNZ: Ok that makes sense, comments explaining that would be great :)

engine/src/main/java/org/destinationsol/components/SurvivesWithZeroHealth.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.destinationsol.components;
+
+import org.terasology.gestalt.entitysystem.component.Component;
+
+/**
+ * If an entity with a {@link Health} component that has zero health should not be destroyed, it should have
+ * this component.
+ */
+public class SurvivesWithZeroHealth implements Component<SurvivesWithZeroHealth> {
```
Adrijaned, line 24: I don't particularly like this. Generally, IMO, entities should have features based on what components they have. Here, however, you are basing feature on the _lack_ of component. Not only is this a bit ugly, it is also quite error-prone - it can be quite easy to forget about the existence of particular component.

engine/src/main/java/org/destinationsol/systems/DamageSystem.java
```diff
@@ -44,12 +49,14 @@ public EventResult onDamage(DamageEvent event, EntityRef entity) {
...
-            int newHealthAmount = health.currentHealth - event.getDamage();
-            if (newHealthAmount < 0) {
-                newHealthAmount = 0;
+            health.currentHealth -= event.getDamage();
+            if (health.currentHealth <= 0) {
+                health.currentHealth = 0;
+                entity.setComponent(health);
+                entitySystemManager.sendEvent(new ZeroHealthEvent(), entity);
+            } else {
+                entity.setComponent(health);
```
NicholasBatesNZ, line 58: I get this, but can we have a brief comment in the code explaining why `entity.setComponent(health);` has to be here twice please :)

NicholasBatesNZ:
#527

PR #522, opened 2020-06-19
ECS structure for using Body

IsaacLic:
This PR fixes issue #521 - it's an ECS-friendly way of interacting with the `Body` class. 

DS utilizes libGDX for its physics handling. The `Body` class corresponds, theoretically, to a component to be attached to an entity. However, the Body class contains functionality and is not serializable, which means that it can't be put into a component. 

This PR contains a system for handling the interactions between Bodies and entities. The system has a HashMap linking the two, and it processes the events that have to do with an entity`s Body.

PR #529, opened 2020-07-10
Loot dropping upon death

IsaacLic:
# Description
This PR adds a system for creating `Loot` objects when an entity with a `DropsLoot` component dies. `Loot` is the DS object that is created when certain `SolObject`s (e.g. ships) are destroyed. The logic was pulled from `Asteroid`, although `SolShip` uses a similar structure.

This is built upon PRs #519 and #522. It makes use of the `DestroyEvent` from 519, as well as `Position` and `Velocity` from 522.

# Testing
First, copy/paste the following three fields to SolApplication:
```
private boolean madeLootEntity = false;
    private int counter = 0;
    EntityRef entity;
```
Next, add the following code after line 231 (inside the `if` statement that confims that `solGame` isn't null):
```
            if (!madeLootEntity) {

            entity = entitySystemManager.getEntityManager().createEntity(new DropsLootOnDeath(), new Position(), new Velocity(), new Size());

            Position position = entity.getComponent(Position.class).get();
            position.position = solGame.getHero().getPosition().cpy();
            position.position.x += 1;
            position.position.y += 3;
            entity.setComponent(position);

            Size size = new Size();
            size.size = 20;
            entity.setComponent(size);

            madeLootEntity = true;
        }
        counter++;
        if (counter == 500) {
            entitySystemManager.sendEvent(new DestroyEvent(), entity);
        }
```

That code does a few things:
* Creates an entity with the requisite components to drop loot upon dying
* Moves the entity one unit down and three to the right, so that the loot won't be absorbed by the station
* After several seconds, the `EntitySystemManager` will send a `DestroyEvent` to the entity, causing loot to appear on screen

### DIFF
--- a/engine/src/main/java/org/destinationsol/loot/systems/LootDroppingSystem.java
+++ b/engine/src/main/java/org/destinationsol/loot/systems/LootDroppingSystem.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2020 The Terasology Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.loot.systems;
+
+import com.badlogic.gdx.math.Vector2;
+import org.destinationsol.common.In;
+import org.destinationsol.common.SolMath;
+import org.destinationsol.common.SolRandom;
+import org.destinationsol.entitysystem.EventReceiver;
+import org.destinationsol.game.SolGame;
+import org.destinationsol.game.item.Loot;
+import org.destinationsol.game.item.MoneyItem;
+import org.destinationsol.location.components.Position;
+import org.destinationsol.location.components.Velocity;
+import org.destinationsol.loot.components.DropsLootOnDeath;
+import org.destinationsol.removal.DefaultDestructionSystem;
+import org.destinationsol.removal.DestroyEvent;
+import org.destinationsol.size.components.Size;
+import org.terasology.gestalt.entitysystem.entity.EntityRef;
+import org.terasology.gestalt.entitysystem.event.Before;
+import org.terasology.gestalt.entitysystem.event.EventResult;
+import org.terasology.gestalt.entitysystem.event.ReceiveEvent;
+
+import java.util.List;
+
+/**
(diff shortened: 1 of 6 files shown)


engine/src/main/java/org/destinationsol/loot/systems/LootDroppingSystem.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @ReceiveEvent(components = {DropsLootOnDeath.class, Position.class, Velocity.class, Size.class})
+    @Before(DefaultDestructionSystem.class)
+    public EventResult onDestroy(DestroyEvent event, EntityRef entity) {
+
+        Vector2 basePosition = entity.getComponent(Position.class).get().position;
+        Vector2 baseVelocity = entity.getComponent(Velocity.class).get().velocity;
+        float size = entity.getComponent(Size.class).get().size;
+
+        float thrMoney = size * 40f * SolRandom.randomFloat(.3f, 1);
```
NicholasBatesNZ, line 68: What is `thrMoney`? Longer, descriptive names are better than short abbreviations.
Also, please comment on use of magic numbers :)

engine/src/main/java/org/destinationsol/loot/systems/LootDroppingSystem.java
```diff
@@ -0,0 +1,84 @@
...
+        List<MoneyItem> moneyItems = itemManager.moneyToItems(thrMoney);
+        for (MoneyItem item : moneyItems) {
+            float velocityAngle = SolRandom.randomFloat(180);
+            Vector2 lootVelocity = new Vector2();
+            SolMath.fromAl(lootVelocity, velocityAngle, SolRandom.randomFloat(0, Loot.MAX_SPD));
+            lootVelocity.add(baseVelocity);
+            Vector2 lootPosition = new Vector2();
+            SolMath.fromAl(lootPosition, velocityAngle, SolRandom.randomFloat(0, size / 2));
+            lootPosition.add(basePosition);
+            Loot l = lootBuilder.build(game, lootPosition, item, lootVelocity, Loot.MAX_LIFE, SolRandom.randomFloat(Loot.MAX_ROT_SPD), null);
```
NicholasBatesNZ, line 78: `l` is not a good variable name :P

engine/src/main/java/org/destinationsol/loot/systems/LootDroppingSystem.java
IsaacLic: I actually copy/pasted that code, so I have no idea what the magic numbers represent XD

engine/src/main/java/org/destinationsol/loot/systems/LootDroppingSystem.java
IsaacLic: That was also copy/pasted. I would never choose that as a name. I'll change it

skaldarnar:
I read "loot dropping" I post links:
- https://lostgarden.home.blog/2014/12/08/loot-drop-tables/
- https://gamedev.stackexchange.com/questions/140979/modern-loot-table-concepts

These are basically two resources I found while briefly looking into _loot tables_ for Terasology. Maybe they are helpful, maybe not - I honestly didn't look at the PR 🙈

Issue #455, opened 2019-12-16
UI Improvements


NicholasBatesNZ:
Our current UI system is not very easy to work with and frankly a bit of a bodge (think magic constants, 0-1 percentages of the screen height, etc.). As such it is not very fun, and rather difficult to work with.

### Option 1: Embed NUI from Terasology

This is the preferred option, however it will likely involve the most work.
[Terasology](https://github.com/MovingBlocks/Terasology) has a fantastic UI library, however it is still embedded within the main engine. Work must first be done to extract NUI from Terasology (see https://github.com/MovingBlocks/Terasology/issues/2307) and then DS would be adapted to use it.

- [x] Resolve dependencies in https://github.com/MovingBlocks/Terasology/issues/2307
- [x] Extract NUI into standalone library
- [x] Adapt Destination Sol to use it
- [x] Port over screens

### ~~Option 2: Revive uiOverhaul~~

There was an attempt at solving these issues which lies in the [`uiOverhaul`](https://github.com/MovingBlocks/DestinationSol/tree/uiOverhaul) branch, however it grew old and stale after months of not receiving the love it deserved (see conflicts in #443 and many screens were not ported over). There were also some [PRs into it](https://github.com/MovingBlocks/DestinationSol/pulls?&q=label%3AUIOverhaul+) (with the `uiOverhaul` label) to port over more screens but they were not seen to, under the expectation that `uiOverhaul` was dying.

This option involves reviving and finishing the `uiOverhaul` effort.

- [ ] ~~Rebase `uiOverhaul`~~
- [ ] ~~Fix any bugs~~
- [ ] ~~Merge the [PRs into it](https://github.com/MovingBlocks/DestinationSol/pulls?&q=label%3AUIOverhaul+) where appropriate~~
- [ ] ~~Fix any more bugs~~
- [ ] ~~Port over any remaining screens~~

### ~~Option 3: Start a new UI Overhaul~~

Start from scratch replacing the current UI system with something better.
 
### Todo before closing issue

- [ ] Delete `uiOverhaul` branch
- [ ] Delete `uiOverhaul` label

Cervator:
On a related note: it would be nice to support sliders for some of the UI settings after a conversion, such as for sound volume

CoffeeSF:
21:9 / 2560x1080
First-time startup in windowed mode

Video:
[https://cdn.discordapp.com/attachments/389093174921920512/733357135555067924/vid.mp4](url)

Images:

[...]

PR #536, opened 2020-07-17
Initial Integration of NUI (Relates to #455)

BenjaminAmos:
# Description
This pull request introduces the NUI UI library to Destination Sol, with the intention of eventually replacing the built-in solution. It also introduces Terasology's console UI into the game and adds a command to display NUI UI screens. This relates to #455.

I have re-created a series of themed widget graphics for Destination Sol, which are roughly based on the existing UI. You can change those or modify them as you see fit, or even replace them entirely. They were designed for testing NUI, as Terasology's UI graphics looked rather out of place in Destination Sol.

Apart from the console screen, none of the pre-existing UI screens have been ported yet. The NUI UI system and the built-in one co-exist, although they may occasionally conflict, as they use a shared `SpriteBatch` instance.
# Testing
##### Configuring NUI
 - Clone my [`nui-extraction-gestalt7`](https://github.com/BenjaminAmos/TeraNUI/tree/nui-extraction-gestalt7) branch and run the command `gradlew publishToMavenLocal` in the cloned project
##### Testing that NUI works
 - Start Destionation Sol and create a new game (any settings should work)
 - Press the ` key to open the console
 - Ensure that the console is visible and shows the correct text
 - Enter some input into the console (e.g. "help")
 - Ensure that a valid response is produced (e.g. a list of commands)
#### Testing NUI's widgets
 - The MigTestScreen from Terasology has been ported as it shows the UI's alignment capabilities, as well as a variety of widgets. You can view it using the command `showNUIScreen engine:migTestScreen`.
 - Ensure that the widgets display correctly (you may have to use Terasology as a reference)
 - Ensure that the widgets can be interacted with correctly (you may have to use Terasology as a reference)
#### Testing that NUI scales
One of the major issues with the existing UI system is that it does not scale very well. UI elements often appear out of alignment when the resolution is changed. The new NUI implementation should not suffer as much from this issue.
 - Follow the instructions from the prior two sections.
 - Close the console and test screen using the `Escape` key.
 - Open the game menu (press `Escape` again) and go back to the main menu.
 - Go into the settings (options) and set change the screen resolution.
 - Continue from your previous game and follow the instructions from the `Testing that NUI works` and `Testing NUI's widgets` sections again. Ensure that the UI appears visually similar and is still functional.
##### Testing with non-discrete resolutions (Windowed mode only)
 - Follow the same instructions as above but instead of changing the resolution in the options menu, open the in-game pause menu and drag the window corners to change the size of the window (or however you would do so natively in your OS).
 - The console UI should still scale correctly, as well as the text within it.
### Testing UI scaling
 - When the game starts, open the `Options->Resolution` menu and press the `NUI UI Scale` button to change the UI scale.
 - Start a game (either create a new one or continue an existing one)
 - Open the console using the ` key
 - The console UI should appear either larger or smaller. The NUI widgets are either being upscaled or downscaled to the target scale factor.
 - Scale factors below 1 should increase the screen size. Scale factors above 1 should decrease it. The screen is scaled from a resolution of `target * scale` to the target resolution.
## Multi-platform testing
 - If you could test these changes on an Android device or emulator, then that would be helpful, as I have not tested these changes on Android extensively. So far as I know though they superficially work though. There may be difficulties using the UI on a touch screen though, as it is fairly desktop-oriented at the moment.
 - I haven't tested this on MacOS or Linux, although they should work with it.
### Reference Renderings
I have taken the following screenshots to show how the NUI UI appears on my set-up. If the appearance differs from these then there might be a rendering bug that I was not aware of.

ConsoleScreen
MigTestScreen
AndroidConsoleScreen
AndroidMigTestScreen

# Notes
 - As with Terasology, NUI operates independantly of the entity-system.
 - The only UI screens that have been ported so-far are the console UI, which I moved from Terasology, and the `MigTestScreen`, as it's a very good test of how the widgets render.
 - This relates to MovingBlocks/Terasology#4017 (which removes the shared NUI code from Terasology) but does not depend on it.
 - The extracted NUI code used here is almost entirely the same as in MovingBlocks/TeraNUI#2 but contains a couple of small commits on-top to make it compatible with gestalt 7 (changes [here](https://github.com/BenjaminAmos/TeraNUI/compare/nui-extraction-merge...BenjaminAmos:nui-extraction-gestalt7)). The branch was re-based on-top of the main `nui-extraction-merge` before submitting this pull request.

# Known Issues
## Minor Issues
 - The console UI does not properly scroll at the moment. This is due to the output history buffer being too short.
 - The console UI wraps text too early. I think this is due to the console implementation, rather than NUI itself, as it does not seem affected by UI scale.
 - The UI does not always scale to very small screens. The layout will sometimes be horizontally truncated. This is most prominent when testing the UI under very small resolutions, or on Android.
## Major issues
 - Currently this pull request depends on the use of the `mavenLocal` repository, as the NUI library needs to be built from source for now.
 - The code is not very well documented yet. I will add some JavaDoc to the new classes/methods soon.
 - The NUI reflection work-around in gradle includes `javax.servlet` into the build, which should not be a compile-time dependency, however it was needed to get the code to work.
 - The textures in the `engine/src/main/resources/org/destinationsol/assets/textures/ui/nui/terasology` folder are taken directly from Terasology, when Destination Sol equivalents should be used instead. The UI theme should be consistent, although most of the UI graphics there are unused at the moment.


Cervator:
Thank you once more for being our friendly neighborhood code Santa showing up bearing wondrous gifts at random times. While this is the first comment on the PR there have been a bunch on Discord already, not sure if you see those but there have been a lot of banter about how amazing all this is 😁 

There is now a TeraNUI build in Jenkins http://jenkins.terasology.io/teraorg/job/Libraries/job/TeraNUI/ and available at `1.0.0-SNAPSHOT` in Artifactory, going with https://github.com/MovingBlocks/Terasology/pull/4017

I added some commits that reworked the Gradle in TeraNUI and hooked up a Jenkinsfile. You up for adding another PR with the three commits you've got on top, bumping the version there to `2.0.0-SNAPSHOT` ? I figure we can keep the two version lines active - 1.x for TS, 2.x for DS (since newer Gestalt requirement - TS isn't ready for that yet). Soon as we merge a `2.0.0-SNAPSHOT` we could use that with this PR.

I haven't personally tested this one although it sounds like a few of the others are. I did manage to fully test the setup for TS though, and this looks equally superb - your PR details are at a whole other level, as usual 👍

BenjaminAmos:
I've updated my `nui-extraction-gestalt7` branch to use the version `2.0.0-SNAPSHOT` now (in MovingBlocks/TeraNUI#4), so the latest commits changed the version number used here. If anyone had previously tested this then you might have to pull the latest changes from my branch and run `gradlew publishToMavenLocal` in the `TeraNUI` repository again to get the updated version. Nothing should have changed apart from the NUI version number.

Cervator:
FYI for testing this PR: the TEraNUI `2.0.0-SNAPSHOT` binary is now available for testing in Artifactory.

engine/src/main/java/org/destinationsol/ui/nui/NUIManager.java
```diff
@@ -0,0 +1,321 @@
...
+     */
+    private Deque<NUIScreenLayer> uiScreens = new LinkedList<>();
+
+    private static final String WHITE_TEXTURE_URN = "engine:uiWhiteTex";
+    private static final String DEFAULT_SKIN_URN = "engine:default";
+    private static final String BUTTON_CLICK_URN = "engine:uiHover";
+    /**
+     * The value 0.9 was found from {@link org.destinationsol.ui.SolInputManager#playClick}, so it was copied here to
+     * retain the same click sound as the built-in UI.
+     */
```
NicholasBatesNZ, line 99: Thank you for this comment. There are far too many magic numbers in the codebase :roll_eyes:

engine/src/main/java/org/destinationsol/ui/nui/NUIManager.java
```diff
@@ -0,0 +1,321 @@
...
+        canvasRenderer.resize(width, height);
+    }
+
+    /**
+     * Sets the UI scale to a specified value. This will internally render the UI at specified scale and up-scale it
+     * to the target resolution.
+     * @param scale the new UI scale to use
+     */
+    public void setUiScale(float scale) {
+        canvas.setUiScale(1.0f / scale);
```
NicholasBatesNZ, line 318: To me it makes sense to have bigger elements when the scale is bigger. Is there any reason you choose to invert this?

engine/build.gradle
```diff
@@ -31,6 +41,9 @@ dependencies {
     compile(group: 'org.terasology.gestalt', name: 'gestalt-entity-system', version: '7.0.5')
     compile(group: 'org.terasology.gestalt', name: 'gestalt-module', version: '7.0.5')
     compile(group: 'org.terasology.gestalt', name: 'gestalt-util', version: '7.0.5')
+    compile group: 'org.terasology.nui', name: 'nui', version: '2.0.0-SNAPSHOT'
+    compile group: 'org.terasology.nui', name: 'nui-libgdx', version: '2.0.0-SNAPSHOT'
+    compile group: 'org.terasology', name: 'TeraMath', version: '1.5.0'
```
NicholasBatesNZ, line 46: Did we end up using TeraMath after all?

engine/src/main/java/org/destinationsol/ui/nui/NUIManager.java
BenjaminAmos: It was inverted mostly to reflect what was occurring internally (in [`LibGDXCanvasRenderer`](https://github.com/MovingBlocks/TeraNUI/blob/e1f7918018061864c99a72ef291f76299cb754d8/nui-libgdx/src/main/java/org/terasology/nui/backends/libgdx/LibGDXCanvasRenderer.java)), where the UI is being rendered at a resolution that is `scale*actual` and then scaled to match the actual resolution. From a visual perspective though, if the internal render resolution is less than the actual output resolution then the UI would appear larger. If the internal render resolution is greater then the UI would appear smaller.

I've changed it now though so that the values are more intuitive from a user-facing perspective (e.g. how it visibly affects the output).

engine/src/main/java/org/destinationsol/game/screens/MainGameScreen.java
```diff
@@ -289,12 +294,33 @@ public void updateCustom(SolApplication solApplication, SolInputManager.InputPoi
...
         }
 
+        if (consoleScreen.isConsoleJustClosed()) {
+            game.setPaused(false);
+            controls.forEach(x -> x.setEnabled(true));
+            consoleControlGrave.setEnabled(true);
+            consoleControlF1.setEnabled(true);
+
+            menuControl.update(inputPointers, false, false, inputMan, solApplication);
+            menuControl.update(inputPointers, false, false, inputMan, solApplication);
```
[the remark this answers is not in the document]
BenjaminAmos: I think that was needed when I was testing the NUI-based console early-on but removing it doesn't seem to make any difference so both those lines have now been removed. I believe it used to resolve certain input-focus issues but since the game is now paused when the console is open it isn't needed anymore.

engine/build.gradle
BenjaminAmos: TeraMath shouldn't have been used, so that was probably included in a previous commit (from this pull request) and then left there. Originally the NUI border class would have been in TeraMath but it was moved directly into NUI instead, so TeraMath is not needed. I have now removed the dependency.

Adrijaned (approved):
The time has come to finally merge this thing, for if we don't now, I reckon we might not at all in the near future! (We are terrible at forcing ourselves to actually do things :sweat_smile: )

PR #541, opened 2020-07-30
Working ECS-based graphics

IsaacLic:
# Description
This PR contains working ECS-based graphics. `SolApplication` implements `ApplicationListener`, which is a libGDX interface with a `render()` method that is called when the application should render itself. On every call to that method, a `RenderEvent` is sent to each entity with a `Graphics` component. The component contains a list of `GraphicsElements`, which have the information necessary for rendering the textures that the sprite is composed of. The `GraphicsRenderer` is the system that draws the entities when it receives a `RenderEvent`.

# Testing
After line 241 in SolApplication (within the `if` statement), add the following block of code:

```
if (!entityCreated) {
        GraphicsElement element = new GraphicsElement();
        element.texture = SolRandom.randomElement(Assets.listTexturesMatching("engine:asteroid_.*"));
        element.relativePosition = new Vector2(0, 0);
        element.drawableLevel = DrawableLevel.BODIES;
        element.width = 2;
        element.height = 2;
        element.tint = Color.YELLOW;
        Graphics graphicsComponent = new Graphics();
        graphicsComponent.elements.add(element);

        Position position = new Position();
        position.position = solGame.getHero().getShip().getPosition().cpy();
        EntityRef entityRef = entitySystemManager.getEntityManager().createEntity(graphicsComponent, position, new Angle());
        entityRef.setComponent(graphicsComponent);
        position.position.y += 1;
        entityRef.setComponent(position);
        entityCreated = true;
}
```

Then add the following field anywhere:
`private boolean entityCreated = false;`

That code creates an entity with the graphics of a gold-colored asteroid, which is positioned right below the player. When the game is started, the asteroid should be visible.

### DIFF
--- a/engine/src/main/java/org/destinationsol/SolApplication.java
+++ b/engine/src/main/java/org/destinationsol/SolApplication.java
@@ -26,11 +26,12 @@ import org.destinationsol.assets.music.OggMusicManager;
 import org.destinationsol.assets.sound.OggSoundManager;
 import org.destinationsol.common.SolColor;
 import org.destinationsol.common.SolMath;
+import org.destinationsol.rendering.components.Renderable;
+import org.destinationsol.rendering.events.RenderEvent;
 import org.destinationsol.entitysystem.ComponentSystemManager;
 import org.destinationsol.entitysystem.EntitySystemManager;
 import org.destinationsol.entitysystem.SerialisationManager;
 import org.destinationsol.game.DebugOptions;
-import org.destinationsol.game.FactionInfo;
 import org.destinationsol.game.ObjectManager;
 import org.destinationsol.game.SaveManager;
 import org.destinationsol.game.SolGame;
@@ -40,6 +41,7 @@ import org.destinationsol.game.context.Context;
 import org.destinationsol.game.context.internal.ContextImpl;
 import org.destinationsol.game.item.ItemManager;
 import org.destinationsol.game.item.LootBuilder;
+import org.destinationsol.location.components.Position;
 import org.destinationsol.menu.MenuScreens;
 import org.destinationsol.menu.background.MenuBackgroundManager;
 import org.destinationsol.modules.ModuleManager;
@@ -230,8 +232,13 @@ public class SolApplication implements ApplicationListener {
     private void draw() {
         Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
         commonDrawer.begin();
+
         if (solGame != null) {
+
             solGame.draw();
+
+            //This event causes each entity with a `Graphics` component to be rendered onscreen
+            entitySystemManager.sendEvent(new RenderEvent(), new Renderable(), new Position());
         }
         uiDrawer.updateMtx();
         inputManager.draw(uiDrawer, this);
--- a/engine/src/main/java/org/destinationsol/SolApplication.java
+++ b/engine/src/main/java/org/destinationsol/SolApplication.java
@@ -234,10 +234,9 @@ public class SolApplication implements ApplicationListener {
         commonDrawer.begin();
 
         if (solGame != null) {
-
             solGame.draw();
 
-            //This event causes each entity with a `Graphics` component to be rendered onscreen
+            //This event causes each entity with a `Renderable` component to be rendered onscreen
             entitySystemManager.sendEvent(new RenderEvent(), new Renderable(), new Position());
         }
         uiDrawer.updateMtx();
--- a/engine/src/main/java/org/destinationsol/game/SolGame.java
+++ b/engine/src/main/java/org/destinationsol/game/SolGame.java
@@ -25,6 +25,7 @@ import org.destinationsol.SolApplication;
 import org.destinationsol.assets.sound.OggSoundManager;
 import org.destinationsol.assets.sound.SpecialSounds;
 import org.destinationsol.common.DebugCol;
+import org.destinationsol.common.In;
 import org.destinationsol.common.SolException;
 import org.destinationsol.common.SolMath;
 import org.destinationsol.common.SolRandom;
@@ -60,6 +61,7 @@ import org.destinationsol.ui.DebugCollector;
 import org.destinationsol.ui.TutorialManager;
 import org.destinationsol.ui.UiDrawer;
 import org.destinationsol.ui.Waypoint;
+import org.destinationsol.util.InjectionHelper;
 import org.terasology.gestalt.entitysystem.entity.EntityRef;
 
 import java.util.ArrayList;
@@ -107,6 +109,8 @@ public class SolGame {
     private SortedMap<Integer, List<UpdateAwareSystem>> onPausedUpdateSystems;
     private SortedMap<Integer, List<UpdateAwareSystem>> updateSystems;
 
+    @In
+    private EntitySystemManager entitySystemManager;
 
     public SolGame(String shipName, boolean isTutorial, boolean isNewGame, CommonDrawer commonDrawer, Context context,
                    WorldConfig worldConfig) {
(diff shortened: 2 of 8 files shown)


engine/src/main/java/org/destinationsol/drawable/components/Graphics.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.destinationsol.drawable.GraphicsElement;
+import org.terasology.gestalt.entitysystem.component.Component;
+
+import java.util.ArrayList;
+
+/**
+ * Contains a list of the different elements that combine to visually represent an entity.
+ */
+public final class Graphics implements Component<Graphics> {
```
Cervator, line 26: Suggestion: Would `Renderable` be a better name? Sometimes it can be nice to name components after verbs since they're treats. Easier to say "That entity is renderable" vs "That entity is graphics" or even "that entity _has_ graphics" - plus I think "Graphics" is somewhat of an ambiguous term.

Inspiration as usual taken from Terasology, where admittedly it gets a bit more complicated since 3D and so on. So there just one of the things is `MeshComponent` but there are a bunch more and even variants of Mesh.

engine/src/main/java/org/destinationsol/drawable/GraphicsElement.java
```diff
@@ -0,0 +1,70 @@
...
+ * limitations under the License.
+ */
+package org.destinationsol.drawable;
+
+import com.badlogic.gdx.graphics.Color;
+import com.badlogic.gdx.graphics.g2d.TextureAtlas;
+import com.badlogic.gdx.math.Vector2;
+import org.destinationsol.game.drawables.DrawableLevel;
+
+public class GraphicsElement {
```
Cervator, line 23: Javadoc at the class level for everything would be very much appreciated and super useful. There is so much old Terasology code that has no class-level javadoc and over time people forget what the things do. Here we're still at the very beginning and can start out everything right :-)

engine/src/main/java/org/destinationsol/drawable/components/Invisibility.java
```diff
@@ -0,0 +1,30 @@
...
+package org.destinationsol.drawable.components;
+
+import org.destinationsol.drawable.systems.GraphicsRenderer;
+import org.destinationsol.game.GameDrawer;
+import org.terasology.gestalt.entitysystem.component.Component;
+
+/**
+ * Denotes that the object should not be drawn by the {@link GraphicsRenderer}.
+ */
+public class Invisibility implements Component<Invisibility> {
```
Cervator, line 25: How about `Invisible` as opposed to `Invisibility` ?

engine/src/main/java/org/destinationsol/drawable/systems/GraphicsRenderer.java
```diff
@@ -0,0 +1,68 @@
...
+import org.destinationsol.location.components.Angle;
+import org.destinationsol.location.components.Position;
+import org.terasology.gestalt.entitysystem.entity.EntityRef;
+import org.terasology.gestalt.entitysystem.event.EventResult;
+import org.terasology.gestalt.entitysystem.event.ReceiveEvent;
+
+/**
+ * This handles the drawing of each entity with a {@link Graphics} component.
+ */
+public class GraphicsRenderer implements EventReceiver {
```
Cervator, line 36: I still need to learn what makes the distinction in the newest Gestalt of an even receiver vs an entity system class. So used to everything with logic being a `System` leading to easy names like "RenderingSystem"

Maybe leaving `Component` off from actual component classes leads to some benefits, but others like event classes still seem to end in `-Event` - I wonder if it would make sense to keep a focus on `System` classes too.  I just really don't know why event handlers and systems are two different things now 🤔 

This isn't on you as much as me just not catching some subtleties or knowing about any sort of reasoning things were split out, yet. Just everything in my head screams that this is a system class :-)

engine/src/main/java/org/destinationsol/drawable/GraphicsElement.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Represents the texture of this graphics element.
```
Cervator, line 26: More of a personal preference than a directive or anything, but I like being able to use the one-line approach for documenting fields - at least while the javadoc is fairly short. So that'd simply mean:

`/** Represents the texture of this graphics element. */

engine/src/main/java/org/destinationsol/game/drawables/RectSprite.java
```diff
@@ -70,8 +70,13 @@ public RectSprite(TextureAtlas.AtlasRegion tex, float texSz, float origPercX, fl
...
     }
 
+    /**
+     * Sets the larger of the dimensions of the texture to the passed-in size, maintaining proportions and
+     * recalculating other fields as needed to fit.
+     * @param textureSize the new size for the sprite
+     */
     public void setTextureSize(float textureSize) {
-        textureSize /= level.depth;
+        textureSize /= level.depth; /** Scales the textureSize for background objects */
```
Cervator, line 79: This on the other hand would *not* be javadoc, since it is in the middle of a method. Good place for a plain old `//` comment at the end of the line, or on the preceding line.

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -230,8 +232,13 @@ private void update() {
     private void draw() {
         Gdx.gl.glClear(GL20.GL_COLOR_BUFFER_BIT);
         commonDrawer.begin();
+
         if (solGame != null) {
+
```
NicholasBatesNZ, line 237: The newline above is meaningful. I don't think that this one is however

engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -107,6 +109,8 @@
     private SortedMap<Integer, List<UpdateAwareSystem>> onPausedUpdateSystems;
     private SortedMap<Integer, List<UpdateAwareSystem>> updateSystems;
 
+    @In
+    private EntitySystemManager entitySystemManager;
```
NicholasBatesNZ, line 113: Is this needed? Is it even going to work? 🤔

engine/src/main/java/org/destinationsol/game/drawables/RectSprite.java
```diff
@@ -70,8 +70,13 @@ public RectSprite(TextureAtlas.AtlasRegion tex, float texSz, float origPercX, fl
...
     }
 
+    /**
+     * Sets the larger of the dimensions of the texture to the passed-in size, maintaining proportions and
+     * recalculating other fields as needed to fit.
+     * @param textureSize the new size for the sprite
+     */
     public void setTextureSize(float textureSize) {
-        textureSize /= level.depth;
+        textureSize /= level.depth; // Scales the textureSize for background objects
```
NicholasBatesNZ, line 79: Thanks for this 👍

engine/src/main/java/org/destinationsol/rendering/RenderableElement.java
```diff
@@ -0,0 +1,59 @@
...
+    public Color tint;
+
+    public void copy(RenderableElement other) {
+        this.texture = other.texture;
+        this.drawableLevel = other.drawableLevel;
+        this.relativePosition = other.relativePosition.cpy();
+        this.relativeAngle = other.relativeAngle;
+        this.width = other.width;
+        this.height = other.height;
+        this.tint = other.tint;
```
NicholasBatesNZ, line 56: Recheck some of these please. In Java, all non-primitives are just references, so these lines will not be making deep copies unless you do something like `.cpy()` which you have done for `relativePosition`.
`width`, `height`, `relativeAngle` and `relativePosition` will be deep copies with this code.
`texture`, `drawableLevel`, and `tint` should all have `copy()` methods

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -230,8 +232,13 @@ private void update() {
...
             solGame.draw();
+
+            //This event causes each entity with a `Graphics` component to be rendered onscreen
```
NicholasBatesNZ, line 240: `Graphics` component?

engine/src/main/java/org/destinationsol/rendering/RenderableElement.java
IsaacLic: DrawableLevel is an enum, so it doesn't need to be copied. I added copying for `texture` and `tint`

PR #543, opened 2020-08-02
Asteroid implementation

IsaacLic:
# Description
This PR contains an implementation for ECS-based asteroids. As of now, most of the asteroids in-game are still the old OOP asteroids, but there is a golden asteroid that uses the new component-based structure.

When the ship collides with the golden asteroid, they both take damage, the same way they would if the asteroid was a regular one. When the golden asteroid dies, it creates Money objects. As of now, the asteroid cannot be shot, because Projectiles don't collide with entities. I'm working on that now.

# Testing
The testing code is currently built in - if you start the game, there will be a golden asteroid visible that has a working collision mesh. If you ram into it, you'll take damage as usual. I set its health to one, so to destroy it, ram into a a few times from far away or use a Loaded Imperial Large ship (they do a lot of collision damage).  It should drop money upon being destroyed. As of now, I haven't developed collision with `Projectiles`, so shooting it won't work. 

# Outstanding Work
I need to add damage events to asteroids that get shot, and I need to move the body creation into a generic, non-asteroid class. The collision mesh is also slightly misaligned, so it needs to be fixed. In addition, I need to have asteroids produce rubble when destroyed. Once those things are done, I'll replace the existing asteroids with ECSteroids.


engine/src/main/java/org/destinationsol/game/SolContactListener.java
```diff
@@ -55,21 +80,44 @@ public void preSolve(Contact contact, Manifold oldManifold) {
...
+        Object dataA = contact.getFixtureA().getBody().getUserData();
+        Object dataB = contact.getFixtureB().getBody().getUserData();
+
+        Vector2 collPos = contact.getWorldManifold().getPoints()[0];
+        float absImpulse = calcAbsImpulse(impulse);
+
+        if (dataA instanceof EntityRef) {
+            entitySystemManager.sendEvent(new ImpulseEvent(collPos, absImpulse), (EntityRef) dataA);
+
+            //This is a patch to smooth over contact between an Entity and a SolObject. Once every SolObject has been
```
Cervator, line 93: This sounds like it deserves a `TODO` at the start to make it formal and allow code analytics to catch on to it :-)

engine/src/main/java/org/destinationsol/game/SolObjectEntityWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+import com.badlogic.gdx.math.Vector2;
+import org.destinationsol.entitysystem.EntitySystemManager;
+import org.destinationsol.game.drawables.Drawable;
+import org.terasology.gestalt.entitysystem.entity.EntityRef;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * This is a {@link SolObject} wrapper for an {@link EntityRef}. This is is a patch for the in-between period between
```
Cervator, line 27: This is great javadoc 👍 Having this sort of thing for each big important class would be awesome.

Since this is meant to be temporary you might want to throw a TODO into the javadoc or even `@deprecate` it to indicate its doomed nature

engine/src/main/java/org/destinationsol/removal/components/SlatedForDeletion.java
```diff
@@ -0,0 +1,36 @@
...
+package org.destinationsol.removal.components;
+
+import com.badlogic.gdx.physics.box2d.Body;
+import org.destinationsol.removal.events.DeletionEvent;
+import org.destinationsol.removal.systems.DeletionUpdateSystem;
+import org.destinationsol.removal.systems.DestructionSystem;
+import org.terasology.gestalt.entitysystem.component.Component;
+
+/**
+ * Indicates that an entity can safely be deleted.
```
Cervator, line 25: Also great javadoc. Nice work 👍

engine/src/main/java/org/destinationsol/removal/components/SlatedForDeletion.java
```diff
@@ -0,0 +1,36 @@
...
+ * <p>
+ * Every tick, the {@link DeletionUpdateSystem} sends a {@link DeletionEvent} to each entity with a SlatedForDeletion
+ * component, which causes the entity to actually be deleted by the {@link DestructionSystem}. This is done so that
+ * certain resources, such as libGDX's {@link Body} class, can be disposed of properly before the entity is deleted.
+ */
+public class SlatedForDeletion implements Component<SlatedForDeletion> {
```
Cervator, line 31: Should this be another `EmptyComponent` ?

engine/src/main/java/org/destinationsol/removal/systems/DestructionSystem.java
```diff
@@ -0,0 +1,65 @@
...
+import org.destinationsol.removal.events.RemovalForOptimizationEvent;
+import org.terasology.gestalt.entitysystem.entity.EntityRef;
+import org.terasology.gestalt.entitysystem.event.EventResult;
+import org.terasology.gestalt.entitysystem.event.ReceiveEvent;
+
+import java.util.HashSet;
+
+/**
+ * When an entity receives a {@link ShouldBeDestroyedEvent} or a {@link RemovalForOptimizationEvent}, this system adds a
+ * {@link SlatedForDeletion} component. Any system that should respond to the destruction of an entity before it happens
```
Cervator, line 31: Does this at all get into the priority levels available on events? That's another thing I'm used to from TS (make events _high priority_ if they're meant to get processed before other key events) - not totally sure if it is a thing in latest Gestalt and/or DS

PR #544, opened 2020-08-03
Cleaning up SolGame, part one

IsaacLic:
# Description
SolGame currently creates approximately thirty instances of different classes, and the main way that those classes are accessed are through thirty different getters. Besides for the architectural issues of having a god class, it can create practical issues. For example, the EntitySystemManager needs to be created after SolGame, because some systems created by the ESM use the classes stored in SolGame. However, since UpdateAwareSystems are created in the constructor of SolGame, the ESM can't be injected into those systems. 

The purpose of this PR is to neatly repackage the objects that SolGame currently creates, but that don't need to be part of SolGame. It does so by putting the objects into the `context` and replacing the getters for those objects with `context.get(Object.class)`. This is first of two or three PRs that will do so. I'm breaking it up into multiple PRs to make review more manageable.

# Testing
This is just code cleanup, so the game should function exactly the same way as it did before.

# Outstanding Work
The public getters for rest of the fields in `SolGame` need to be refactored to use `Context`, unless there is a compelling argument for that field being an attribute of SolGame.


engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -117,17 +115,33 @@ public SolGame(String shipName, boolean isTutorial, boolean isNewGame, CommonDra
         // TODO: make this non-static
         FactionInfo.init();
 
+        this.context = context;
+
         this.isTutorial = isTutorial;
         solApplication = context.get(SolApplication.class);
-        ModuleManager moduleManager = context.get(ModuleManager.class);
+
+        //This no longer needs to be instantiated in SolGame
```
Adrijaned, line 123: Should this perhaps be marked TODO? At least that's how the comment reads (Also applies to the following)

Adrijaned (approved):
Code definitely looks like a step in a right direction, if someone can playtest this, could be merged.

engine/src/main/java/org/destinationsol/game/ObjectManager.java
```diff
@@ -44,12 +45,13 @@
     private final World myWorld;
     private final Box2DDebugRenderer myDr;
     private final HashMap<SolObject, Float> myRadii;
+    private final Context myContext;
```
NicholasBatesNZ, line 48: We're trying to phase out all the `myX` variables so let's just call this `context` please.

engine/src/main/java/org/destinationsol/game/ObjectManager.java
```diff
@@ -189,7 +192,7 @@ private void removeObjNow(SolGame game, SolObject o) {
         myObjs.remove(o);
         myRadii.remove(o);
         o.onRemove(game);
-        game.getDrawableManager().removeObject(o);
+        game.getContext().get(DrawableManager.class).removeObject(o);
```
NicholasBatesNZ, line 195: Is there any reason you use `game.getContext()` here rather than `myContext`?

engine/src/main/java/org/destinationsol/game/ObjectManager.java
```diff
@@ -198,7 +201,7 @@ public void addObjNow(SolGame game, SolObject o) {
         }
         myObjs.add(o);
         recalcRadius(o);
-        game.getDrawableManager().addObject(o);
+        game.getContext().get(DrawableManager.class).addObject(o);
```
NicholasBatesNZ, line 204: Same question here

engine/src/main/java/org/destinationsol/game/screens/MainGameScreen.java
```diff
@@ -673,7 +675,7 @@ protected boolean shouldWarn(SolGame game) {
             float heroCap = HardnessCalc.getShipDmgCap(hero.getShip());
             List<SolObject> objs = game.getObjectManager().getObjects();
             FactionManager fm = game.getFactionMan();
-            SolCam cam = game.getCam();
+            SolCam cam = game.getContext().get(SolCam.class);
```
NicholasBatesNZ, line 678: and here

engine/src/main/java/org/destinationsol/game/screens/MainGameScreen.java
```diff
@@ -446,7 +448,7 @@ public void drawImages(UiDrawer uiDrawer, SolApplication solApplication) {
         myMoneyExcessTp.text = null;
 
         maybeDrawHeight(uiDrawer);
-        borderDrawer.draw(uiDrawer, solApplication);
+        borderDrawer.draw(uiDrawer, solApplication, solApplication.getContext());
```
NicholasBatesNZ, line 451: Here too. Is the SolApp context different? Maybe it should be... :thinking:

engine/src/main/java/org/destinationsol/game/ObjectManager.java
IsaacLic: Nope - that was a mistake. Fixing it now

engine/src/main/java/org/destinationsol/game/screens/MainGameScreen.java
IsaacLic: This one is actually in a private static class, so it doesn't have access to the enclosing class's reference

engine/src/main/java/org/destinationsol/game/screens/MainGameScreen.java
IsaacLic: It's the same as the SolGame context for now. I think #586 might makes some changes to that, but this should be independent of that.

PR #546, opened 2020-08-06
Gestalt update

IsaacLic:
# Description
This PR updates the gestalt version from `7.0.5` to `7.0.6-SNAPSHOT`. The new version fixes a bug that prevented the creation of more than 1024 entities, and it also adds an `EmptyComponent` class for components that don't contain data. This PR also refactors the existing data-less components to use `EmptyComponent`.

# Testing
* Run `gradlew clean cleanIdea idea jar` in the terminal
* Start the game. It should run normally.

### DIFF
--- a/config/gradle/common.gradle
+++ b/config/gradle/common.gradle
@@ -43,6 +43,8 @@ repositories {
     maven { url "https://jitpack.io" }
 
     maven { url "https://maven.google.com" }
+
+    maven { url = 'https://heisluft.tk/maven/' }
 }
 
 test {
--- a/engine/src/main/java/org/destinationsol/entitysystem/EntitySystemManager.java
+++ b/engine/src/main/java/org/destinationsol/entitysystem/EntitySystemManager.java
@@ -34,6 +34,7 @@ import org.terasology.gestalt.entitysystem.event.impl.EventSystemImpl;
 import org.terasology.gestalt.entitysystem.prefab.GeneratedFromRecipeComponent;
 import org.terasology.gestalt.module.ModuleEnvironment;
 
+import java.lang.reflect.Modifier;
 import java.util.List;
 
 public class EntitySystemManager {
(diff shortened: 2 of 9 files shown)


config/gradle/common.gradle
```diff
@@ -43,6 +43,8 @@ repositories {
     maven { url "https://jitpack.io" }
 
     maven { url "https://maven.google.com" }
+
+    maven { url = 'https://heisluft.tk/maven/' }
```
NicholasBatesNZ, line 47: Please include a comment on why this is here

engine/src/test/java/org/destinationsol/gestalt/BulkEntityCreationStressTest.java
```diff
@@ -0,0 +1,46 @@
...
+    public void setUp() throws Exception {
+        ModuleManager moduleManager = new ModuleManager();
+        moduleManager.init();
+        entitySystemManager = new EntitySystemManager(moduleManager.getEnvironment(), new ComponentManager(), new ContextImpl());
+    }
+
+    @Test
+    public void testBulkEntityCreation() {
+        for (int i = 0; i < 100000; i++) {
+            entitySystemManager.getEntityManager().createEntity();
```
NicholasBatesNZ, line 43: This test isn't particularly relevant to Destination Sol as it is. Perhaps you could make it more unique by creating a few thousand asteroids and blowing them up? You may want to split that into multiple tests

engine/src/main/java/org/destinationsol/entitysystem/EntitySystemManager.java
```diff
@@ -48,8 +49,10 @@ public EntitySystemManager(ModuleEnvironment environment, ComponentManager compo
 
         List<ComponentStore<?>> stores = Lists.newArrayList();
         for (Class<? extends Component> componentType : environment.getSubtypesOf(Component.class)) {
-            stores.add(
-                    new ConcurrentComponentStore<>(new ArrayComponentStore<>(componentManager.getType(componentType))));
+            if (!Modifier.isAbstract(componentType.getModifiers())) {
```
NicholasBatesNZ, line 52: A comment on why this filter is here would be nice too

PR #548, opened 2020-08-13
Projectiles affect entities

IsaacLic:
# Description
This adds working projectile collision with entities. When an entity is shot, this PR makes it so that the entity takes damage and absorbs the projectile properly.

# Testing
When the game starts, a golden asteroid should appear. Shooting it should destroy it.

To test the area-of-effect damage, run `groovyw module get makeshift -remote ThisIsPIRI` then start the game as the `Titanomachy`. Shooting the asteroid with the nuclear missile should destroy the asteroid.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/SolContactListener.java
+++ b/engine/src/main/java/org/destinationsol/game/SolContactListener.java
@@ -63,15 +63,15 @@ public class SolContactListener implements ContactListener {
 
         //TODO This is legacy code for handling contact with a Projectile, which currently is designed to work with SolObjects.
         // Once Projectile has been converted to be an entity, this should be refactored.
-        SolObject oA = (SolObject) dataA;
-        SolObject oB = (SolObject) dataB;
-        boolean aIsProj = oA instanceof Projectile;
-        if (!aIsProj && !(oB instanceof Projectile)) {
+        SolObject firstSolObject = (SolObject) dataA;
+        SolObject secondSolObject = (SolObject) dataB;
+        boolean firstSolObjectIsProjectile = firstSolObject instanceof Projectile;
+        if (!firstSolObjectIsProjectile && !(secondSolObject instanceof Projectile)) {
             return;
         }
-        Projectile proj = (Projectile) (aIsProj ? oA : oB);
-        SolObject o = aIsProj ? oB : oA;
-        proj.setObstacle(o, myGame);
+        Projectile projectile = (Projectile) (firstSolObjectIsProjectile ? firstSolObject : secondSolObject);
+        SolObject solObject = firstSolObjectIsProjectile ? secondSolObject : firstSolObject;
+        projectile.setObstacle(solObject, myGame);
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


engine/src/main/java/org/destinationsol/game/projectile/PointProjectileBody.java
```diff
@@ -107,15 +109,22 @@ private ProjectileRayBack(Projectile projectile, SolGame game) {
...
-                }
+
+            //TODO This is a patch to smooth over contact between an Entity and a Projectile. Once Projectile has been
+            // converted to be an Entity, this can be removed.
+            Object userData = fixture.getBody().getUserData();
+            if (userData instanceof EntityRef) {
+                userData = new SolObjectEntityWrapper((EntityRef) userData);
+            }
+
+            SolObject o = (SolObject) userData;
```
NicholasBatesNZ, line 120: If `userData` is still not a `SolObject` at this point we should break / return -1

engine/src/main/java/org/destinationsol/game/SolContactListener.java
```diff
@@ -50,24 +50,28 @@ public void beginContact(Contact contact) {
...
+        }
+        if (dataB instanceof EntityRef) {
+            dataB = new SolObjectEntityWrapper((EntityRef) dataB);
         }
 
-        //TODO handle contact between an entity and a SolObject
+        //TODO This is legacy code for handling contact with a Projectile, which currently is designed to work with SolObjects.
+        // Once Projectile has been converted to be an entity, this should be refactored.
+        SolObject oA = (SolObject) dataA;
+        SolObject oB = (SolObject) dataB;
```
NicholasBatesNZ, line 67: Again here, you removed an earlier check to make sure that `dataA` and `dataB` are both `SolObjects`. It was probably there for a good reason :)

engine/src/main/java/org/destinationsol/game/projectile/Projectile.java
```diff
@@ -115,25 +119,64 @@ private DSParticleEmitter buildEffect(SolGame game, EffectConfig ec, DrawableLev
...
-                collided(game);
-                if (config.emTime > 0 && obstacle instanceof SolShip) {
-                    ((SolShip) obstacle).disableControls(config.emTime, game);
+
+                if (config.density > 0) {
+                    obstacle = null;
+                    wasDamageDealt = true;
                 }
+
+                collided(game);
```
NicholasBatesNZ, line 149: I'm assuming you know what you're doing by not putting this in an `else` block :+1:

engine/src/main/java/org/destinationsol/game/projectile/Projectile.java
```diff
@@ -115,25 +119,64 @@ private DSParticleEmitter buildEffect(SolGame game, EffectConfig ec, DrawableLev
...
+
+                //TODO Once SolShip is an entity, this should be refactored to work with it
```
NicholasBatesNZ, line 151: Have you got a document somewhere listing all of the TODOs that should become TODONEs by the end of the project? It would be really helpful if you could share that

engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -551,6 +555,10 @@ public TutorialManager getTutMan() {
         return tutorialManager;
     }
 
+    public EntitySystemManager getEntitySystemManager() {
```
NicholasBatesNZ, line 558: What are the chances of injecting `Projectile` with the ESM so that we don't have to do this?

engine/src/main/java/org/destinationsol/game/projectile/PointProjectileBody.java
IsaacLic: `userData` will always be a `SolObject` by this point. The previous code assumed it would always be a SolObject, and my changes added the possibility of it being an EntityRef. I'll put in a comment that explains that

engine/src/main/java/org/destinationsol/game/SolContactListener.java
IsaacLic: I put that in [here](https://github.com/MovingBlocks/DestinationSol/pull/543/files#diff-36cb9a67d69750ed5e3e1e539965b7c1R56). Before I set up the contact handling with `Projectiles`, I needed to make sure that the game wouldn't crash. This removes the need for the check, because it now works properly.

engine/src/main/java/org/destinationsol/game/projectile/Projectile.java
IsaacLic: Good catch! I think the `else` got lost in the shuffle

PR #549, opened 2020-08-14
Impulse handling

IsaacLic:
# Description
This changes impulse handling from being asteroid-specific to being general purpose by adding a `mass` field in `BodyLinked` and by creating a `Durability` component. It also changes the `AsteroidImpulseHandler` to be a general `ImpulseHandlingSystem`.

# Testing
Start the game with a Loaded Imperial Large ship. The asteroid should be harder to destroy (because it now has a mass that is higher than one).

### DIFF
--- a/engine/src/main/java/org/destinationsol/body/components/BodyLinked.java
+++ b/engine/src/main/java/org/destinationsol/body/components/BodyLinked.java
@@ -37,8 +37,9 @@ public class BodyLinked implements Component<BodyLinked> {
         if (mass <= 0) {
             logger.error("Invalid value for the mass. It can't be less than or equal to zero.");
             this.mass = 1;
+        } else {
+            this.mass = mass;
         }
-        this.mass = mass;
     }
 
     public float getMass() {
--- a/engine/src/main/java/org/destinationsol/body/components/BodyLinked.java
+++ b/engine/src/main/java/org/destinationsol/body/components/BodyLinked.java
@@ -16,10 +16,30 @@
 package org.destinationsol.body.components;
 
 import com.badlogic.gdx.physics.box2d.Body;
-import org.terasology.gestalt.entitysystem.component.EmptyComponent;
+import org.destinationsol.body.systems.BodyHandlerSystem;
+import org.terasology.gestalt.entitysystem.component.Component;
+
 
 /**
- * Indicates that there is a {@link Body} associated with the entity.
+ * Indicates that there is a {@link Body} associated with the entity. It also contains the mass of the entity.
  */
-public class BodyLinked extends EmptyComponent<BodyLinked> {
+public class BodyLinked implements Component<BodyLinked> {
+
+    private float mass;
+
+    /**
+     * Sets the mass of the entity. This is called every tick by the {@link BodyHandlerSystem}.
+     */
+    public void setMass(float mass) {
+        this.mass = mass;
+    }
+
+    public float getMass() {
+        return mass;
+    }
+
+    @Override
+    public void copy(BodyLinked other) {
+        this.mass = other.getMass();
+    }
 }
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/destinationsol/force/systems/ImpulseHandlingSystem.java
```diff
@@ -13,37 +13,40 @@
...
 
-    @ReceiveEvent(components = {AsteroidMesh.class, Health.class, BodyLinked.class})
-    public EventResult onImpulse(ImpulseEvent event, EntityRef entity){
+    @ReceiveEvent(components = {Health.class, BodyLinked.class})
+    public EventResult onImpulse(ImpulseEvent event, EntityRef entity) {
 
-        //TODO get the mass from the body
-        float mass = 1;
+        float mass = entity.getComponent(BodyLinked.class).get().getMass();
+        float damage = event.getMagnitude() / mass;
```
NicholasBatesNZ, line 43: Division by zero if mass is zero. Might want to put a constraint + log in the component

engine/src/main/java/org/destinationsol/force/systems/ImpulseHandlingSystem.java
```diff
@@ -13,37 +13,40 @@
...
 
-        float damage = event.getMagnitude() / mass / DURABILITY;
-        entitySystemManager.sendEvent(new DamageEvent((int) damage), entity);
+        if (entity.hasComponent(Durability.class)) {
+            entity.getComponent(Durability.class).get().getDurability();
```
NicholasBatesNZ, line 46: Maybe I'm too tired but you're not actually doing anything with this are you?

IsaacLic: Fixed :)

engine/src/main/java/org/destinationsol/body/components/BodyLinked.java
```diff
@@ -16,10 +16,37 @@
...
+    /**
+     * Sets the mass of the entity. This is called every tick by the {@link BodyHandlerSystem}.
+     */
+    public void setMass(float mass) {
+        if (mass <= 0) {
+            logger.error("Invalid value for the mass. It can't be less than or equal to zero.");
+            this.mass = 1;
+        }
+        this.mass = mass;
+    }
```
NicholasBatesNZ, line 42: You set `this.mass` to 1 if the parameter is a bad value and then proceed to overwrite it with the bad value :wink:

engine/src/main/java/org/destinationsol/force/components/Durability.java
```diff
@@ -0,0 +1,54 @@
...
+     * This sets the durability of an entity. If the value passed in is less than or equal to zero, the durability
+     * defaults to be one.
+     */
+    public void setDurability(float durability) {
+        if (durability <= 0) {
+            logger.error("Invalid value for the durability. It can't be less than or equal to zero.");
+            this.durability = 1;
+        }
+        this.durability = durability;
+    }
```
NicholasBatesNZ, line 48: Same thing here with setting `this.durability` to a potentially bad value

engine/src/main/java/org/destinationsol/force/systems/ImpulseHandlingSystem.java
```diff
@@ -13,37 +13,41 @@
...
+        float damage = event.getMagnitude() / mass;
 
-        float damage = event.getMagnitude() / mass / DURABILITY;
-        entitySystemManager.sendEvent(new DamageEvent((int) damage), entity);
+        if (entity.hasComponent(Durability.class)) {
+            float durability = entity.getComponent(Durability.class).get().getDurability();
+            damage /= durability;
+        }
 
+        entitySystemManager.sendEvent(new DamageEvent((int) damage), entity);
```
NicholasBatesNZ, line 50: Huh? I thought `DamageEvent` took a float :thinking:

NicholasBatesNZ (approved):

PR #550, opened 2020-08-16
Rubble creation

IsaacLic:
# Description
This PR adds rubble creation when an entity with a `CreatesRubbleOnDestruction` component is destroyed. It also refactors `Shard` to be called `Rubble`, and `DropsMoneyOnDeath` to be called `DropsMoneyOnDestruction`.

# Testing
Start the game and shoot the golden asteroid. When it is destroyed, it should explode into rubble.

### DIFF
--- a/engine/src/main/java/org/destinationsol/rubble/systems/RubbleCreationSystem.java
+++ b/engine/src/main/java/org/destinationsol/rubble/systems/RubbleCreationSystem.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2020 The Terasology Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


engine/src/main/java/org/destinationsol/rubble/systems/RubbleCreationSystem.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * When an entity with a {@link CreatesRubbleOnDestruction} component is destroyed, this creates {@link Rubble}s where
+     * the entity was, unless the entity is in {@link Stasis}.
+     */
+    @ReceiveEvent(components = {CreatesRubbleOnDestruction.class, Position.class, Size.class})
+    @Before(DestructionSystem.class)
+    public EventResult onDeletion(DeletionEvent event, EntityRef entity) {
+        if (!entity.hasComponent(Stasis.class)) {
+            Vector2 position = entity.getComponent(Position.class).get().position;
+            Vector2 velocity;
```
NicholasBatesNZ, line 59: If you initially set the velocity to a `new Vector2` you can get rid of the else block.

Something else that I haven't remembered until now is that you really want to be using `SolMath.getVec()` to initialise a vector. It does fancy optimisations with pools. Then use `SolMath.free(vector)` when you're done with it.

IsaacLic: Good point! I'll change all the vectors in a separate PR.

NicholasBatesNZ (approved):

PR #551, opened 2020-08-20
improvement: textures scale to size

IsaacLic:
# Description
This fixes the graphics scaling and aligns the sprite with the collision mesh. As of now, it is slightly inaccurate.

# Testing
Start the game with a small ship and poke the edges gently to see if the asteroid's meshes are aligned with the sprite. Repeat at a few different sizes.

# Outstanding Work
I'm not sure that the multi-sprite renderables work (e.g. a ship with a gun). That needs to be tested and possibly fixed. The collision mesh alignment is also inaccurate, so that needs to be fixed. See issue #552


NicholasBatesNZ (approved):
Approving with #552

PR #570, opened 2020-12-04
Update drome.json

Esnardo:
# Description
> Buffs drome health by 2.5x. These things should now be able to survive the occasional hunter visit, as well as the usual heavy bombers and heavy impacts

# Testing
> Go to the nearest drome and start shooting.

# Notes
I might have overdone it a bit.

It's literally 1 character of change

### DIFF
--- a/modules/core/assets/ships/drome/drome.json
+++ b/modules/core/assets/ships/drome/drome.json
@@ -1,6 +1,6 @@
 {
     "size": 2.5,
-    "maxLife": 200,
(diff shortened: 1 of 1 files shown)


Cervator (approved):
Tests out and seems reasonable 👍 

Before the change taking out a ground station with a fully loaded imperial heavy was trivial. After the change it still isn't too hard (2.5-3 clips) if not under heavy fire, but a fair bit trickier on a desert planet with stronger enemies.

PR #579, opened 2020-12-13
Finalized Wave Gun Fix

Graviton48:
Wave Gun was still acting weird after previous commit, changed physSize value to make it work correctly.

# Description
> now Wave Gun works beautifully, no weirdness

# Testing
> obtain wave gun and shoot while on planet

### DIFF
--- a/modules/core/assets/configs/projectilesConfig.json
+++ b/modules/core/assets/configs/projectilesConfig.json
@@ -321,7 +321,7 @@
     "wave": {
         "dmg": 13,
         "dmgType": "energy",
-        "physSize": 0.6,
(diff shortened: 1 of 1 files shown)


Cervator (approved):
Tested out fine, can confirm wave gun seems to work normally 👍 

Perhaps oddly though I noticed that it will auto-hit if fired from inside a drome's shield - machine gun bullets will fly out through it fine, unless of course you hit the building. Wave gun just hits the shield immediately. Odd. Probably a pre-existing consequence of faction changes from a while ago, so not worried too much about it. Merging !

Graviton48:
cool :)

PR #583, opened 2021-02-08
Fix tests and migrate tests to Junit5

DarkWeird:
# Description

* Fix tests, which using Assets(interoduced special initializer for tests)  and Events
* Fix tests, which using Box2D (introduced special initializer for tests)
* Migrate tests to Junit5 for future.

# Testing

1. Check checks of this PR.
2. ???
3. Profit!

### DIFF
--- a/engine/src/test/java/org/destinationsol/testsupport/AssetsHelperInitializer.java
+++ b/engine/src/test/java/org/destinationsol/testsupport/AssetsHelperInitializer.java
@@ -1,0 +1,29 @@
+package org.destinationsol.testsupport;
+
+import com.badlogic.gdx.backends.headless.HeadlessApplication;
+import org.destinationsol.assets.AssetHelper;
+import org.destinationsol.assets.Assets;
+import org.destinationsol.modules.ModuleManager;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.parallel.ResourceLock;
+import org.terasology.gestalt.entitysystem.component.management.ComponentManager;
+
+/**
+ * Use it for tests, which used {@link org.destinationsol.assets.Assets}
+ */
+@ResourceLock("AssetsHelper")
+public interface AssetsHelperInitializer {
+
+    @BeforeAll
+    static void initAssets() throws Exception {
+        if (Assets.getAssetHelper() == null) {
+            new HeadlessApplication(null);
+            AssetHelper helper = new AssetHelper();
+            ModuleManager moduleManager = new ModuleManager();
+            moduleManager.init();
+            helper.init(moduleManager.getEnvironment(), new ComponentManager(), false);
+            Assets.initialize(helper);
+        }
+    }
+
+}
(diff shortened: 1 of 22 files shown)


engine/src/test/java/org/destinationsol/testsupport/AssetsHelperInitializer.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.jupiter.api.parallel.ResourceLock;
+import org.terasology.gestalt.entitysystem.component.management.ComponentManager;
+
+/**
+ * Use it for tests, which used {@link org.destinationsol.assets.Assets}
+ */
+@ResourceLock("AssetsHelper")
+public interface AssetsHelperInitializer {
+    ThreadLocal<TestGameStateObject> state = new ThreadLocal<>();
+
```
IsaacLic, line 20: Some Javadoc here would be helpful for figuring out what exactly this does. From what I can tell, this sets up certain assets for a test, but I'm not clear exactly what. Something like: "Initializes x, y, and z, so that cases such as a, b, and c have access to those resources without building a full game object."

engine/src/test/java/org/destinationsol/testsupport/TestGameStateObject.java
```diff
@@ -0,0 +1,38 @@
+package org.destinationsol.testsupport;
+
+import com.badlogic.gdx.Application;
+import org.destinationsol.modules.ModuleManager;
+import org.terasology.gestalt.entitysystem.component.management.ComponentManager;
+
```
IsaacLic, line 6: I'm not clear exactly what this class does. I can tell it's some sort of data structure for the game state, but I don't know exactly what it does or what the fields represent. A little bit of Javadoc would go a long way ;)

engine/src/test/java/org/destinationsol/testsupport/AssetsHelperInitializer.java
```diff
@@ -0,0 +1,57 @@
...
+    @BeforeEach
+    default void initAssets() throws Exception {
+        TestGameStateObject stateObject = new TestGameStateObject();
+        state.set(stateObject);
+        stateObject.setApplication(new HeadlessApplication(new ApplicationAdapter() {
```
IsaacLic, line 25: This might be a personal style thing, but I find nested constructors like this hard to read. I think it'd be clearer if the `HeadlessApplication` was pulled out and given a descriptive variable name that indicates what it does. Something along the lines of:

```
Application mockApplication = new HeadlessApplication(new ApplicationAdapter() {
      ...
});
stateObject.setApplication(mockApplication);
```

I find that that makes it clearer what the object being passed in represents.

IsaacLic (commented):
This looks great! I have a few minor clarity/documentation comments, but other than that, I think this can be merged.

engine/src/test/java/org/destinationsol/testsupport/AssetsHelperInitializer.java
DarkWeird: It is interfaces. interface can store only `public` `static` fields.
Then we should to use `ThreadLocal` + Before/After to store field only for  this Test.
It is not specific usage nor my created idea just now.
It is like standard for interfaces with defaults in junit5.

Give me time after this PR. then I will create Junit5's Extension for tests. ok?

IsaacLic (approved):

Cervator (approved):
PR looks fine, game works, tests now _also_ work, hooray! Merging, thanks :-)

PR #586, opened 2021-02-28
feat: implementation of gestalt-DI

pollend:
here is some experimental work with: https://github.com/MovingBlocks/gestalt/pull/87\

This is an implementation of gestalt DI. At the moment there is a ApplicationContext --> GameContext. Application context will contains all top level dependencies such as the audio system and the game context is the actual gameplay systems. This could be managed and split into different layers depending on different states of the application. each module has its own context so

This is how the current import scheme works with Gestalt modules so each module is only dependent on a child module and so forth. there is no way that moduel2 has a cyclic dependency on Module1

```
Application --> GameContext --> Module1 --> Module2 --> ... --> ModuleN  
```

This is a straightforward replacement of the context system that was pulled in from Terasology. Each context defines the services that are needed and at request the object is built up from the dependencies within that context and the parent context if it fails to get resolved in the current context. anything that used to be dependent on pulling from the context will declare the dependencies that are needed to build that object and the context will resolve building that object.

Unit tests are a matter of providing a serviceRegistry with the needed dependencies and the injection system will resolve those as need be. it should be easy enough to construct a mock object within the context and the system should handle the rest. reduces the burden of working backwards from the source to construct a test environment.   

This handles a lot of the clear problems with the old scheme.

- dependencies are added in at some arbitrary point so its hard to determine what is and isn't available
- constructing a test environment is very reliant on understanding the couplings within the system   so its necessary to work out dependencies to construct a working environment.

A lot of the code that is used to construct the game is gone and and has been handed over the DI system.

This also does away with reflections. this is also managed by gestalt-DI. I think @DarkWeird would have more information about that portion of the implementation. basically lets us do some scanning for a given type without paying the cost of walking the class path. accomplishes a similar goal to reflections bug handled through the annotation processor.


IsaacLic (commented):
Looks good as far as I can tell, although I don't really understand how it works XD

I have two main comments:
* In my opinion, `@Introspected` doesn't communicate what what exactly the annotation does. It seems to me like something that would make more sense if I understood how the DI system works. Presumably, the annotation should be used by people who don't know all of the nuances of how the system works, right? For example, if GSOC participants are creating classes that do whatever type of thing `@Introspected` is for, I think there's a decent chance that they'll be confused about the purpose of the annotation.
* As I mentioned in Discord, I think `@Inject` being used for two different meanings (injected fields and constructors used by the DI system) makes it harder for people to understand and use it. In my opinion, one of the two usages should have a different name, unless there's some technical reason why they need to have the same annotation.

That being said, this system looks really cool, and I'm excited for it to be put into place :)

pollend:
> Looks good as far as I can tell, although I don't really understand how it works XD
> 
> I have two main comments:
> 
>     * In my opinion, `@Introspected` doesn't communicate what what exactly the annotation does. It seems to me like something that would make more sense if I understood how the DI system works. Presumably, the annotation should be used by people who don't know all of the nuances of how the system works, right? For example, if GSOC participants are creating classes that do whatever type of thing `@Introspected` is for, I think there's a decent chance that they'll be confused about the purpose of the annotation.
> 
>     * As I mentioned in Discord, I think `@Inject` being used for two different meanings (injected fields and constructors used by the DI system) makes it harder for people to understand and use it. In my opinion, one of the two usages should have a different name, unless there's some technical reason why they need to have the same annotation.
> 
> 
> That being said, this system looks really cool, and I'm excited for it to be put into place :)

I realized introspected is not necessary If I can just scan for the presence of @Inject. 

https://github.com/MovingBlocks/gestalt/pull/87/commits/c5dad5c1be2817d8bc5dbb64bb86cc1e1e1a1686

pollend:
for SolCam I would like to use a qualifier to drive the injection.

```
@MainCamera
@Inject
SolCam cam;
```

pollend:
this seems to build umm there is some problems with NUI but this seems to be a good start since we now have a published snapshot. @BenjaminAmos @DarkWeird .

BenjaminAmos:
The primary NUI issue is due to the widget classes being obtained previously via gestalt's reflection in `AssetHelper.java`.
https://github.com/MovingBlocks/DestinationSol/blob/ddeb0dd88a46bd2205a0e39cc4146889bb9d39fd/engine/src/main/java/org/destinationsol/assets/AssetHelper.java#L77-L83
MovingBlocks/TeraNUI#44 should help with this (changing from a dependency here on `nui-gestalt7` to one on `nui-gestalt-di`) but will not solve all the NUI issues. Since the NUI classes are not indexed via-annotation in the NUI pull request, sub-classes of those NUI classes found in Destination Sol (and/or its modules) are not found either.

pollend:
@BenjaminAmos when do we want to go forward with this?

skaldarnar:
> when do we want to go forward with this? I guess we need to publish a new nui package right?

I don't know. The linked NUI PR is not merged, and I don't know whether this has any impact on NUI in Terasology. Feel free to go head with this in DestSol land if it does not disrupt Terasology.

pollend:
@BenjaminAmos and @DarkWeird so @Cervator created a snapshot but i'm not sure what needs to be changed. when I attempt to use WidgetLibrary I get this exception is this something to do with packaging?

```
WARN [main] (Reflections.java:181) - given scan urls are empty. set urls in the configuration
ERROR [main] (SolDesktop.java:170) - This exception was not caught:
com.badlogic.gdx.utils.GdxRuntimeException: java.lang.NoSuchMethodError: 'org.reflections.Reflections org.terasology.gestalt.module.Module.getModuleManifest()'
        at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:117)
        at org.destinationsol.desktop.SolDesktop.main(SolDesktop.java:129)
Caused by: java.lang.NoSuchMethodError: 'org.reflections.Reflections org.terasology.gestalt.module.Module.getModuleManifest()'
        at org.terasology.reflection.ModuleTypeRegistry.initializeReflections(ModuleTypeRegistry.java:42)
        at org.terasology.reflection.ModuleTypeRegistry.reload(ModuleTypeRegistry.java:35)
        at org.terasology.reflection.ModuleTypeRegistry.<init>(ModuleTypeRegistry.java:29)
        at org.destinationsol.assets.AssetHelper.init(AssetHelper.java:79)
        at org.destinationsol.SolApplication.create(SolApplication.java:166)
        at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.initializeListener(Lwjgl3Window.java:422)
        at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.update(Lwjgl3Window.java:370)
        at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:137)
        at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:111)

```

Cervator:
FYI to get an update in here as well - see https://github.com/MovingBlocks/TeraNUI/pull/44#issuecomment-846494330 for details, in short it has been built locally and released as a snapshot that @pollend got to work with this PR.

Assuming we can move forward with both NUI and Gestalt then yeah this could continue - but I do kinda believe we should finalize those two as releases first. Not great to continue here while depending on a stack of lib snapshots. This is still an experiment - I want to see it realized, but need to balance that with the health of DS overall 👍

pollend:
Its now possible to run gestalt DI with destination sol with the added changes to TeraNUI. There is  now a 4.x snapshot branch that is getting pulled in to run this.  this drops the current context in favor of this gestalt depdency injection scheme. going forward we would want to better organize how the game is initialized and separate the configuration into separate service registries. It would make sense to move things into separate subsystems.

It should be enough to check out this branch and start the game. I have a temporary wrapper around the context that will need to dropped at some point.

engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -106,73 +140,17 @@
...
-        asteroidBuilder = new AsteroidBuilder();
-        lootBuilder = new LootBuilder();
-        mapDrawer = new MapDrawer();
-        rubbleBuilder = new RubbleBuilder();
-        context.put(RubbleBuilder.class, rubbleBuilder);
-        galaxyFiller = new GalaxyFiller(hullConfigManager);
-        starPortBuilder = new StarPort.Builder();
-        drawableDebugger = new DrawableDebugger();
-        mountDetectDrawer = new MountDetectDrawer();
-        beaconHandler = new BeaconHandler();
```
pollend, line 175: all this has been moved into a single ServiceRegistry. might make sense to split this into different registries for testable. i.e constructing a world generator.

engine/src/main/java/org/destinationsol/ContextWrapper.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public <T, U extends T> void put(Class<T> type, U object) {
+        throw new RuntimeException("Unsupported");
+    }
+
+    @Override
+    public <T, U extends T> void remove(Class<T> type, U object) {
+        throw new RuntimeException("Unsupported");
+    }
+}
```
pollend, line 55: some code is dependent on this so I have a wrapper around the old context but probably should be removed at a later point when the code is organized.

templates/build.gradle
```diff
@@ -165,6 +162,17 @@ sourceSets {
     test.java.outputDir = new File("$buildDir/testClasses")
 }
 
+// Adds Resources as parameter for AnnotationProcessor (gather ResourceIndex,
+// also add resource as input for compilejava, for re-gathering ResourceIndex, when resource was changed.
+compileJava { // TODO adds assets
```
BenjaminAmos, line 167: This should already be including the assets anyway. See the `syncAssets` task below.

engine/src/main/java/org/destinationsol/game/ObjectManager.java
```diff
@@ -356,7 +360,8 @@ public void addFarObjNow(FarObject fo) {
         return myFarPorts;
     }
 
-    public void dispose() {
+    @Override
+    public void close() throws Exception {
         myWorld.dispose();
```
BenjaminAmos, line 365: This doesn't appear to be called anywhere. Might this cause a memory leak?

engine/build.gradle
```diff
@@ -2,16 +2,11 @@ apply from: '../config/gradle/common.gradle'
...
-import org.reflections.scanners.SubTypesScanner;
-import org.reflections.scanners.TypeAnnotationsScanner;
-import org.reflections.util.ConfigurationBuilder;
-import org.reflections.util.ClasspathHelper;
-
 // Dependencies needed for what our Gradle scripts themselves use. It cannot be included via an external Gradle file :-(
 buildscript {
     repositories {
         mavenCentral()
+        mavenLocal()
```
BenjaminAmos, line 9: Whilst this is very useful for local development, it often causes problems too. I've had issues with unpublished older snapshots of a library in the `mavenLocal` repository superceeding published newer ones before, so I wouldn't recommend including it long-term.

engine/src/main/java/org/destinationsol/game/ObjectManager.java
pollend: Autoclosable is automatically called when the context is freed

Cervator:
Ran this through a quick test on Android - got adjustments made for a Gestalt ES change elsewhere, and I manually published the Android component from Gestalt at snapshot v8 as that was missing. With that in place the game works! Nicely done.

Have not looked at the code but it tests out. I think we should aim to merge #607 and #605 first, but then really about everything should get merged, and it seems like everything should about work. Really may just be taking the time to go through to test and review, then merge, and resolve conflicts. Hopefully this won't get too tricky with other stuff merging

PR #592, opened 2021-03-27
Updated NUI to version 3.0.0-SNAPSHOT

BenjaminAmos:
# Description
This pull request updates the version of NUI used to `3.0.0-SNAPSHOT`. It brings the project back up-to-date with the most recent NUI bugfixes and feature additions. This is the first release of NUI that can operate independently of gestalt (MovingBlocks/TeraNUI#47).

# Testing
The testing for this is mostly the same as that for the initial NUI integration (see #536), apart from that you should not need to clone and locally publish a new version of NUI.

# Notes
- Keyboard input and character input are now separate things in NUI.
- `UISkin` is no longer an asset type and the asset equivalent is `UISkinAsset`. A `UISkinAsset` is not an instance of `UISkin`, however it provides access to one.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,6 +27,7 @@ allprojects {
         appName = 'DestinationSol'
         gdxVersion = '1.9.8'
         roboVMVersion = '2.3.3'
+        nuiVersion = '3.0.0-SNAPSHOT'
     }
 }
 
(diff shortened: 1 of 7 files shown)


Cervator (approved):
Tests out! Went through all the original NUI test cases, including Android. As noted just about everything seemed fine, with occasionally quirks or pre-existing issues (like checkboxes on the mig test screen being _really_ hard to spot)

Thanks again!

PR #593, opened 2021-04-02
Refactored Rubble to use ECS

IsaiahBlanks:
This pull request refactors rubble to use the Entity Component System architecture. Currently, only rubble spawned from the test  asteroid at spawn will be created using ECS, as that asteroid is the only entity with the CreatesRubbleOnDestruction component. 

**Outstanding Issues**
There are some issues concerning the physics of the rubble, as it seems there is not a way to add angular velocity with the velocity component yet. The textures also appears to be offset slightly from the collision mesh.

### DIFF
--- a/engine/src/main/java/org/destinationsol/rubble/systems/RubbleCreationSystem.java
+++ b/engine/src/main/java/org/destinationsol/rubble/systems/RubbleCreationSystem.java
@@ -75,8 +75,6 @@ public class RubbleCreationSystem implements EventReceiver {
         if (!entity.hasComponent(Stasis.class)) {
             Velocity velocityComponent;
             Angle angleComponent;
-            Size sizeComponent;
-            Position positionComponent;
 
             if (entity.hasComponent(Velocity.class)) {
                 velocityComponent = entity.getComponent(Velocity.class).get();
@@ -85,7 +83,6 @@ public class RubbleCreationSystem implements EventReceiver {
                 velocityComponent.velocity = new Vector2();
             }
 
-
             if (entity.hasComponent(Angle.class)) {
                 angleComponent = entity.getComponent(Angle.class).get();
             } else {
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/destinationsol/rubble/systems/RubbleCreationSystem.java
```diff
@@ -39,13 +51,76 @@
...
+
+            //Create size component
+            Size sizeComponent = new Size();
+            sizeComponent.size = scale;
+
+            //Create velocity component
+            Velocity velocityComponent = new Velocity();
+            Vector2 velocity = SolMath.fromAl(velocityAngle, SolRandom.randomFloat(MAX_SPD));
+            velocity.add(vel.velocity);
+            SolMath.free(velocity);
```
IsaacLic, line 116: By freeing the velocity before you set it in the component, you're telling the game that it can re-use the vector, which can lead to unspecified behavior. This line should probably be moved to after the entity is created.

IsaacLic (changes requested):
Looks good! Most of my comments are small naming/documentation things, other than the one issue about freeing the Vector too early. 

I have not been able to test it (still don't have my workspace set up well), so I haven't been able to see the physics issues. One that I'm noticing (that could be a relevant factor) is that there isn't a good way to set the Velocity, Angle, AngularVelocity, or Position after the entity has already been created. Fixing that is outside of the scope of this PR, of course, but your comment about AngularVelocity made me realize the issue. I'll fix that at some point.

engine/src/main/java/org/destinationsol/rubble/systems/RubbleCreationSystem.java
```diff
@@ -39,32 +51,122 @@
...
+
+            if (entity.hasComponent(Angle.class)) {
+                angleComponent = entity.getComponent(Angle.class).get();
+            } else {
+                angleComponent = new Angle();
+                angleComponent.setAngle(0f);
+            }
+
+
+            if (entity.hasComponent(Size.class)) {
```
IsaacLic, line 97: Since `@ReceiveEvent` requires that the entity have a Size component, the `if` statement is unnecessary. Same with the Position

IsaacLic (approved):
This looks good! @NicholasBatesNZ can you review/merge this?

Cervator:
[Finish out Rubble PR](https://trello.com/c/HzbEZbd2/11-finish-out-rubble-pr)

Cervator:
Tested this out real quick, after a trivial conflict fix. Can confirm rubble comes out of the golden asteroid, but ... the rubble is invincible? 😅 I can tell I'm hitting it, physics shape seems alright but since so small it is hard to be sure.

Also, should the golden asteroid break into incrementally smaller pieces, rather than go to the tiniest size immediately, or is that outside the scope of this PR? Since that asteroid is a test piece anyway.

Cervator:
Oh, and the invincible rubble seems to be pre-existing specific to the stuff the golden asteroid pops out.

Conflict resolution was just accepting all from this PR then adjusting three "shard" to "rubble"

Cervator:
Merged this after resolving the conflict locally. Submitted #617 for follow-up fixes. Since it is a test asteroid it doesn't exactly have to be perfect - yet 👍

PR #595, opened 2021-05-30
Generified collision mesh generation

IsaacLic:
# Description
This PR generifies collision mesh creation. Rather than needing to create a separate system for each type of `Body` with a new instance of `CollisionMeshLoader`, the new `CollisionMeshBuilderSystem` handles all `Body` creation. 

The `BodyLinked` component has two new fields - one for the JSON path to the mesh data, and another for the schema file that details how the data is laid out. When an entity has a `BodyLinked` component, the system will build a `Body` using the mesh data when it receives a `GenerateBodyEvent` (which will happen at the next tick).

# Testing
Shoot the testing asteroid (the golden asteroid visible on startup). The projectile should still collide with it, so the asteroid should be destroyed.

# Outstanding Work
This PR will cause merge conflicts with #593. I'm marking this as a draft until after that PR is merged, and then I'll address the conflicts.


Cervator:
[Destroy the Golden Asteroid, once and for all!](https://trello.com/c/xdD46zp2/12-destroy-golden-asteroid-once-and-for-all)

Cervator:
#593 has been merged, ready for conflict resolution and finalizing 👍 

Does this likely clash any with the updated mesh work for mazes in the GSOC project? I guess that might relate more to just arranging meshes rather than generating them. Mazes are still pretty whacky on collision shapes but I think that's pre-existing.

PR #596, opened 2021-06-07
Skeleton classes for world-gen project

IsaiahBlanks:
# Description
This PR contains classes that represent the skeleton of the world-gen project

# Testing
There should be a couple of print statements showing that the Generators are called. The Generator Implementation classes should have their build() method called

# Outstanding Work
Several aspects of this PR may need to be changed in the future, including the data structure used to hold the Generators and the particular way those Generators are instantiated

### DIFF
--- a/engine/src/main/java/org/destinationsol/world/WorldBuilder.java
+++ b/engine/src/main/java/org/destinationsol/world/WorldBuilder.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2020 The Terasology Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world;
+
+
+import org.destinationsol.game.context.Context;
+import org.destinationsol.modules.ModuleManager;
+import org.destinationsol.world.generators.MazeGenerator;
+import org.destinationsol.world.generators.PlanetGenerator;
+import org.destinationsol.world.generators.SolSystemGenerator;
+
+import java.util.ArrayList;
+
+/**
+ * This class is the starting point for world generation
+ */
+public class WorldBuilder {
+    ArrayList<Object> solSystemGenerators = new ArrayList<>();
+    ArrayList<Object> planetGenerators = new ArrayList<>();
+    ArrayList<Object> mazeGenerators = new ArrayList<>();
+    Context context;
+
+    public WorldBuilder(Context context) {
+        this.context = context;
+        populateGeneratorList();
+
+        initializeSolSystemGenerators();
+    }
+
+    /**
+     * This method will use reflection to get all generator classes (using the @WorldGenerator annotation)
+     */
+    private void populateGeneratorList() {
+
+        for (Class generator : context.get(ModuleManager.class).getEnvironment().getSubtypesOf(SolSystemGenerator.class)) {
+            if (generator.isAnnotationPresent(WorldGenerator.class)) {
+                try {
+                    Object solSystemGenerator = generator.newInstance();
+                    solSystemGenerators.add(solSystemGenerator);
+                } catch (InstantiationException | IllegalAccessException e) {
+                    e.printStackTrace();
+                }
+            }
+        }
+
+        for (Class generator : context.get(ModuleManager.class).getEnvironment().getSubtypesOf(PlanetGenerator.class)) {
+            if (generator.isAnnotationPresent(WorldGenerator.class)) {
+                try {
+                    Object planetGenerator = generator.newInstance();
+                    planetGenerators.add(planetGenerator);
+                } catch (InstantiationException | IllegalAccessException e) {
+                    e.printStackTrace();
+                }
+            }
+        }
+
+        for (Class generator : context.get(ModuleManager.class).getEnvironment().getSubtypesOf(MazeGenerator.class)) {
+            if (generator.isAnnotationPresent(WorldGenerator.class)) {
+                try {
+                    Object mazeGenerator = generator.newInstance();
+                    mazeGenerators.add(mazeGenerator);
+                } catch (InstantiationException | IllegalAccessException e) {
+                    e.printStackTrace();
+                }
+            }
+        }
+    }
+
+    /**
+     * Eventually this method will be used to pick a random assortment of the available generators. The number of
+     * SolSystemGenerators used will depend on how many Sol Systems the world should have
+     */
+    private void initializeSolSystemGenerators() {
+        for (Object gen : solSystemGenerators) {
+            SolSystemGenerator generator = (SolSystemGenerator) gen;
+            generator.setPlanetGenerators(planetGenerators);
+            generator.setMazeGenerators(mazeGenerators);
+            generator.build();
+        }
+    }
+}
--- a/engine/src/main/java/org/destinationsol/world/WorldBuilder.java
+++ b/engine/src/main/java/org/destinationsol/world/WorldBuilder.java
@@ -19,44 +19,43 @@ import org.destinationsol.common.SolRandom;
 import org.destinationsol.game.context.Context;
 import org.destinationsol.modules.ModuleManager;
 import org.destinationsol.world.generators.FeatureGenerator;
-import org.destinationsol.world.generators.SolSystemGenerator;
+import org.destinationsol.world.generators.SolarSystemGenerator;
 
 import java.lang.reflect.Modifier;
 import java.util.*;
 
 /**
  * This class is the starting point for world generation. When a new world is created, this class first retrieves
- * all Generator classes and then initiates each SolSystem's build process. Two types of world generation classes are
- * retrieved: those that subclass SolSystemGenerator and those at subclass FeatureGenerator.
+ * all Generator classes and then initiates each SolarSystem's build process. Two types of world generation classes are
+ * retrieved: those that subclass SolarSystemGenerator and those at subclass FeatureGenerator.
  */
 public class WorldBuilder {
-    /**
-     * These arrays hold instances of any class which extends SolSystemGenerator or FeatureGenerator, respectively.
-     */
-    private ArrayList<SolSystemGenerator> solSystemGenerators = new ArrayList<>();
+
+    //These ArrayLists hold instances of any class which extends SolarSystemGenerator or FeatureGenerator, respectively
+    private ArrayList<SolarSystemGenerator> solarSystemGenerators = new ArrayList<>();
     private ArrayList<FeatureGenerator> featureGenerators = new ArrayList<>();
     private Context context;
     private final int numberOfSystems;
-    private int systemsBuilt; //This field is for testing purposes
+    private int systemsBuilt; //This field is for testing whether the correct number of SolarSystems are built
 
     public WorldBuilder(Context context, int numSystems) {
         this.context = context;
         numberOfSystems = numSystems;
-        populateSolSystemGeneratorList();
+        populateSolarSystemGeneratorList();
         populateFeatureGeneratorList();
         systemsBuilt = 0;
     }
 
     /**
-     * This method uses reflection to retrieve all SolSystemGenerator classes. They are added to the list
-     * of SolSystemGenerators.
+     * This method uses reflection to retrieve all SolarSystemGenerator classes. They are added to the list
+     * of SolarSystemGenerators.
      */
-    private void populateSolSystemGeneratorList() {
+    private void populateSolarSystemGeneratorList() {
 
-        for (Class generator : context.get(ModuleManager.class).getEnvironment().getSubtypesOf(SolSystemGenerator.class)) {
+        for (Class generator : context.get(ModuleManager.class).getEnvironment().getSubtypesOf(SolarSystemGenerator.class)) {
             try {
-                SolSystemGenerator solSystemGenerator = (SolSystemGenerator) generator.newInstance();
-                solSystemGenerators.add(solSystemGenerator);
+                SolarSystemGenerator solarSystemGenerator = (SolarSystemGenerator) generator.newInstance();
+                solarSystemGenerators.add(solarSystemGenerator);
             } catch (InstantiationException | IllegalAccessException e) {
                 e.printStackTrace();
             }
@@ -81,32 +80,32 @@ public class WorldBuilder {
     }
 
     /**
-     * This method initializes the SolSystemGenerators. How many generators are initialized depends on the number
-     * of SolSystems the world is set to have. When there are multiple types of SolSystemGenerators available, this
+     * This method initializes the SolarSystemGenerators. How many generators are initialized depends on the number
+     * of SolarSystems the world is set to have. When there are multiple types of SolarSystemGenerators available, this
      * method chooses randomly from the list of all generators to decide which to create.
      *
-     * This method calls setFeatureGenerators() on each SolSystemGenerator to give each SolSystem access to other
-     * FeatureGenerators (Planets, Mazes, etc.) so they can be initialized and incorporated into the SolSystem.
+     * This method calls setFeatureGenerators() on each SolarSystemGenerator to give each SolarSystem access to other
+     * FeatureGenerators (Planets, Mazes, etc.) so they can be initialized and incorporated into the SolarSystem.
      */
-    private void initializeSolSystemGenerators() {
+    private void initializeRandomSolarSystemGenerators() {
         for (int i = 0; i < numberOfSystems; i++) {
-            SolSystemGenerator solSystemGenerator = solSystemGenerators.get(SolRandom.seededRandomInt(solSystemGenerators.size()));
-            solSystemGenerator.setFeatureGenerators(featureGenerators);
-            solSystemGenerator.build();
+            SolarSystemGenerator solarSystemGenerator = solarSystemGenerators.get(SolRandom.seededRandomInt(solarSystemGenerators.size()));
+            solarSystemGenerator.setFeatureGenerators(featureGenerators);
+            solarSystemGenerator.build();
             systemsBuilt++;
         }
     }
 
-    public void build() {
-        initializeSolSystemGenerators();
+    public void buildWithRandomSolarSystemGenerators() {
+        initializeRandomSolarSystemGenerators();
     }
 
     public int getSystemsBuilt() {
         return systemsBuilt;
     }
 
-    public ArrayList<SolSystemGenerator> getSolSystemGenerators() {
-        return solSystemGenerators;
+    public ArrayList<SolarSystemGenerator> getSolarSystemGenerators() {
+        return solarSystemGenerators;
     }
 
     public ArrayList<FeatureGenerator> getFeatureGenerators() {
--- a/engine/src/main/java/org/destinationsol/world/WorldBuilder.java
+++ b/engine/src/main/java/org/destinationsol/world/WorldBuilder.java
@@ -22,7 +22,7 @@ import org.destinationsol.world.generators.FeatureGenerator;
 import org.destinationsol.world.generators.SolarSystemGenerator;
 
 import java.lang.reflect.Modifier;
-import java.util.*;
+import java.util.ArrayList;
 
 /**
  * This class is the starting point for world generation. When a new world is created, this class first retrieves
--- a/engine/src/main/java/org/destinationsol/world/WorldBuilder.java
+++ b/engine/src/main/java/org/destinationsol/world/WorldBuilder.java
@@ -27,7 +27,7 @@ import java.util.ArrayList;
 /**
  * This class is the starting point for world generation. When a new world is created, this class first retrieves
  * all Generator classes and then initiates each SolarSystem's build process. Two types of world generation classes are
- * retrieved: those that subclass SolarSystemGenerator and those at subclass FeatureGenerator.
+ * retrieved: those that subclass SolarSystemGenerator and those that subclass FeatureGenerator.
  */
 public class WorldBuilder {
 
--- a/engine/src/main/java/org/destinationsol/world/WorldBuilder.java
+++ b/engine/src/main/java/org/destinationsol/world/WorldBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 The Terasology Foundation
+ * Copyright 2021 The Terasology Foundation
  *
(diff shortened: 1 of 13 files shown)


NicholasBatesNZ:
Congratulations on your first official PR for GSoC :tada: 
You're off to a good start but I can see some problems. The way you've written this doesn't seem to allow for the generic feature generator classes we talked about a few weeks ago, unless I'm missing something and you can enlighten me?
As I see it, the `WorldBuilder` should only be pulling in two types of classes. 1) classes that extend `SolSystemGenerator` and 2) classes that extend, let's say, a `FeatureGenerator`.
You should be thinking in terms of "how modular can I be?" Having the engine explicitly look for a `PlanetGenerator` is not particularly modular.
Ping me on Discord for more clarity, or to tell me why I'm wrong :)
What is the purpose of the `WorldGenerator` annotation? It seems a bit superfluous to me.

engine/src/main/java/org/destinationsol/world/generators/SolSystemGenerator.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+import java.util.ArrayList;
+
+public abstract class SolSystemGenerator {
+
+    //somehow this needs the get the right PlanetGenerators for the Specific SolSystemGenerator
```
IsaacLic, line 22: This should probably have a TODO in front of it, right?

IsaacLic (commented):
Most of my comments are minor. I have two main comments:
1) As @NicholasBatesNZ said, I'm not sure what purpose the `WorldGenerator` annotation is serving. A bit of Javadoc would go a long way, which brings me to my second point:
2) Your code could definitely benefit from some documentation. Imagine someone encountering your code two years down the line. A bit of documentation could make the difference between five minutes and five hours spent understanding what a class does. IIRC, the policy is to have class-level documentation for every class, and to use method-level documentation whenever necessary. I could be wrong about that, though.

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+public class WorldBuilder {
+    ArrayList<SolSystemGenerator> solSystemGenerators = new ArrayList<>();
+    ArrayList<FeatureGenerator> featureGenerators = new ArrayList<>();
+    Context context;
+    private final int numberOfSystems;
+
+
+    /**
+     * Initialize the WorldBuilder class
+     * @param context System Context
```
IsaacLic, line 41: This line is kinda unclear, especially given the use of the word "System" elsewhere used to mean a solar system. Something like "the game's context" might be better.

Tbh, I don't think the constructor necessarily needs Javadoc, unless you're talking about what actually happens when you initialize an instance of the class. Maybe @Cervator or @NicholasBatesNZ can chime in here.

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+     * method chooses randomly from the list of all generators to decide which to create.
+     */
+    private void initializeSolSystemGenerators() {
+        Random random = new Random();
+        for (int i = 0; i < numberOfSystems; i++) {
+            int systemIndex = random.nextInt(solSystemGenerators.size());
+            SolSystemGenerator solGenerator = solSystemGenerators.get(systemIndex);
+            solSystemGenerators.remove(systemIndex);
+            solGenerator.setFeatureGenerators(featureGenerators);
+            solGenerator.build();
```
IsaacLic, line 101: I think building the actual systems should be separate from the setup code. Even if it's called right after in `SolGame`, it still is conceptually separate, and moving it to a separate method call (`worldBuilder.build()`) allows for it to be created whenever makes the most sense.

engine/src/main/java/org/destinationsol/world/generators/FeatureGenerator.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+/**
+ * This class represents the framework
```
IsaacLic, line 19: Can you elaborate what framework? While it may be clear from context now, think about a person who is designing their own feature generator, without having followed the creation of this code. What will help that person understand how to use this code?

engine/src/main/java/org/destinationsol/world/generators/MazeGeneratorImpl.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+/**
+ * This class is a concrete implementation of a Maze and handles its creation
```
IsaacLic, line 19: "of a Maze generator" ?

engine/src/main/java/org/destinationsol/world/generators/MazeGenerator.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+/**
+ * This class defines the contract for Maze generators. Any Maze will be created from a concrete
```
IsaacLic, line 19: Can you elaborate on what the contract is?

engine/src/main/java/org/destinationsol/world/generators/PlanetGenerator.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+/**
+ * This class defines the contract for Planet generators. Any Planet will be created from a concrete
```
IsaacLic, line 19: See above

engine/src/main/java/org/destinationsol/world/generators/PlanetGeneratorImpl.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+/**
+ * This class is a concrete implementation of a Planet and handles its creation
```
IsaacLic, line 19: See above

engine/src/main/java/org/destinationsol/world/generators/SolSystemGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+import java.util.ArrayList;
+
+/**
+ * This class defines the contract for SolSystem generators. Any SolSystem in the game will be created from a concrete
```
IsaacLic, line 21: See above

engine/src/main/java/org/destinationsol/world/generators/SolSystemGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+ * implementation of this class. Every SolSystem is given access to all the available features. Particular
```
IsaacLic, line 22: What available features? An example might clarify things

engine/src/main/java/org/destinationsol/world/generators/SolSystemGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+    //This field is protected as that allows subclasses to access it
+    protected ArrayList<FeatureGenerator> featureGenerators = new ArrayList<>();
+
+    /**
+     * This method allows {@link org.destinationsol.world.WorldBuilder} to give SolSystems access to other
+     * world features so they can be initialized and incorporated into the SolSystem
+     * @param generators The list of FeatureGenerators available
+     */
+    public void setFeatureGenerators(ArrayList<FeatureGenerator> generators) {
+        for (FeatureGenerator feature : generators) {
```
IsaacLic, line 35: This can be simplified using `generators.clone()`. I wonder if this should make a deep copy, though. @NicholasBatesNZ @Cervator ?

engine/src/main/java/org/destinationsol/world/generators/SolSystemGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+            featureGenerators.add(feature);
+        }
+    }
+
+    /**
+     * This method is intended to first set up the SolSystem during world generation and then initialize all the
```
IsaacLic, line 41: I'm not sure I understand. Does this build the actual SolSystem? If it does, wouldn't the FeatureGenerators have been initialized beforehand?

engine/src/main/java/org/destinationsol/world/generators/SolSystemGeneratorImpl.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+/**
+ * This class is a concrete implementation of a SolSystem and handles its creation. This class has access to the
```
IsaacLic, line 19: See above

IsaacLic (commented):
Looking much better :)

I have a bunch of minor comments about improving the Javadoc. In short: think about what it'd be like if you were encountering this code for the first time, and had no idea what it did or how it worked. The Javadoc should provide an overview, so that the code can be easily understood.

engine/src/main/java/org/destinationsol/world/generators/SolSystemGenerator.java
IsaiahBlanks: @IsaacLic The FeatureGenerators are the next step in the world-gen process. Meaning first the SolSystem's own features are generated (belts, position in the world, deciding which Features to use), then the FeatureGenerators themselves are built (i.e. Planets, Mazes, etc. have their own build() function, and SolSystemGenerator will finish up by calling those). I think the chart in the proposal clarifies a little bit [here](https://docs.google.com/document/d/12KnOR3guxP6OKnrY9TX_29dOwVF7fi1Paa3bJNJS4qo/edit)

IsaacLic: That makes sense. How about "set up the broader SolSystem" and "use FeatureGenerators to do X" ? That might clarify the relationship between the two

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
Cervator: Yeah I'd leave off Javadoc on constructors that don't do anything particularly unusual. With a little luck the variables being set themselves have good enough Javadoc to not also need them here.

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * This method initializes the SolSystemGenerators. How many generators are initialized depends on the number
+     * of SolSystems the world is set to have. When there are multiple types of SolSystemGenerators available, this
+     * method chooses randomly from the list of all generators to decide which to create.
+     *
+     * This method calls setFeatureGenerators() on each SolSystemGenerator to give each SolSystem access to other
+     * FeatureGenerators (Planets, Mazes, etc.) so they can be initialized and incorporated into the SolSystem.
+     */
+    private void initializeSolSystemGenerators() {
```
IsaacLic, line 91: I think the discussion in Discord about random vs selected generators is worthwhile, but shouldn't hold back the basic worldgen structure from progressing. If you rename this something like `initializeSolarSystemGeneratorsUsingRandomGenerators()` and rename `build()` to `buildUsingRandomGenerators()`, that leaves open the possibility of using selected generators without needing to code a specific implementation of that idea. Lmk if that makes sense to you

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+            solSystemGenerator.build();
+            systemsBuilt++;
+        }
+    }
+
+    public void build() {
+        initializeSolSystemGenerators();
+    }
+
+    public int getSystemsBuilt() {
```
IsaacLic, line 104: What does this do? if this is only for testing, a comment to make that clear could be useful :)

engine/src/main/java/org/destinationsol/world/generators/FeatureGenerator.java
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+/**
+ * This class represents the abstract version of any feature which will populate the game's SolSystems, such as Planets,
```
IsaacLic, line 19: Will the subclasses _create_ the planets/mazes, or will they _be_ the planets/mazes? It could be that I'm not understanding how you're using the word "feature", but it sounds like the subclasses will be instances of features, meaning that the subclasses would be Planet instances.

engine/src/main/java/org/destinationsol/world/generators/SolSystemGeneratorImpl.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.destinationsol.world.generators;
+
+/**
+ * This class is a concrete implementation of a SolSystem and handles creation of elements
+ * specific to this type of SolSystem (such as how many Planets to generate, how large to make
+ * the SolSystem, etc). It defines the behavior specific to the default Planets of Destination: Sol.
+ * This class also has access to the featureGenerators list from {@link SolSystemGenerator}.
+ * This allows it to choose which FeatureGenerators to use in populating the SolSystem.
+ */
```
IsaacLic, line 24: TODO here

engine/src/test/java/org/destinationsol/world/WorldBuilderTest.java
```diff
@@ -0,0 +1,73 @@
...
+        moduleManager.init();
+        context.put(ModuleManager.class, moduleManager);
+    }
+
+    @Test
+    void populatesSolSystemsList() {
+        int testNumberSystems = 2;
+        worldBuilder = new WorldBuilder(context, testNumberSystems);
+        worldBuilder.build();
+        assertNotNull(worldBuilder.getSolSystemGenerators());
```
IsaacLic, line 46: This checks asserts that the list itself is not null, not that the generators themselves exist. checking its size might be more in line with what you're trying to do

engine/src/test/java/org/destinationsol/world/WorldBuilderTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    }
+
+    @Test
+    void populatesFeatureGeneratorsList() {
+        int testNumberSystems = 2;
+        worldBuilder = new WorldBuilder(context, testNumberSystems);
+        worldBuilder.build();
+        assertNotNull(worldBuilder.getFeatureGenerators());
```
IsaacLic, line 55: Same here

IsaacLic (commented):
This looks pretty good! I have some minor comments, but nothing that'd be a big refactoring or anything

engine/src/main/java/org/destinationsol/world/generators/FeatureGenerator.java
IsaiahBlanks: @IsaacLic Yeah I think the word 'be' is more correct. I had meant create as in place themselves into the world. Meaning each SolarSystemGenerator will pick whichever FeatureGenerators it wants to use, and then those FeatureGenerators will be capable of placing an instance of themselves into the game

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
IsaiahBlanks: @IsaacLic @Cervator mentioned that putting comments on getters and setters is probably not necessary. The field it retrieves already has a comment by it.

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+public class WorldBuilder {
+    /**
+     * These arrays hold instances of any class which extends SolSystemGenerator or FeatureGenerator, respectively.
+     */
+    private ArrayList<SolSystemGenerator> solSystemGenerators = new ArrayList<>();
+    private ArrayList<FeatureGenerator> featureGenerators = new ArrayList<>();
+    private Context context;
+    private final int numberOfSystems;
+    private int systemsBuilt; //This field is for testing purposes
```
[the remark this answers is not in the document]
IsaiahBlanks: @IsaacLic see the message I posted in discord

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
IsaiahBlanks: @IsaacLic I think the change you are suggesting makes sense (though I might think `initializeRandomSolarSystemGenerators()` might by a little less wordy. What do you think?)

I do wonder what this would look like in `SolGame`?  I suppose that from the UI there might be a string passed in that would contain the world-gen packages the use wants to enable. If the string is empty, then call `buildUsingRandomGenerators()`. If not, then call `buildUsingPackageGenerators(String s)`

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * This method is intended to first set up the SolarSystem during world generation and then initialize all the
+     * FeatureGenerators of that SolarSystem
+     * @return
+     */
+    public abstract void build();
+
+    public void setFeatureGenerators(ArrayList<FeatureGenerator> generators) {
+        featureGenerators = (ArrayList<FeatureGenerator>) generators.clone();
```
NicholasBatesNZ, line 38: Two questions:
1) Do we really need the cast here?
2) Is it necessary to take a deep copy or could we instead use the references with `featureGenerators.addAll(generators)`?

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGeneratorImpl.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * This class also has access to the featureGenerators list from {@link SolarSystemGenerator}.
+ * This allows it to choose which FeatureGenerators to use in populating the SolarSystem.
+ * TODO: Define the behavior of default Planets in this class (As it is implemented in the game currently)
+ */
+public class SolarSystemGeneratorImpl extends SolarSystemGenerator {
+
+    @Override
+    public void build() {
+        for(FeatureGenerator generator : featureGenerators) {
```
NicholasBatesNZ, line 31: `for(` -> `for (` for consistency
(could also use a `.forEach` lambda but that's not important :P )

engine/src/test/java/org/destinationsol/world/WorldBuilderTest.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+package org.destinationsol.world;
+
+import org.destinationsol.game.context.Context;
+import org.destinationsol.game.context.internal.ContextImpl;
+import org.destinationsol.modules.ModuleManager;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
```
NicholasBatesNZ, line 24: star import

NicholasBatesNZ (changes requested):

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
IsaacLic: Sounds good to me :)

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGeneratorImpl.java
```diff
@@ -0,0 +1,35 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+/**
+ * This class is a concrete implementation of a SolarSystem and handles creation of elements
```
IsaacLic, line 19: SolarSystemGenerator ?

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
```diff
@@ -0,0 +1,118 @@
...
+import org.destinationsol.world.generators.FeatureGenerator;
+import org.destinationsol.world.generators.SolarSystemGenerator;
+
+import java.lang.reflect.Modifier;
+import java.util.*;
+
+/**
+ * This class is the starting point for world generation. When a new world is created, this class first retrieves
+ * all Generator classes and then initiates each SolarSystem's build process. Two types of world generation classes are
+ * retrieved: those that subclass SolarSystemGenerator and those at subclass FeatureGenerator.
```
IsaacLic, line 30: those _that_ subclass FetureGenerator

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
IsaacLic: I believe `arraylist.clone()` actually makes a shallow copy, so either could work

PR #597, opened 2021-06-23
Initial SolarSystemGenerator design

IsaiahBlanks:
# Description
This PR defines some initial sections of the SolarSystemGenerator class. This class is an abstract class that will handle much of the logic behind SolarSystem generation. So far, this code creates placements for SolarSystems and Mazes, and generates the correct number of SolarSystemGenerators, MazeGenerators, and PlanetGenerators.

# Testing
When you run the game, you will see print statements describing what position has been set for each generator created (these do not show up in the actual game yet).

# Design Questions

1. I'm not sure about my use of Class variable to instantiate new Maze and Planet generators in the SolarSystemGenerator class. Previously, the code just called featureGenerator.build() on each available FeatureGenerator. The issue with this is that each FeatureGenerator actually needs to be initialized as it's own object, instead of being a reference to the one is the list (So that they can each have their own position, size, etc.). However, I'm unsure if this is the best way to handle that. 

2. Another issue is the order of world generation. Currently, WorldBuilder places each SolarSystemGenerator before calling the build method of each system. This means that the radius of the System has to be determined before build() is called. Therefore, the value of how many planets the SolarSystemGenerator will create needs to be known before any methods from SolarSystemGeneratorImpl can be called. Therefore, the only way for SolarSystemGeneratorImplementations to set the number of planets before their build() method is called is to pass it into the super() constructor. That way it can be taken into account when the superclass's constructor builds the system radius. 
- It might seem like it would be best to create the SystemRadius during the build() phase of SolarSystemGenerator, however that is problematic because it would mean the WorldBuilder would have to position the System _after_ the build method of SolarSystemGenerator. This would position Mazes wrong, as it would not be possible to place them relative to the SolarSystem position. The WorldBuilder _could_ place the Mazes, but then it would need references to each MazeGeneratorImpl from the SolarSystemGenerators. Also, since SolarSystemGenerators have mazes associated to them in this design, it seems it would be best for them to handle the Maze placement.


IsaacLic (commented):
If I'm a modder and I want to change one element, how much will I have to re-write the same code, but with a small tweak? For example, what if I want to make a system with a black hole instead of a sun - will I have to rewrite the whole `SolarSystemGenerator` to change the radius calculation?

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
```diff
@@ -79,33 +83,99 @@ private void populateFeatureGeneratorList() {
...
+     * This method initiates the build process of each SolarSystemGenerator instance.
+     */
+    private void buildSolarSystems() {
+        for (SolarSystemGenerator solarSystemGenerator : activeSolarSystemGenerators) {
+            solarSystemGenerator.build();
+        }
+    }
+
+    /**
+     * This method runs a loop which tests 20 random angles at increasing radii starting from the center of the world
```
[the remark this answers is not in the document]
IsaiahBlanks: Yep! This is the game's original logic

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGeneratorImpl.java
```diff
@@ -22,14 +22,34 @@
...
 public class SolarSystemGeneratorImpl extends SolarSystemGenerator {
 
     @Override
     public void build() {
-        for (FeatureGenerator generator : featureGenerators) {
-            generator.build();
+        setPlanetCount(5);
+        setPossibleBeltCount(1);
+        setMazeCount(2);
+        setSunCount(1);
```
IsaacLic, line 34: These four lines seem to me like they should really be one method with four parameters, possibly in the abstract class's constructor

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGeneratorImpl.java
```diff
@@ -22,14 +22,34 @@
...
+        initializeRandomSunGenerators();
+        initializeRandomPlanetGenerators();
+        initializeRandomMazeGenerators();
+        initializeRandomBeltGenerators(.8f);
```
IsaacLic, line 38: These do not seem like they require a specific order, even though they do

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGeneratorImpl.java
```diff
@@ -22,14 +22,34 @@
...
+        buildFeatureGenerators();
+        setRadius(calcSolarSystemRadius());
+        calculateSunPositionOneSun();
+        calculateMazePositions();
+        calculatePlanetPositions();
+        calculateBeltPositions();
+
```
IsaacLic, line 45: Is there a way that this could be partially or fully moved into the abstract class? Modders will have to copy-paste several lines of code for the basic build process to happen

IsaacLic (commented):
I have two main thoughts:
1) I don't think that any code should specifically address the concept of multiple suns - see my comments below.
2) If there's a particular order for how things need to get invoked, make sure that it always happens that way (again, see my comments below).

This is very exciting progress!

engine/src/main/java/org/destinationsol/world/generators/BeltGeneratorImpl.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package org.destinationsol.world.generators;
+
+public class BeltGeneratorImpl extends BeltGenerator {
+
+
+    @Override
+    public void build() {
+        System.out.println("Building a belt now!");
+        setRadius(DEFAULT_BELT_HALF_WIDTH);
```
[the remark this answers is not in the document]
IsaiahBlanks: That would be diameter (width means from one side to the other). I took that name from the current codebase

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +30,417 @@
...
+    protected static final float MAX_MAZE_RADIUS = 40f;
+    protected static final float MAZE_GAP = 10f;
+
+    //This field is protected to allow subclasses to access it
+    protected ArrayList<FeatureGenerator> featureGeneratorTypes = new ArrayList<>();
+    ArrayList<FeatureGenerator> activeFeatureGenerators = new ArrayList<>();
+    private Vector2 position;
+    private float radius;
+    private boolean positioned;
+    private int planetCount = 5;
```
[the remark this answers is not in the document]
IsaiahBlanks: I meant to have these numbers here as the 'default' values, so if a modder in their implementation didn't want to, they wouldn't have to call setPlanetCount() or whichever value they don't want to set

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGeneratorImpl.java
IsaiahBlanks: These actually do not have a specific order. Each one just get random Planet, Maze, Belt, or Sun types and adds instances of them to the activeFeatureGenerators list

engine/src/main/java/org/destinationsol/game/planet/BeltConfigManager.java
```diff
@@ -0,0 +1,121 @@
...
+            boolean hard = node.optBoolean("hard", false);
+            Map<String, BeltConfig> configsToLoad = hard ? hardConfigs : configs;
+
+            SpaceEnvConfig envConfig = new SpaceEnvConfig(node.getJSONObject("environment"));
+
+            BeltConfig beltConfig = new BeltConfig(name, new ArrayList<>(), envConfig, new ArrayList<>(), new ArrayList<>(), new TradeConfig(), new ArrayList<>(), hard);
+            configsToLoad.put(name, beltConfig);
+        }
+
+        Set<ResourceUrn> configUrnList = Assets.getAssetHelper().listAssets(Json.class, "asteroidBeltsConfig", new Name("engine"));
```
IsaacLic, line 66: See above about generification

engine/src/main/java/org/destinationsol/game/planet/BeltConfigManager.java
```diff
@@ -0,0 +1,121 @@
...
+
+        for (ResourceUrn configUrn : configUrnList) {
+            rootNode = Validator.getValidatedJSON(configUrn.toString(), "engine:schemaSystemsConfig");
```
IsaacLic, line 69: same

engine/src/main/java/org/destinationsol/game/planet/BeltConfigManager.java
```diff
@@ -0,0 +1,121 @@
...
+                }
+                JSONObject node = rootNode.getJSONObject(s);
+                String name = s;
+
+                boolean hard = node.optBoolean("hard", false);
+                Map<String, BeltConfig> configsToLoad = hard ? hardConfigs : configs;
+
+                BeltConfig config = configsToLoad.get(name);
+
+                // TODO : Maybe add sanity checks for config?
```
IsaacLic, line 83: What do you mean?

engine/src/main/java/org/destinationsol/game/planet/BeltConfigManager.java
```diff
@@ -0,0 +1,121 @@
...
+
+                config.tempEnemies.addAll(ShipConfig.loadList(node.has("temporaryEnemies") ? node.getJSONArray("temporaryEnemies") : null, hullConfigs, itemManager));
+                config.innerTempEnemies.addAll(ShipConfig.loadList(node.has("innerTemporaryEnemies") ? node.getJSONArray("innerTemporaryEnemies") : null, hullConfigs, itemManager));
```
IsaacLic, line 86: What do these do?

engine/src/main/java/org/destinationsol/game/planet/SolarSystemConfig.java
```diff
@@ -0,0 +1,48 @@
...
+    public final String name;
+    public final ArrayList<ShipConfig> tempEnemies;
+    public final SpaceEnvConfig envConfig;
+    public final ArrayList<ShipConfig> constEnemies;
+    public final ArrayList<ShipConfig> constAllies;
+    public final TradeConfig tradeConfig;
+    public final ArrayList<ShipConfig> innerTempEnemies;
+    public final boolean hard;
+
+    public SolarSystemConfig(String name, ArrayList<ShipConfig> tempEnemies, SpaceEnvConfig envConfig,
```
IsaacLic, line 35: I wonder if the Builder pattern might be better here, to avoid having a bunch of parameters listed like this

IsaacLic (commented):
Here's part one of my review - stopping for tonight

engine/src/main/java/org/destinationsol/game/planet/SolarSystemConfig.java
IsaiahBlanks: @IsaacLic Turns out most of the parameters are pretty useless, as the are only used to pass in empty arrays or objects, and then those are set from JSON data by the load() method. I'm changing it so that those empty objects are initialized in the constructor instead of being passed in

engine/src/main/java/org/destinationsol/game/planet/SolarSystemConfigManager.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public SolarSystemConfigManager(HullConfigManager hullConfigs, ItemManager itemManager) {
+        configs = new HashMap<>();
+        hardConfigs = new HashMap<>();
+
+        load(hullConfigs, false, itemManager);
+    }
+
+    private void load(HullConfigManager hullConfigs, boolean b, ItemManager itemManager) {
+        JSONObject rootNode = Validator.getValidatedJSON("engine:systemsConfig", "engine:schemaSystemsConfig");
```
IsaacLic, line 48: See `BeltConfigManager` about my JSON comments

engine/src/main/java/org/destinationsol/game/planet/SolarSystemConfigManager.java
```diff
@@ -0,0 +1,124 @@
...
+            boolean hard = node.optBoolean("hard", false);
+            Map<String, SolarSystemConfig> configsToLoad = hard ? hardConfigs : configs;
+
+            SpaceEnvConfig envConfig = new SpaceEnvConfig(node.getJSONObject("environment"));
+
+            SolarSystemConfig solarSystemConfig = new SolarSystemConfig(name, new ArrayList<>(), envConfig, new ArrayList<>(), new ArrayList<>(), new TradeConfig(), new ArrayList<>(), hard);
+            configsToLoad.put(name, solarSystemConfig);
+        }
+
+        Set<ResourceUrn> configUrnList = Assets.getAssetHelper().listAssets(Json.class, "systemsConfig", new Name("engine"));
```
IsaacLic, line 66: This should be made module-friendly

engine/src/main/java/org/destinationsol/world/Orbital.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+package org.destinationsol.world;
+
+import org.destinationsol.world.generators.FeatureGenerator;
+
+/**
+ * This class represents one 'ring' around a SolarSystem's center. It is used to keep track of what Features a SolarSystemGenerator
+ * should generate. It keeps track of where those Features will be placed, and whether or not a Feature has been chosen
+ * for a specific orbital. Each orbital has a width, which is the distance from the inner edge of the orbital to the
+ * outer edge.
```
IsaacLic, line 24: Excellent javadoc

engine/src/main/java/org/destinationsol/world/Orbital.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+public class Orbital {
+    int positionInSolarSystem;
+    float width;
+    float startingDistanceFromSystemCenter;
```
IsaacLic, line 29: What's the difference between this and `positionInSolarSystem`?

engine/src/main/java/org/destinationsol/world/Orbital.java
```diff
@@ -0,0 +1,82 @@
...
+     * @return distance from the center of the SolarSystem to place the Feature
+     */
+    public float calculateDistanceFromCenterOfSystemForFeature() {
+        return getStartingDistanceFromSystemCenter() + (getWidth() / 2);
+    }
+
+    public String getSuperTypeName() {
+        return superTypeName;
+    }
+}
```
IsaacLic, line 82: Small organizational thing: I'd pair each getter with each setter, rather than having all the setters followed by all the getters.

engine/src/main/java/org/destinationsol/world/WorldBuilder.java
```diff
@@ -69,47 +69,115 @@ private void populateSolarSystemGeneratorList() {
...
-        initializeRandomSolarSystemGenerators();
+    /**
+     * This method iterates through the loop of SolarSystemGenerators that have been initialized and assigns each of
+     * them a position in the world.
+     */
+    private void positionSolarSystems() {
+        for (SolarSystemGenerator generator : activeSolarSystemGenerators) {
+            Vector2 position = calculateSolarSystemPosition(activeSolarSystemGenerators, generator.getRadius());
+            generator.setPosition(position);
+            //Printout of generator position for for testing (as these positions don't have a representation in the game yet)
```
IsaacLic, line 117: "for for"

engine/src/main/java/org/destinationsol/game/planet/BeltConfig.java
```diff
@@ -0,0 +1,48 @@
...
+public class BeltConfig {
+
+    public final String name;
+    public final ArrayList<ShipConfig> tempEnemies;
+    public final SpaceEnvConfig envConfig;
+    public final ArrayList<ShipConfig> constEnemies;
+    public final ArrayList<ShipConfig> constAllies;
+    public final TradeConfig tradeConfig;
+    public final ArrayList<ShipConfig> innerTempEnemies;
+    public final boolean hard;
```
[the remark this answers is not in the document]
IsaiahBlanks: Unfortunately it will be hard to change the name, as it needs to be named the same as the key in the JSON will be named

engine/src/test/java/org/destinationsol/world/generators/SolarSystemGeneratorTest.java
```diff
@@ -0,0 +1,147 @@
...
+        if(solarSystemGenerator.getClass().equals(SolarSystemGeneratorImpl.class)) {
+            assertEquals(solarSystemGenerator.getPossibleBeltCount(), 1);
+        }
+    }
+
+    @Test
+    void hasCorrectRadius() {
+        float radiusFromFeatures = 0;
+        radiusFromFeatures += SunGenerator.SUN_RADIUS;
+        for (FeatureGenerator featureGenerator : solarSystemGenerator.getActiveFeatureGenerators()) {
```
IsaacLic, line 83: Shouldn't this only be adding the Orbital radius size?

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -16,38 +16,90 @@
...
  */
 public abstract class SolarSystemGenerator {
-    protected static final float MAX_MAZE_RADIUS = 40f;
-    protected static final float MAZE_GAP = 10f;
+
+    /**
+     * This enum represents the three available sizes of SolarSystems. They can have either 3, 5, or 7 orbital objects
+     */
+    public enum SolarSystemSize {
+        SMALL(3), MEDIUM(5), LARGE(7);
```
IsaacLic, line 42: Interesting design decision - let's discuss on Discord

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -16,38 +16,90 @@
...
+        }
+        sizeSolarSystemOrbitals();
+        this.radius = calcSolarSystemRadius();
+    }
+
+    private void sizeSolarSystemOrbitals() {
+        float distanceFromSolarSystemCenter = SUN_RADIUS;
+        for (int i = 0; i < solarSystemOrbitals.size(); i++) {
+            solarSystemOrbitals.get(i).setStartingDistanceFromSystemCenter(distanceFromSolarSystemCenter);
+            solarSystemOrbitals.get(i).setWidth(ORBITAL_FEATURE_BUFFER + PlanetGenerator.PLANET_MAX_DIAMETER + ORBITAL_FEATURE_BUFFER);
```
IsaacLic, line 96: This should probably be a single constant

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -85,13 +151,15 @@ protected void calculateSunPositionOneSun() {
...
      * degrees apart so they do not overlap
      */
-    protected void calculateMazePositions() {
+    private void calculateMazePositions() {
         for (FeatureGenerator generator : activeFeatureGenerators) {
             Vector2 result = new Vector2();
             List<Float> usedAnglesInSolarSystem = new ArrayList<>();
             if (generator.getClass().getSuperclass().equals(MazeGenerator.class)) {
+                //set the outermost orbital to have a MazeGenerator
+                solarSystemOrbitals.get(solarSystemOrbitals.size() - 1).setFeatureGenerator(generator);
```
IsaacLic, line 160: Is there a way that we can make use of multiple `MazeGenerators` in one system?

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -134,44 +200,49 @@ protected void calculatePlanetPositions() {
...
-                float distanceFromSolarSystemCenter = planetDistances.get(orbitInsertionPosition);
+                int orbitInsertionPosition = chooseOrbitInsertionPosition();
+                float distanceFromSolarSystemCenter = solarSystemOrbitals.get(orbitInsertionPosition).calculateDistanceFromCenterOfSystemForFeature();
+                ((BeltGenerator) featureGenerator).setDistanceFromCenterOfSolarSystem(distanceFromSolarSystemCenter);
 
-                distanceFromSolarSystemCenter += Const.PLANET_GAP;
-                distanceFromSolarSystemCenter += featureGenerator.getRadius();
-                ((BeltGenerator) featureGenerator).setDistanceFromCenter(distanceFromSolarSystemCenter);
+                //remove the planet that was going to be placed in this orbital
+                featureGeneratorsToRemove.add(solarSystemOrbitals.get(orbitInsertionPosition).getFeatureGenerator());
```
IsaacLic, line 214: I'm not clear on what this does

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -134,44 +200,49 @@ protected void calculatePlanetPositions() {
...
+        calculatePlanetPositions();
+        calculateBeltPositions();
     }
 
     /**
-     * This will create new instances of SunGenerator implementations from the list of available feature generators.
+     * This will create a new instance of a SunGenerator implementation from the list of available feature generators.
      * It will only choose FeatureGenerators which are subclasses of {@link SunGenerator}.
-     * It will continue to make new instances until the correct number of Suns is reached
+     * It will make one sun generator instance.
```
IsaacLic, line 235: I think this line of comment is likely unnecessary

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -134,44 +200,49 @@ protected void calculatePlanetPositions() {
...
-                Class<? extends FeatureGenerator> newFeatureGeneratorType = featureGeneratorTypes.get(index).getClass();
-                try {
-                    FeatureGenerator newFeatureGenerator = newFeatureGeneratorType.newInstance();
-                    activeFeatureGenerators.add(newFeatureGenerator);
-                    sunsLeft--;
-                } catch (Exception e) {
-                    e.printStackTrace();
-                }
+    protected void initializeRandomSunGenerator() {
+        int index = SolRandom.seededRandomInt(featureGeneratorTypes.size());
```
IsaacLic, line 238: You do still need a while loop to ensure that the sun actually gets created

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -204,11 +275,12 @@ protected void initializeRandomPlanetGenerators() {
      * It will continue to make new instances until the correct number of Mazes is reached
      */
     protected void initializeRandomMazeGenerators() {
-        int mazesLeft = getMazeCount();
+        //we will initialize up to 12 mazes to ensure they fit around the SolarSystem
```
IsaacLic, line 278: Good thinking

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -273,75 +353,30 @@ public void buildFeatureGenerators() {
...
-    }
-
     /**
      * This method checks if two angles are within a certain number of degrees of each other or not
-     * @param firstAngleToCheck angle to check
+     *
+     * @param firstAngleToCheck  angle to check
      * @param secondAngleToCheck angle to compare it to
-     * @param degrees how many degrees apart to ensure angles are between
+     * @param degrees            how many degrees apart to ensure angles are between
```
IsaacLic, line 379: Probably too much space ;)

IsaacLic (commented):
Mostly minor comments

engine/src/main/java/org/destinationsol/world/Orbital.java
IsaiahBlanks: @IsaacLic positionInSolarSystem represents the cardinal order of this orbital. Like the orbital closes to the sun is orbital 0, then orbital 1, etc. I'll come up with a better name

engine/src/test/java/org/destinationsol/world/generators/SolarSystemGeneratorTest.java
IsaiahBlanks: The sun adds radius to the total SolarSystem radius while not being in an orbital (since it is in the center of the system it doesn't orbit anything, but it still has width so it contributes to the radius).

Maybe a good way to think of it is that the first orbital starts at the edge of the sun

engine/src/main/java/org/destinationsol/world/generators/PlanetGenerator.java
```diff
@@ -15,11 +15,53 @@
...
+    public static final float PLANET_MAX_DIAMETER = 78f;
+    protected static final float DEFAULT_MAX_GROUND_HEIGHT = 25f;
+    protected static final float DEFAULT_ATMOSPHERE_HEIGHT = 14f;
+    float groundHeight;
+    //distance from ground of planet to atmosphere
+    float atmosphereHeight;
+    float angleInSolarSystem;
+    float distanceFromSolarSystemCenter;
+
+    public void setAtmosphereHeight(float aHeight) {
```
NicholasBatesNZ, line 35: Please avoid single letter prefixes too - `float atmosphereHeight`

engine/src/main/java/org/destinationsol/world/generators/PlanetGenerator.java
```diff
@@ -15,11 +15,53 @@
...
+        atmosphereHeight = aHeight;
+    }
+
+    public float getAtmosphereHeight() {
+        return atmosphereHeight;
+    }
+
+    public void setGroundHeight(float gHeight) {
```
NicholasBatesNZ, line 43: `groundHeight`

engine/src/main/java/org/destinationsol/world/generators/SunGeneratorImpl.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * This class is a concrete implementation of a SunGenerator and handles its creation. This class defines the
+ * behavior specific to the default Sun of Destination: Sol.
+ * TODO: Implement the default behavior of a SunGenerator in this class (As it is implemented in the game currently)
+ */
+public class SunGeneratorImpl extends SunGenerator{
+    @Override
+    public void build() {
+        System.out.println("Building a sun now!");
```
NicholasBatesNZ, line 26: logger

engine/src/main/java/org/destinationsol/game/HardnessCalc.java
```diff
@@ -190,11 +192,11 @@ public static float getMazeDps(MazeConfig c) {
         return res * 1.25f;
     }
 
-    public static float getBeltDps(SysConfig c) {
```
NicholasBatesNZ, line 193: Are there unused imports now due to this?

engine/src/main/java/org/destinationsol/game/planet/BeltConfigManager.java
```diff
@@ -0,0 +1,132 @@
...
+            boolean hard = node.optBoolean("hard", false);
+            Map<String, BeltConfig> configsToLoad = hard ? hardConfigs : configs;
+
+            SpaceEnvConfig envConfig = new SpaceEnvConfig(node.getJSONObject("environment"));
+
+            BeltConfig beltConfig = new BeltConfig(name, envConfig, hard);
+            configsToLoad.put(name, beltConfig);
+        }
+
+        Set<ResourceUrn> configUrnList = Assets.getAssetHelper().listAssets(Json.class, assetType, new Name(moduleName));
```
NicholasBatesNZ, line 79: Why are you excluding assets from this module?

engine/src/main/java/org/destinationsol/game/planet/BeltConfigManager.java
```diff
@@ -0,0 +1,132 @@
...
+        for (BeltConfig sc : configs.values()) {
+            shipConfigs.addAll(sc.tempEnemies);
+            shipConfigs.addAll(sc.innerTempEnemies);
+        }
+
+        for (BeltConfig sc : hardConfigs.values()) {
+            shipConfigs.addAll(sc.tempEnemies);
+            shipConfigs.addAll(sc.innerTempEnemies);
+        }
+    }
```
NicholasBatesNZ, line 131: Javadocs please :)

engine/src/main/java/org/destinationsol/game/planet/BeltConfigManager.java
```diff
@@ -0,0 +1,132 @@
...
+        return hardConfigs.get(name);
+    }
+
+    public BeltConfig getRandomBeltConfig(boolean hard) {
+        Map<String, BeltConfig> config = hard ? hardConfigs : configs;
+        return SolRandom.seededRandomElement(new ArrayList<>(config.values()));
+    }
+
+    public void addAllConfigs(ArrayList<ShipConfig> shipConfigs) {
+        for (BeltConfig sc : configs.values()) {
```
NicholasBatesNZ, line 122: two letter variable name

engine/src/main/java/org/destinationsol/game/planet/SolarSystemConfigManager.java
```diff
@@ -0,0 +1,141 @@
...
+    private HullConfigManager hullConfigManager;
+    private ItemManager itemManager;
+
+    public SolarSystemConfigManager(HullConfigManager hullConfigManager, ItemManager itemManager) {
+        configs = new HashMap<>();
+        hardConfigs = new HashMap<>();
+        this.hullConfigManager = hullConfigManager;
+        this.itemManager = itemManager;
+
+        //load(hullConfigs, false, itemManager);
```
NicholasBatesNZ, line 47: What's this?

engine/src/main/java/org/destinationsol/game/planet/SolarSystemConfigManager.java
```diff
@@ -0,0 +1,141 @@
...
+            boolean hard = node.optBoolean("hard", false);
+            Map<String, SolarSystemConfig> configsToLoad = hard ? hardConfigs : configs;
+
+            SpaceEnvConfig envConfig = new SpaceEnvConfig(node.getJSONObject("environment"));
+
+            SolarSystemConfig solarSystemConfig = new SolarSystemConfig(name, envConfig, hard);
+            configsToLoad.put(name, solarSystemConfig);
+        }
+
+        Set<ResourceUrn> configUrnList = Assets.getAssetHelper().listAssets(Json.class, assetType, new Name(moduleName));
```
NicholasBatesNZ, line 81: Why are you excluding this module?

engine/src/main/java/org/destinationsol/game/planet/SolarSystemConfigManager.java
```diff
@@ -0,0 +1,141 @@
...
+            shipConfigs.addAll(sc.innerTempEnemies);
+        }
+
+        for (SolarSystemConfig sc : hardConfigs.values()) {
+            shipConfigs.addAll(sc.constAllies);
+            shipConfigs.addAll(sc.constEnemies);
+            shipConfigs.addAll(sc.tempEnemies);
+            shipConfigs.addAll(sc.innerTempEnemies);
+        }
+    }
```
NicholasBatesNZ, line 139: Javadocs?

engine/src/main/java/org/destinationsol/game/planet/SolarSystemConfigManager.java
```diff
@@ -0,0 +1,141 @@
...
+        return hardConfigs.get(name);
+    }
+
+    public SolarSystemConfig getRandomSolarSystemConfig(boolean hard) {
+        Map<String, SolarSystemConfig> config = hard ? hardConfigs : configs;
+        return SolRandom.seededRandomElement(new ArrayList<>(config.values()));
+    }
+
+    public void addAllConfigs(ArrayList<ShipConfig> shipConfigs) {
+        for (SolarSystemConfig sc : configs.values()) {
```
NicholasBatesNZ, line 126: two letter variable name

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGeneratorImpl.java
```diff
@@ -15,21 +15,47 @@
...
+    @Override
+    public SolarSystem build() {
+        getSolarSystemConfigManager().loadDefaultSolarSystemConfigs();
+        setSolarSystemConfig(getSolarSystemConfigManager().getRandomSolarSystemConfig(getSolarSystemNumber() < 1));
+        setName(SolRandom.seededRandomElement(getDefaultSolarSystemNames()));
+        initializeRandomDefaultFeatureGenerators(.8f);
+        buildFeatureGenerators();
+        calculateFeaturePositions();
+
+        System.out.println(this + " radius: " + getRadius());
```
NicholasBatesNZ, line 52: Logger

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGeneratorImpl.java
```diff
@@ -15,21 +15,47 @@
...
+
+        //Just temporary to see where everything is placed
```
NicholasBatesNZ, line 54: ``//TODO:``

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGeneratorImpl.java
```diff
@@ -15,21 +15,47 @@
...
+        for (FeatureGenerator generator : activeFeatureGenerators) {
+            System.out.println(generator + ": " + generator.getPosition());
```
NicholasBatesNZ, line 56: Logger

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+    private boolean positioned;
+    private int solarSystemNumber;
+    private int planetCount;
+    private int possibleBeltCount;
+    private int mazeCount;
+    private int customFeaturesCount;
+
+    public SolarSystemGenerator() {
+
+        position = Vector2.Zero;
```
NicholasBatesNZ, line 79: vector pool

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+        solarSystemSize = getSolarSystemSize();
+        //There cannot be more custom features than orbitals available
+        if (customFeaturesCount < solarSystemSize.getNumberOfOrbitals()) {
+            customFeaturesCount = getCustomFeaturesCount();
+        } else {
+            customFeaturesCount = solarSystemSize.getNumberOfOrbitals();
+        }
```
NicholasBatesNZ, line 86: Maybe add a newline here

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+        if (solarSystemSize.equals(SolarSystemSize.SMALL) || solarSystemSize.equals(SolarSystemSize.MEDIUM)) {
+            //Custom Features will replace planets, depending on how many are being used
+            planetCount = solarSystemSize.getNumberOfOrbitals() - getCustomFeaturesCount();
+            possibleBeltCount = 1;
+            mazeCount = 2;
+        } else {
+            planetCount = solarSystemSize.getNumberOfOrbitals() - getCustomFeaturesCount();
+            possibleBeltCount = 2;
+            mazeCount = 3;
+        }
```
NicholasBatesNZ, line 96: Maybe add a newline here

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+     *
+     * @return number of custom features
+     */
+    public abstract int getCustomFeaturesCount();
+
+    /**
+     * This method calculates the radius for this SolarSystem. It uses the Features that are included in this system
+     * to determine what the radius should be.
+     */
+    protected float calcSolarSystemRadius() {
```
NicholasBatesNZ, line 146:
```suggestion
    protected float calculateSolarSystemRadius() {
```

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+        float solarSystemRadius = SUN_RADIUS;
+        for (Orbital orbital : solarSystemOrbitals) {
+            solarSystemRadius += orbital.getWidth();
+            System.out.println("Orbital " + orbital.getPositionInSolarSystem() + ", distance: " + orbital.getStartingDistanceFromSystemCenter() + ", width: " + orbital.getWidth());
```
NicholasBatesNZ, line 150: logger

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+        }
+    }
+
+    /**
+     * This places mazes within the outer orbital of the solar system. It checks to make sure they are placed at least 15
+     * degrees apart so they do not overlap
+     */
+    private void calculateMazePositions() {
+        for (FeatureGenerator generator : activeFeatureGenerators) {
+            Vector2 result = new Vector2();
```
NicholasBatesNZ, line 172: vector pool

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+            List<Float> usedAnglesInSolarSystem = new ArrayList<>();
+            if (generator.getClass().getSuperclass().equals(MazeGenerator.class)) {
+                //set the outermost orbital to have a MazeGenerator
+                solarSystemOrbitals.get(solarSystemOrbitals.size() - 1).setFeatureGenerator(generator);
+                float angle = SolRandom.seededRandomFloat(180);
+                SolMath.fromAl(result, angle, solarSystemOrbitals.get(solarSystemOrbitals.size() - 1).calculateDistanceFromCenterOfSystemForFeature());
```
NicholasBatesNZ, line 178: Same radian degrees comment as eariler

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+    /**
+     * This method goes through every active PlanetGenerator and assigns a position to it in the SolarSystem. That
+     * position will be based of a random starting angle in the system, which is generated by the PlanetGenerator. It adds
+     * a small gap between the orbit of each planet, so they do not touch. The width of each orbit is based on the Planet's
+     * radius, which is determined by the PlanetGenerator.
+     */
+    private void calculatePlanetPositions() {
+        int orbitalPosition = 0;
+        for (FeatureGenerator featureGenerator : activeFeatureGenerators) {
+            Vector2 result = new Vector2();
```
NicholasBatesNZ, line 197: vector pool

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+                PlanetGenerator generator = (PlanetGenerator) featureGenerator;
+                Orbital orbital = solarSystemOrbitals.get(orbitalPosition);
+                orbital.setFeatureGenerator(generator);
+
+                generator.setDistanceFromSolarSystemCenter(orbital.calculateDistanceFromCenterOfSystemForFeature());
+                SolMath.fromAl(result, generator.angleInSolarSystem, generator.getDistanceFromSolarSystemCenter());
+                result.add(position);
+                generator.setPosition(result);
+
+                System.out.println(generator + " distance from center: " + generator.distanceFromSolarSystemCenter);
```
NicholasBatesNZ, line 208: logger

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+                int orbitInsertionPosition = chooseOrbitInsertionPosition();
+                float distanceFromSolarSystemCenter = solarSystemOrbitals.get(orbitInsertionPosition).calculateDistanceFromCenterOfSystemForFeature();
+                ((BeltGenerator) featureGenerator).setDistanceFromCenterOfSolarSystem(distanceFromSolarSystemCenter);
+
+                /* Since a belt will be placed at a random orbital, but it isn't guaranteed that a belt will be placed,
+                first each orbital (except the maze orbitals) have Planets inserted into them, and then when a belt
+                is placed, a planet is removed and a maze set in its place */
+                featureGeneratorsToRemove.add(solarSystemOrbitals.get(orbitInsertionPosition).getFeatureGenerator());
+                solarSystemOrbitals.get(orbitInsertionPosition).setFeatureGenerator(featureGenerator);
+                System.out.println("Distance for belt: " + distanceFromSolarSystemCenter);
```
NicholasBatesNZ, line 234: logger

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
+            orbitPosition = SolRandom.seededRandomInt(1, solarSystemOrbitals.size() - 1);
+            if (!solarSystemOrbitals.get(orbitPosition).getSuperTypeName().equals(BeltGenerator.class.getTypeName())) {
+                positionChosen = true;
+            }
+        }
+        return orbitPosition;
+    }
+
+    /**
+     * This method checks if two angles are within a certain number of degrees of each other or not
```
NicholasBatesNZ, line 399: If it only checks two angles then why is the first argument a list?

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -24,17 +37,493 @@
...
 
+    public void setRadius(float radius) {
+        this.radius = radius;
+    }
+
+    public float getRadius() {
+        return radius;
+    }
+
+    public void setPositioned(boolean p) {
```
NicholasBatesNZ, line 470: single letter var name

NicholasBatesNZ (changes requested):
My brain got fried long before I finished this review. Remind me to merge more often...

PR #602, opened 2021-07-21
Initial PlanetGenerator Setup

IsaiahBlanks:
# Description
This PR adds the default implementation of Planets to the new modular world-gen system. It also sets up the game so that SolarSystems and Planets generated by Generator classes show up in the game.

In this implementation, many options are available to the Planet via getting the config in your particular implementation, and then changing various values

# Testing
Run the game, and you should see SolarSystems running with Planets in them, but no belts and no mazes. There are two types of SolarSystems that will generate, the default systems and large SolarSystems. Generate a world with a good amount of systems (at least 4) to make sure you see the different types.

# Outstanding Work
Two features still need to be added for Planets. One is the ability to place custom objects/ modify terrain, which will involve modularizing the PlanetObjectsBuilder class to accept types of objects, instead of specific objects. The other is the ability to load custom PlanetConfig files. Additionally, tests need to be written still. There is an issue with testing PlanetGenerators as the require the PlanetConfigs class to be loaded, and this seems to be dependent on the class CollisionMeshLoader, which is throwing an error when running tests

### DIFF
--- a/engine/src/main/java/org/destinationsol/world/generators/PlanetGenerator.java
+++ b/engine/src/main/java/org/destinationsol/world/generators/PlanetGenerator.java
@@ -16,22 +16,30 @@
 package org.destinationsol.world.generators;
 
 import com.badlogic.gdx.math.Vector2;
+import org.destinationsol.common.SolMath;
+import org.destinationsol.common.SolRandom;
+import org.destinationsol.game.ShipConfig;
 import org.destinationsol.game.SolNames;
+import org.destinationsol.game.planet.DecoConfig;
 import org.destinationsol.game.planet.Planet;
 import org.destinationsol.game.planet.PlanetConfig;
 import org.destinationsol.game.planet.PlanetConfigs;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
 
 /**
  * This class defines the general behavior for Planet generators (such as ground, sky, gravity, etc). Any Planet will be
  * created from a concrete implementation of this class, with behavior specific to that Planet defined there.
- * TODO: Implement behavior common to all Planets as concrete methods in this class
  */
 public abstract class PlanetGenerator extends FeatureGenerator {
     public static final float PLANET_MAX_DIAMETER = 78f;
-    protected static final float DEFAULT_MAX_GROUND_HEIGHT = 25f;
-    protected static final float DEFAULT_ATMOSPHERE_HEIGHT = 14f;
     public static final float PLANET_SPEED = .2f;
     public static final float PLANET_GROUND_SPEED = .2f;
+    protected static final float DEFAULT_MAX_GROUND_HEIGHT = 25f;
+    protected static final float DEFAULT_ATMOSPHERE_HEIGHT = 14f;
+    private static final Logger logger = LoggerFactory.getLogger(PlanetGenerator.class);
     protected SolNames solNames = new SolNames();
     private Planet planet;
     private Vector2 solarSystemPosition = new Vector2();
@@ -45,7 +53,6 @@ public abstract class PlanetGenerator extends FeatureGenerator {
     private float angleInSolarSystem;
     private float distanceFromSolarSystemCenter;
     private float initialPlanetAngle;
-
     //This is the speed the Planet orbits around the SolarSystem at
     private float orbitSolarSystemSpeed;
     //This is the speed the Planet spins around it's axis at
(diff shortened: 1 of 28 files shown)


Cervator:
Tests out!

Was able to run the game in my Android Studio setup, although I had to delete my local `warp` module (incompatible with the worldgen changes, although probably an easy fix). Did get some weird errors from trying `gradlew clean buildDebug run` but that might just be the workspace. `gradlew jar run` worked fine after I deleted `warp`

Can confirm the game runs as described, I started with a few more systems than usual and got a nice big game world, no belts or mazes - which is fine even for merging since there's a target topic branch. Was able to travel to the nearest planet and see it looking pretty normal. One oddity: on my way there I flew by a hostile space station - I *think* those are limited to the second solar system in the classic setup? With worldgen changing it makes sense some things like that will be thrown off, the starter system should probably be special and have fairly mundane spawning rules.

It might also help this PR with a quick description of a test that'd vary a planet or planet type, although that may well be more for a tutorial like Terasology's worldgen one :-)

I did eventually crash, after arriving at a different solar system and heading toward a planet there, with a `java.lang.AssertionError: normalizing NaN angle` - DS code tends to crash rather overly hard with these assertion errors over even tiny things. We probably have existing bugs floating around still that cause the issue, but maybe this is new since the stacktrace includes some planet gen. Full stacktrace:

```
com.badlogic.gdx.utils.GdxRuntimeException: java.lang.AssertionError: normalizing NaN angle
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:117)
	at org.destinationsol.desktop.SolDesktop.main(SolDesktop.java:149)
Caused by: java.lang.AssertionError: normalizing NaN angle
	at org.destinationsol.common.SolMath.norm(SolMath.java:107)
	at org.destinationsol.common.SolMath.angle(SolMath.java:354)
	at org.destinationsol.game.planet.PlanetObjectsBuilder.buildGroundShip(PlanetObjectsBuilder.java:336)
	at org.destinationsol.game.planet.PlanetObjectsBuilder.buildGroundEnemies(PlanetObjectsBuilder.java:107)
	at org.destinationsol.game.planet.PlanetObjectsBuilder.createShips(PlanetObjectsBuilder.java:88)
	at org.destinationsol.game.planet.PlanetObjectsBuilder.createPlanetObjs(PlanetObjectsBuilder.java:75)
	at org.destinationsol.game.planet.Planet.update(Planet.java:79)
	at org.destinationsol.game.planet.PlanetManager.update(PlanetManager.java:68)
	at org.destinationsol.game.SolGame.lambda$null$2(SolGame.java:385)
	at java.util.ArrayList.forEach(ArrayList.java:1249)
	at org.destinationsol.game.SolGame.lambda$update$3(SolGame.java:384)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at org.destinationsol.game.SolGame.update(SolGame.java:383)
	at org.destinationsol.SolApplication.update(SolApplication.java:251)
	at org.destinationsol.SolApplication.safeUpdate(SolApplication.java:225)
	at org.destinationsol.SolApplication.render(SolApplication.java:189)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Window.update(Lwjgl3Window.java:392)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.loop(Lwjgl3Application.java:137)
	at com.badlogic.gdx.backends.lwjgl3.Lwjgl3Application.<init>(Lwjgl3Application.java:111)
	... 1 more


```

If this isn't an obvious bug it may be a call to harden the code _somewhere_ to catch this better and either log more informative details or even skip over something small if it is like a single turret that fails to place (still logging an error, of course

Other stuff:

- [ ] Eventually make a compatibility PR for the `warp` module and any other modules the GSOC work may break - which I suspect is very few as code in modules so far is really rare
- [ ] Probably just for making an issue to process later, but: are starlanes being placed excessively, to where you can get to many places from many places, rather than have to hop hop hop to get somewhere distant? I got into a _way_ long one to get to an outer system (played with like 6-7 systems). Probably this needs to be balanced by either breaking the starlanes into shorter multi-hop trips or maybe vary the speed based on the length to reach a comparable short-ish duration on any sort of travel.

engine/src/main/java/org/destinationsol/world/generators/PlanetGenerator.java
```diff
@@ -15,25 +15,244 @@
...
+     * This calculates the proper speed for a planet to rotate around its axis. It is randomized slightly, so not all
+     * planets rotate at the same speed.
+     * @return float representing rotation speed
+     */
+    protected float calculateDefaultPlanetRotationSpeed() {
+        return SolMath.arcToAngle(PLANET_GROUND_SPEED, getGroundHeight()) * SolMath.toInt(SolRandom.seededTest(.5f));
+    }
+
+    /**
+     * This method modifies the density of low orbit ships that generate on the Planet. The density represents how
```
Cervator, line 104: I wonder if "density" here is the right word, another possible refactoring that could be bigger than this PR / do sometime later. When I read density I think a desired target state of x ships per y area. But is it really more like a _probability_ to spawn ships that is based on a desired target density? Generally spawn chances are thought of in probabilities, although I can see how a target "coverage" of sorts could be a goal. *If* I'm even reading that right - you could have logic that would go past a target "density" depending on how it was written.

engine/src/main/java/org/destinationsol/world/generators/PlanetGenerator.java
```diff
@@ -15,25 +15,244 @@
...
+            for (ShipConfig shipConfig : getPlanetConfig().groundEnemies) {
+                shipConfig.density *= densityMultiplier;
+            }
+        } else {
+            logger.error("Ship density multiplier must be at least 0");
+        }
+    }
+
+    /**
+     * This method modifies the percentage of the atmosphere at which low orbit ships start spawning.
```
Cervator, line 149: The _percentage_ of the atmosphere? Huh. As in, related to height/distance from the surface of a planet? I guess in theory you could think of an atmosphere as thinning out (thus lower percentage the higher you go), but I wonder if a simply height offset or something would be clearer (I could imagine exotic / magical atmospheres that could be equally thick from surface to top...) . As usual probably a separate refactor.

engine/src/main/java/org/destinationsol/world/generators/SmallPlanetGenerator.java
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+package org.destinationsol.world.generators;
+
+import org.destinationsol.common.SolRandom;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class is a concrete implementation of a PlanetGenerator and handles its creation. This class creates planets
```
Cervator, line 23: "its" creation being relevant to the planet generator or a planet itself? Nitpick :-)

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -204,12 +208,18 @@ private void calculatePlanetPositions() {
                 Orbital orbital = solarSystemOrbitals.get(orbitalPosition);
                 orbital.setFeatureGenerator(generator);
 
+                generator.setAngleInSolarSystem(SolRandom.seededRandomFloat(180));
```
Cervator, line 211: Magic number (descriptive constant for `180` probably would be good?)

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -382,6 +393,21 @@ public void buildFeatureGenerators() {
         }
     }
 
+    /**
+     * This method returns the Planet objects built by the PlanetGenerators of this SolarSystem. It is most useful
+     * after the build method is called
+     * @return List of Planets
+     */
+    ArrayList<Planet> getBuiltPlanets() {
```
Cervator, line 401: "Built" sounds funny to me here. "Generated" ? What exact state are the planets in before, after, and what happens next? Is surface feature placement and other stuff still to do? This is where again a diagram or two would be awesome to show the phases / state changes :-)

engine/src/test/java/org/destinationsol/world/generators/SolarSystemGeneratorTest.java
```diff
@@ -95,7 +104,7 @@ private HullConfigManager setupHullConfigManager(ItemManager itemManager) {
 
     private void setupSolarSystemGenerator() {
         solarSystemGenerator.initializeRandomDefaultFeatureGenerators(1f);
-        solarSystemGenerator.buildFeatureGenerators();
+        //solarSystemGenerator.buildFeatureGenerators();
```
Cervator, line 107: Might be good with a quick TODO to note why this is commented out

Cervator (changes requested):
Egads I have had time to do a semi-thorough review! Probably not really though, it is limited how much focus is available with a newborn half-sleeping behind you. Consider this some starting recommendations rather than necessarily an involved treatise on the architectural applicability of the PR overall - would probably be good with another set of eyes to consider that :-)

I think one thing I run into repeatedly is separating your new/moved code with legacy code. We could probably talk about refactoring naming and such all day long yet that's at least somewhat beyond the scope of your GSOC. So wherever I'm poking at a naming issue feel free to skip if it is preexisting _or_ better yet add some todos or issues so somebody else can return and address it later at some point :-)

Cervator (approved):
Good changes! Was able to test a bit more without crashing (but not without dying! Woo)

PR #603, opened 2021-08-04
Maze Generators and Belts Generators

IsaiahBlanks:
# Description
This PR adds Belts and Mazes using Generator classes. It brings the GSoC 2021 Destination: Sol project up to have all the features of develop, but now modular! It includes tests for the BeltGenerator and MazeGenerator classes. 

# Testing
Load up the game, especially with 4 or more Solar Systems. You'll see belts and mazes througout the Systems. There are two types of belts and two types of mazes to see. They should generate properly.

# Notes
There is still more to be done on the GSoC 2021 project, including writing a tutorial for doing mods and a few other things.


engine/src/main/java/org/destinationsol/world/generators/DenseBeltGenerator.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DenseBeltGenerator extends BeltGenerator {
+    private static final Logger logger = LoggerFactory.getLogger(DenseBeltGenerator.class);
+
+    @Override
+    public void build() {
+        logger.info("Building a belt now!");
```
Cervator, line 26: This isn't very informative, especially at info level ;-)

You coould perhaps throw in some sort of details from the belt config? If this gets triggered per belt is there anything unique about each belt we could log?

engine/src/main/java/org/destinationsol/world/generators/LargeSolarSystemGenerator.java
```diff
@@ -26,7 +26,7 @@
     @Override
     public SolarSystem build() {
         getSolarSystemConfigManager().loadDefaultSolarSystemConfigs();
-        setSolarSystemConfig(getSolarSystemConfigManager().getRandomSolarSystemConfig(getSolarSystemNumber() < 1));
+        setSolarSystemConfig(getSolarSystemConfigManager().getRandomSolarSystemConfig(getSolarSystemNumber() > 0));
```
Cervator, line 29: Eh?

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -324,6 +330,7 @@ protected void initializeRandomMazeGenerators() {
                 Class<? extends FeatureGenerator> newFeatureGeneratorType = featureGeneratorTypes.get(index);
                 try {
                     FeatureGenerator newFeatureGenerator = newFeatureGeneratorType.newInstance();
+                    ((MazeGenerator) newFeatureGenerator).setMazeConfigManager(context.get(MazeConfigs.class));
```
Cervator, line 333: The need for this cast (and later similar stuff) seems like a code smell of some sort. Near midnight so my brain is too hazy to describe it better 😅

Cervator (approved):
Did a quick initial pass, seems okay, but this is a _light_ approval 😁 Would be nice to get a quick pass from one of @IsaacLic @NicholasBatesNZ @Adrijaned or somebody else. I figure ultimately the architecture / class tree could be improved a bit better, but I don't think we need to accomplish that all in one go - iterate iterate iterate.

Tests out too, that was one *dense* belt I flew through, I like it!

If nobody else gets a chance to review this deeper we can probably merge it soon'ish even without further approval :-)

Nice work! Hooray nearly complete refactoring!

engine/src/main/java/org/destinationsol/world/generators/LargeSolarSystemGenerator.java
IsaiahBlanks: @Cervator this was the cause of the bug causing the first system to be harder than it should have been. I'll inline an explanation

IsaacLic (commented):
This looks excellent! I didn't _quite_ finish reviewing - I have four files left to look at - but this looks amazing :)

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
IsaacLic: I think this one has to do with the way `Context` works

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
```diff
@@ -349,6 +356,8 @@ protected void initializeRandomBeltGenerators(float beltChance) {
                 try {
                     if (SolRandom.seededTest(beltChance)) {
                         FeatureGenerator newFeatureGenerator = newFeatureGeneratorType.newInstance();
```
IsaacLic, line 358: Why are you declaring it a `FeatureGenerator` instance if it is a `BeltGenerator`? If you're going to cast it, you should probably either do that here or at the beginning of the previous `if` (before the line `Class<? ...`)

IsaacLic (commented):
Looks excellent! One small comment about the casting thing, but otherwise looks great :)

engine/src/main/java/org/destinationsol/world/generators/BeltGenerator.java
```diff
@@ -31,5 +98,41 @@ public float getDistanceFromCenterOfSolarSystem() {
...
+
+    public void setBeltConfig(BeltConfig beltConfig) {
+        this.beltConfig = beltConfig;
+    }
+
+    public BeltConfig getBeltConfig() {
+        return beltConfig;
+    }
+
+    public SystemBelt getSystemBelt() {
```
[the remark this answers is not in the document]
IsaiahBlanks: You won't be setting the SystemBelt object, you will just be calling instantiateSystemBelt() at the end of your generator's build() method

engine/src/main/java/org/destinationsol/world/generators/MazeGeneratorImpl.java
```diff
@@ -15,21 +15,28 @@
  */
 package org.destinationsol.world.generators;
 
+import com.badlogic.gdx.math.Vector2;
+import com.google.gson.internal.bind.util.ISO8601Utils;
```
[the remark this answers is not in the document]
IsaiahBlanks: I have no idea. It doesn't need to be there, I'll delete it

engine/src/main/java/org/destinationsol/world/generators/SolarSystemGenerator.java
IsaiahBlanks: Good point, I'll address that.

engine/src/main/java/org/destinationsol/game/planet/SystemsBuilder.java
```diff
@@ -169,11 +169,6 @@ private SolarSystem createSystem(List<Float> groundHeights, Vector2 systemPositi
                 Planet planet = createPlanet(planetDist, system, groundHeight, planetConfig, names);
                 planets.add(planet);
                 system.getPlanets().add(planet);
-            } else {
```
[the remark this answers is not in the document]
IsaiahBlanks: It creates a maze object if object should be a maze (the legacy code marked mazes with a negative ground height). This whole method is no longer accessed, maybe even the whole class

PR #588, opened 2021-03-03
Partially adapted MainGameScreen to use NUI

BenjaminAmos:
# Description
This pull request mostly serves as an example of how to port existing UI screens to use NUI. It partially ports one of the more significant in-game UI screens, the `MainGameScreen`. This screen was responsible for rendering the in-game HUD, warn-drawers and overlays. This pull request also includes a port of the `ShipKbControl` screen in its entirety, which is used for the touchscreen controls.

The following elements of the HUD have been ported to use NUI:
 - Side-menu Options
 - Touchscreen Controls
 
The following elements are still using the pre-existing UI system:
 - Player status bars (heath, shields, bullets remaining etc.)
 - Warn drawers (messages that show things like "Enemy Nearby")
 - Overlay screens (these use the pre-existing system, so they by-nature cannot be ported directly to NUI)

The code here mostly works, however some features are broken due to existing issues in NUI's LibGDX input back-end. Some of these issues could be resolved by porting MovingBlocks/TeraNUI#24 to NUI 2.0 but others will require the [`org.terasology.nui.backends.libgdx.NUIInputProcessor`](https://github.com/MovingBlocks/TeraNUI/blob/7e9c5d5472978d6f89c24af1dd55630b3f3a34b0/nui-libgdx/src/main/java/org/terasology/nui/backends/libgdx/NUIInputProcessor.java) class to be re-written.

# Testing
- Run the game and either start a new game or continue an existing one
#### Menu button
- Click on the `Menu` button and the pause menu should be shown
- Close the pause menu with the `escape` key
#### Map button
- Click on the `Map` button and the map screen should be shown
- Close the map screen with the `escape` key
#### Items button
- Click on the `Items` button and a list of the player's current inventory items should be shown
- Close the items screen
#### Talk button
- Approach a station, if not already near one
- Click on the `Talk` button and the talk screen should be shown
- Close the talk screen
#### Mercs button
- Click on the `Mercs` button and the mercenaries screen should be shown
#### Tutorial
- Exit to the main menu and follow the tutorial
#### Changing key bindings
- Due to the previously-mentioned input issues, to test this fully you might have to change the key bindings for certain actions. Both the space and tab keys appear to be fairly unreliable at the moment.
 - Run the game and enter the `Options->Controls` menu
 - Change the `Map` key to `Q`
 - Change the `Shoot` key to `Left-Alt`
#### Menu button
- Press the `escape` key and the pause menu should be shown
- Close the pause menu using the `Resume` button
#### Map button
[...]

PR #605, opened 2021-09-05
NUI: Add new specialised KeyActivatedButton and UIWarnButton widgets to mimic existing UI behaviour

BenjaminAmos:
# Description
This pull request adds the `KeyActivatedButton` and `UIWarnButton` widgets to mimic the UI behaviour of the pre-existing `SolUiControl` buttons.
- A `KeyActivatedButton` widget will activate when a key is pressed. If the key is held, then it will act the same as if a left-mouse button was holding it down.
- A `UIWarnButton` can flash an orange colour to notify the user that an event has occured, such as when new items are present in the inventory. It is also used in the tutorial to indicate which buttons to click on.

The pull request also adds a separate button disabled texture (previously missing) and moves the MigTestScreen into a seperate test screens package.
image

# Testing
- Start a new game, or continue an existing one
- Open the console with the \` key or `F1`
- Use the command `showNUIScreen engine:solButtonsTestScreen`
- Press the keys shown on the buttons to press the corresponding buttons
- Hold the keys shown on the buttons to hold the corresponding buttons down
- The `I` button should be flashing orange

# Notes
- This pull request depends on MovingBlocks/TeraNUI#63 for testing.
- These changes were included in #588 but are being submitted separately so that work can start on porting other UI screens to NUI


Cervator:
With #588 merged this is partially obsolete, but the additional test screen is new and the old one moved. @BenjaminAmos want to redo that setup real quick or apply anything else left from this PR? No rush of course :-)

BenjaminAmos:
The new test screens should probably go in a separate module and all the other changes in this PR merged as part of #588, so I'm going to close this for now.

PR #607, opened 2021-09-12
GSOC 2021 project by IsaiahBlanks - improved galaxy generation

Cervator:
See https://destinationsol.org/2021/08/20/gsoc-project-summary.html

This has been tested fairly thoroughly by my understanding. Not sure if my Android setup is quite right, I crash over a natives issue trying to run on my physical device, yet can get to the main menu just fine on an emulated Nexus 5. Then trying to enter the game world leaves me with a black screen and no useful logging. Hopefully that's just me and doesn't happen to anybody else?

If we can get that cleared I think we can merge this already. Making PR from the topic branch for final validation, commit checks, updates, etc.

Edit: Via the emulated Nexus 5 the game _does_ work via `develop` so it may be something we need to address


BenjaminAmos (commented):
I've pointed-out all the uses of `Class#getSuperclass` that I could find in these changes, as I think it could be better replaced with `Class#isAssignableFrom` personally. All of the suggestions here are untested though. I've also added some minor Android compatibility remarks.

Cervator:
I tried to sort out LGTM (see #609) but it didn't work and I imagine that service is effectively abandoned anyway since they were acquired by GitHub/Microsoft with nothing obvious coming out of that process. So I just disabled it outright for this repo, to avoid bad checks coming from there. Updated this branch from develop so it is current and will trigger a recheck

Cervator:
GitHub won't let me approve this since I authored it even though it came from @IsaiahBlanks - so I can just say that I approve, and have tested now including on Android after grabbing latest from his branch :-)

We've done tons of past review and testing of this so it is time to make it official!

PR #608, opened 2021-09-12
Fixed crashing of game when minimised

ujjman:
# Description
>This PR will fix the crashing of DestinationSol which crashes while  minimizing the application in main menu. 

# Testing
> To test the PR, just minimize the DestinationSol window and the application will not crash

# Notes
> Include any extra notes here

- [ ] Code has been scanned with [SonarLint](http://www.sonarlint.org/intellij/)
- [ ] There are no errors present in the project
- [ ] Code has been formatted and indented
- [ ] Methods have appropriate Javadoc ([How to write Javadoc](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html))

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,4 +78,3 @@ engine/src/main/resources/entity_store.dat
 
 # Ignore donwloaded JREs
 jre/**
-Log.log.txt
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,4 @@ engine/src/main/resources/entity_store.dat
 
 # Ignore donwloaded JREs
 jre/**
+Log.log.txt
--- a/engine/src/main/java/org/destinationsol/SolApplication.java
+++ b/engine/src/main/java/org/destinationsol/SolApplication.java
@@ -176,6 +176,7 @@ public class SolApplication implements ApplicationListener {
     @SuppressWarnings("checkstyle:ParameterAssignment")
     @Override
     public void resize(int newWidth, int newHeight) {
+        //To prevent application crashing, dont resize the height and width to 0 s, this condition checks it
         if(newWidth==0 && newHeight==0){
             newWidth=1366;
             newHeight=768;
--- a/engine/src/main/java/org/destinationsol/SolApplication.java
+++ b/engine/src/main/java/org/destinationsol/SolApplication.java
@@ -176,7 +176,7 @@ public class SolApplication implements ApplicationListener {
     @Override
     public void resize(int newWidth, int newHeight) {
         //To prevent application crashing, dont resize the height and width to 0 s, this condition checks it
-        if(newWidth==0 && newHeight==0){
+        if (newWidth == 0 && newHeight == 0) {
             return;
         }
         displayDimensions.set(newWidth, newHeight);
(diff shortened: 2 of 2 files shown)


engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -173,8 +173,14 @@ public void create() {
         nuiManager = new NUIManager(this, context, commonDrawer, options);
     }
 
+    @SuppressWarnings("checkstyle:ParameterAssignment")
```
BenjaminAmos, line 176: Warnings are normally there to help you. Don't suppress them unless you have a really good reason to.

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -173,8 +173,14 @@ public void create() {
...
     @Override
     public void resize(int newWidth, int newHeight) {
+        //To prevent application crashing, dont resize the height and width to 0 s, this condition checks it
+        if(newWidth==0 && newHeight==0){
+            newWidth=1366;
+            newHeight=768;
+        }
```
BenjaminAmos, line 183: What if the user's window size is not 1366x768? Maybe just don't set it to 0 at all?
```suggestion
        if (newWidth == 0 && newHeight == 0) {
            return;
        }
```

BenjaminAmos (commented):
A couple of comments.

ujjman (commented):
yeah actually you are right I made changes as suggested by you

.gitignore
```diff
@@ -78,3 +78,4 @@ engine/src/main/resources/entity_store.dat
 
 # Ignore donwloaded JREs
 jre/**
+Log.log.txt
```
Cervator, line 81: Huh, this is an odd one. Where did this file come from, or did this ignore entry get added by accident?

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -175,6 +175,10 @@ public void create() {
 
     @Override
     public void resize(int newWidth, int newHeight) {
+        //To prevent application crashing, dont resize the height and width to 0 s, this condition checks it
+        if(newWidth==0 && newHeight==0){
```
Cervator, line 179: Please be sure you have the Checkstyle plugin enabled in your IDE and look for formatting warnings - this isn't formatted right when it comes to whitespace :-)

Cervator (changes requested):
Played around with some minimizing and it seems like no crashing anymore, thanks! :-)

I left a couple small requests that should be trivial to update then we can merge.

One question: did you use the GitHub website to edit any? Two of your commits have the placeholder commit message that'll result from that. Website will also skip style checks and such, so it is easy to introduce issues. Best to use a local IDE and make sure all is clear :-)

engine/src/main/java/org/destinationsol/SolApplication.java
ujjman: oh actually I just ignored the checkstyle errors but now it seems to be okay

.gitignore
ujjman: I removed the file name now it will be fine

Cervator (approved):

Cervator:
Thanks for final tweaks - merging! Using squash merge to cut down the commits a bit since they're generic messages and this is a pretty tiny change (which makes a big difference!)

PR #609, opened 2021-09-12
Attempt to exclude LGTM processing of now-deleted embedded JRE files (and update from develop)

Cervator:
The GSOC branch for some reason trips up LGTM into thinking Javascript files have changed, which it then promptly fails to build, as we don't have any. Once upon a time we did, when we had LWJGL/JRE files embedded in the repo, but those have long since been deleted. Seemingly every PR since just thought the JS hadn't changed and thus didn't try to rerun it and left the check green. The GSOC one retriggered the JS build earning it a red x - this attempts to entirely get rid of JS as a considered language by excluding the directory they used to live in. Lets see what happens!

This also updates the gsoc branch from develop as I'm missing the handy little button that lets us do that via the UI - not sure if maybe that's only available if branch protection is on in some fashion but not really worried about figuring that out right now.


Cervator:
Didn't work and LGTM is probably abandoned anyway. Just discontinued the app for this repo via GitHub.

Issue #616, opened 2021-09-26
Minor UI quirks / bugs


Cervator:
Written up as part of testing #588:

* Mouse clicking on the in-game right side menu seems to have input consumption quirks - screens may open then immediately re-close and/or I can provoke that behavior deeper in screens if I leave a "bad click" live - just once then everything is fine
* ESC closes the main menu and exits the game. Huh. That seems kinda severe
* The "Shoot" binding doesn't show up if you are in mouse/keyboard mode. I figure it might be better if we do show them but keep them locked/disabled if we want to highlight that they're not changeable for the current mode

First one is a bug for sure and not new. Annoying, easy workaround, *probably* a somewhat straight forward bug to hunt down

Second one is subjective, but I think we should _not_ close the game on ESC in the main menu? It should probably just close screens within the game, but not close the game entirely? Happy to hear other opinions here

Third one might be bigger - the control bindings are essentially completely separate per input mode (keyboard, keyboard/mouse, and game controller I figure). Maybe we should just make that more obvious by putting a label on each variant that points out "Note: Different input modes have their own bindings scheme with different functions available for remapping and do not affect each other" ?

Might feel a bit odd that a "global" function like opening the map could differ per mode, and/or that changing it in one mode does not affect the others. This might be another subjective call though - probably just pointing out that the modes are independent would be good enough. I could also just be getting old and exhausted from raising tiny humans.

PR #621, opened 2021-10-01
Escape key on the main menu wont exit the app

Mystic-Slice:
# Description
> I agree with @Cervator on the 2nd point in issue #616 (about the Esc key closing the app itself). It seems very abrupt. I do not think the Esc key should close the app. But if it is needed, then there should be at least a confirmation dialog before quitting the app.

# Testing
> Open the app and try pressing the Esc key. The app won't close.


Cervator (approved):
Yay less abrupt :-)

PR #622, opened 2021-10-01
feat: implementation of gestalt-DI

pollend:
here is some experimental work with: https://github.com/MovingBlocks/gestalt/pull/87\

This is an implementation of gestalt DI. At the moment there is a ApplicationContext --> GameContext. Application context will contains all top level dependencies such as the audio system and the game context is the actual gameplay systems. This could be managed and split into different layers depending on different states of the application. each module has its own context so

This is how the current import scheme works with Gestalt modules so each module is only dependent on a child module and so forth. there is no way that moduel2 has a cyclic dependency on Module1

```
Application --> GameContext --> Module1 --> Module2 --> ... --> ModuleN  
```

This is a straightforward replacement of the context system that was pulled in from Terasology. Each context defines the services that are needed and at request the object is built up from the dependencies within that context and the parent context if it fails to get resolved in the current context. anything that used to be dependent on pulling from the context will declare the dependencies that are needed to build that object and the context will resolve building that object.

Unit tests are a matter of providing a serviceRegistry with the needed dependencies and the injection system will resolve those as need be. it should be easy enough to construct a mock object within the context and the system should handle the rest. reduces the burden of working backwards from the source to construct a test environment.   

This handles a lot of the clear problems with the old scheme.

- dependencies are added in at some arbitrary point so its hard to determine what is and isn't available
- constructing a test environment is very reliant on understanding the couplings within the system   so its necessary to work out dependencies to construct a working environment.

A lot of the code that is used to construct the game is gone and and has been handed over the DI system.

This also does away with reflections. this is also managed by gestalt-DI. I think @DarkWeird would have more information about that portion of the implementation. basically lets us do some scanning for a given type without paying the cost of walking the class path. accomplishes a similar goal to reflections bug handled through the annotation processor.  

need changes from this PR: https://github.com/MovingBlocks/TeraNUI/pull/65

this PR replaces this PR: https://github.com/MovingBlocks/DestinationSol/pull/586


engine/src/main/java/org/destinationsol/game/screens/ShipMixedControl.java
```diff
@@ -63,18 +63,18 @@
 
     @Override
     public void update(SolApplication solApplication, boolean enabled) {
-        Context context = solApplication.getContext();
+//        Context context = solApplication.getContext();
```
BenjaminAmos, line 66: If this line isn't used anymore, can you remove it?

engine/src/main/java/org/destinationsol/game/screens/MapScreen.java
```diff
@@ -78,7 +78,7 @@
 
     @Override
     public void updateCustom(SolApplication solApplication, SolInputManager.InputPointer[] inputPointers, boolean clickedOutside) {
-        Context context = solApplication.getContext();
+//        Context context = solApplication.getContext();
```
BenjaminAmos, line 81: If this line isn't used anymore, can you remove it?

engine/src/main/java/org/destinationsol/game/screens/MapScreen.java
```diff
@@ -118,15 +118,15 @@ public void updateCustom(SolApplication solApplication, SolInputManager.InputPoi
...
+            float rotateAngle = game.getCam().getAngle();
             mapDrawer.getMapDrawPositionAdditive().add(im.getDrag().scl(scrollFactor).rotate(rotateAngle));
         }
 
         if (isPickingWaypointSpot) {
             if (inputPointers[0].isJustUnPressed() && !addWaypointControl.isJustOff()) {
-                SolCam camera = context.get(SolCam.class);
-                float camAngle = camera.getAngle();
-                Vector2 mapCamPos = camera.getPosition().add(mapDrawer.getMapDrawPositionAdditive());
+//                SolCam camera = context.get(SolCam.class);
```
BenjaminAmos, line 127: If this line isn't used anymore, can you remove it?

engine/src/main/java/org/destinationsol/game/screens/ShipMouseControl.java
```diff
@@ -44,13 +44,13 @@
 
     @Override
     public void update(SolApplication solApplication, boolean enabled) {
-        Context context = solApplication.getContext();
+//        Context context = solApplication.getContext();
```
BenjaminAmos, line 47: If this line isn't used anymore, can you remove it?

engine/src/main/java/org/destinationsol/ui/nui/screens/MainGameScreen.java
```diff
@@ -55,8 +54,16 @@
...
-    @In
-    private SolApplication solApplication;
+    @Inject
+    protected SolApplication solApplication;
+
+    public MainGameScreen() {
+
+    }
+
+    public void setSolApplication(SolApplication application) {
```
BenjaminAmos, line 64: Are NUI screens injected by `gestalt-di` yet? If so, then why is this method needed?

engine/src/main/java/org/destinationsol/ui/TutorialManager.java
```diff
@@ -61,9 +57,15 @@ public TutorialManager(GameScreens screens, org.destinationsol.ui.nui.screens.Ma
         stepIndex = 0;
     }
 
+    public void init() {
+
+    }
```
BenjaminAmos, line 62: Is this method needed anymore?

engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -126,113 +161,49 @@
...
+        this.solApplication = solApplication;
         boolean isMobile = solApplication.isMobile();
+
         if (!isMobile) {
             mainGameScreen = (MainGameScreen) Assets.getAssetHelper().get(new ResourceUrn(NUI_MAIN_GAME_SCREEN_DESKTOP_URI), UIElement.class).get().getRootWidget();
         } else {
             mainGameScreen = (MainGameScreen) Assets.getAssetHelper().get(new ResourceUrn(NUI_MAIN_GAME_SCREEN_MOBILE_URI), UIElement.class).get().getRootWidget();
         }
+        // TODO: resolve hack to set solApplication
+        mainGameScreen.setSolApplication(solApplication);
```
BenjaminAmos, line 184: If you're still planning to resolve this, then you can start by looking at the injection code in `NuiManager#pushScreen`.

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -91,28 +87,34 @@
...
-    private SolLayouts layouts;
     private GameOptions options;
-    private CommonDrawer commonDrawer;
     private String fatalErrorMsg;
     private String fatalErrorTrace;
     private SolGame solGame;
     private ParameterAdapterManager parameterAdapterManager;
     private NUIManager nuiManager;
-    private Context context;
+//    private Context context;
```
BenjaminAmos, line 116: If this isn't used anymore, can it be removed?

engine/src/main/java/org/destinationsol/SolApplication.java
```diff
@@ -121,57 +123,62 @@
...
+    private BeanContext appContext;
+    private BeanContext gameContext;
+
     // TODO: Make this non-static.
     private static Set<ResizeSubscriber> resizeSubscribers;
 
-    public SolApplication(ModuleManager moduleManager, float targetFPS) {
+    @Inject
+    protected SolApplication() {
+        throw  new RuntimeException("Unimplemented");
```
BenjaminAmos, line 134: Maybe make the error message here a bit more helpful? Perhaps something like `Cannot instantiate SolApplication via DI`?

BenjaminAmos (changes requested):
I think that this is a great step towards modularising the codebase further and removing dependencies on the monolithic `SolApplication` and `SolGame` classes. The changes made are fairly clear, although the whole `gestalt-di` system is going to need a lot more explicit documentation at some point.

Since this is a major change, like all other major changes, it broke the Android build. I did have some patches that fixed it again for an earlier iteration of this code but I've overhauled the Android asset loading since, so it may take some time before I can fix it again.

There are some API access exceptions thrown when trying to use code modules with these changes. Code modules are a very under-used feature of Destination Sol right now, so it is understandable that they were overlooked. In future, maybe try testing with my `codeModuleTest` module in the workspace, which you can obtain using `groovyw module get codeTestModule -remote BenjaminAmos`? I haven't even tried `warp` with this yet, which is a more complex code  module, although you could try that one too.

With these changes, the golden asteroid is gone! Whilst it was only ever a temporary thing to test the ECS systems, the code to spawn it can still be found in `SolApplication#draw`. This implies that ECS code no longer works with these changes. When you start playing, a golden asteroid should (at least for now) appear in your spawn area, where you can shoot it to split it into many tiny pieces of debris.

engine/src/main/java/org/destinationsol/ui/nui/screens/MainGameScreen.java
pollend: there is a problem here so I kind of have a hacky workaround.

engine/src/main/java/org/destinationsol/game/SolGame.java
pollend: hmm, how would this work in practice, if we push a screen it should be in the context it was instantiated in :?.

BenjaminAmos: The existing behaviour is that the screen is re-injected when pushed onto the UI stack (displayed). It should probably use the currently active context (e.g. `appContext` for the main menu and `gameContext` for in-game).
https://github.com/MovingBlocks/DestinationSol/blob/30700e93270716c5e57fff16784f87153a97b9f3/engine/src/main/java/org/destinationsol/ui/nui/NUIManager.java#L287-L295
(I think that replacing `InjectionHelper#inject` with `BeanContext#inject` would probably be sufficient)

#631 re-works the logic in `NUIManager` to inject the screens upon creation, rather than upon display, so then they would then inherit the context they were instantiated in. For now though, the existing behaviour should be preserved.

pollend:
I guess we need these changes for the build to work: https://github.com/MovingBlocks/DestSolAndroid

engine/src/main/java/org/destinationsol/entitysystem/ComponentSystemManager.java
```diff
@@ -17,25 +17,21 @@
...
-                InjectionHelper.inject(system, context);
-                componentSystems.add(system);
-            } catch (Exception e) {
-                e.printStackTrace();
-            }
-        }
+    @Inject
+    public ComponentSystemManager(List<ComponentSystem> componentSystems) {
+      this.componentSystems = componentSystems;
     }
```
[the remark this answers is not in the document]
pollend: its because its under CoreService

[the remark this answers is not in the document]
pollend: you have to pull it from ModuleManager if you have some registries bound to GameContext. then they will be missed. 

AppContext --> GameContext 
       \--> ModuleManager --> ModuleContext1 --> ModuleContext2 --> etc ...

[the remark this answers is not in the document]
pollend: I have a workaround to do this pre begin setup. we should probably instantiate the ModuleManager when the gameContext is created?

DarkWeird:
Needs https://github.com/MovingBlocks/TeraNUI/pull/70/ to work.

* restore playable state for desktop
* Create BeanClassFactory for injecting to Asset related classes
* Provide @Inject for every Asset related class
* using scanner for scan `assets` and `ui` packages (im too lazy to explicitly write all them in service)

build.gradle
```diff
@@ -39,6 +39,7 @@ repositories {
 
     // Good ole Maven central
     mavenCentral()
+    mavenLocal()
```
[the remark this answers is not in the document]
DarkWeird:
```suggestion
```

BenjaminAmos (commented):
It's great to have this back in a mergable state again! Some parts are still lacking explainatary javadoc but it's not worth holding this up any longer over that.

I've left a few very minor comments. Other than that, these changes are an overall improvement. Let's hope the gains from this were well worth the vast effort involved to get there.

engine/build.gradle
```diff
@@ -2,53 +2,46 @@ apply from: '../config/gradle/common.gradle'
...
+    api "org.terasology.gestalt:gestalt-asset-core:$gestaltVersion"
+    api "org.terasology.gestalt:gestalt-entity-system:$gestaltVersion"
+    api "org.terasology.gestalt:gestalt-module:$gestaltVersion"
+    api "org.terasology.gestalt:gestalt-util:$gestaltVersion"
+
+    api "org.terasology.gestalt:gestalt-di:$gestaltVersion"
+    api "org.terasology.gestalt:gestalt-inject:$gestaltVersion"
+    annotationProcessor "org.terasology.gestalt:gestalt-inject-java:$gestaltVersion"
+
+    implementation "net.jcip:jcip-annotations:1.0"
```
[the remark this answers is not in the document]
BenjaminAmos: It seems like the annotations here are for something different. It's probably a harmless dependency.
https://jcip.net/annotations/doc/index.html

engine/src/main/java/org/destinationsol/game/SolGame.java
```diff
@@ -458,7 +426,7 @@ public void drawDebug(GameDrawer drawer, Context context) {
 
     private void drawDebugPoint(GameDrawer drawer, Vector2 dp, Color col) {
         if (dp.x != 0 || dp.y != 0) {
-            float sz = context.get(SolCam.class).getRealLineWidth() * 5;
+            float sz = getContext().get(SolCam.class).getRealLineWidth() * 5;
```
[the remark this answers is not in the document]
BenjaminAmos: This is debug code, so it isn't running most of the time. Since this just replaces the existing code with a functional replacement, nothing needs to change here.

engine/src/main/java/org/destinationsol/assets/fonts/ScaledFontProducer.java
```diff
@@ -39,14 +40,15 @@ public class ScaledFontProducer implements AssetDataProducer<FontData> {
...
     /**
      * Creates a new {@link ScaledFontProducer} to produce scaled fonts assets depending on the ResourceUrn provided.
-     * @param assetManager the asset manager to use when obtaining base font data
+     * @param assetHelper the asset manager to use when obtaining base font data
      */
-    public ScaledFontProducer(AssetManager assetManager) {
-        this.assetManager = assetManager;
+    @Inject
+    public ScaledFontProducer(AssetHelper assetHelper) {
+        this.assetHelper = assetHelper; // TODO provide AssetManager
```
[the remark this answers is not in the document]
BenjaminAmos: This isn't important. `ScaledFontProducer` still works the same as before, so nothing needs to change really.

engine/src/main/java/org/destinationsol/BeanClassFactory.java
```diff
@@ -0,0 +1,20 @@
+package org.destinationsol;
+
+import org.terasology.gestalt.di.BeanContext;
+import org.terasology.gestalt.util.reflection.ClassFactory;
+
+import javax.inject.Provider;
+import java.util.Optional;
+
+public class BeanClassFactory implements ClassFactory {
```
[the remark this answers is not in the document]
BenjaminAmos: On further investigation, `BeanClassFactory` is used by gestalt to instantiate asset format classes. It does essentially what the name implies. I don't see any need to document this futher right now, given the simplicity.

engine/build.gradle
```diff
@@ -73,30 +66,19 @@ dependencies {
...
-    doLast {
-        // Without the .mkdirs() we might hit a scenario where the classes dir doesn't exist yet
-        Reflections reflections = new Reflections(new ConfigurationBuilder()
-                .addUrls(inputs.files.collect { it.toURI().toURL() })
-                .filterInputsBy({ it.startsWith("org.destinationsol") || it.startsWith("org.terasology.nui") })
-                .setScanners(new TypeAnnotationsScanner(), new SubTypesScanner()))
-        reflections.save("$buildDir/resources/main/org/destinationsol/reflections.cache")
-    }
+// Adds Resources as parameter for AnnotationProcessor (gather ResourceIndex,
+// also add resource as input for compilejava, for re-gathering ResourceIndex, when resource was changed.
```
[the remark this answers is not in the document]
BenjaminAmos: I just checked and you can still just drop-in modules after the game's been built. Some comments are better than none and this one is just as valuable as any other.

BenjaminAmos (approved):
The game runs successfully on both desktop and android and the build passes.

PR #624, opened 2021-10-09
De-duplicate assets on Android

BenjaminAmos:
# Description
This pull request contains the corresponding engine changes for the Android pull request MovingBlocks/DestSolAndroid#22. See that pull request for a description of the build system changes.

The engine changes are as follows:
 - Changed the output path for the engine module reflections cache to make it easier to symlink
 - Reduced the classes included in reflection caches in general, leading to an overall module size reduction
 - Updated gestalt to the latest release.
 - Updated `AndroidOggMusicFileFormat` and `AndroidOggSoundFormat` to use a new method for reading files that eliminates the need to copy them onto the filesystem on Android. This method requires the music and sound files to be present in a directory module directly in the Android `assets` directory instead of in a jar, which the Android build scripts 

# Testing
- Android should still work the same as before with these changes. The APK file sizes should be considerably smaller though.
- As a comparison, the v2.0.0 APK (with duplicated assets) is approximately 121MB. This is for the base game only, with no additional modules.
- With these changes, the APK size is approximately 107MB (base game only).

# Notes
- This depends on MovingBlocks/DestSolAndroid#22
- Breaking changes in the latest gestalt (7.x) release mean that this will likely break existing ECS pull requests. The `Component#copy` method has been renamed to `Component#copyFrom`.
- With these changes, it is no longer possible to use archive modules (jar/zip modules) on Android. For a from-source workspace though, there should be no functional difference.

### DIFF
--- a/engine/build.gradle
+++ b/engine/build.gradle
@@ -29,7 +29,7 @@ buildscript {
         // HACK: Needed for NUI and gestalt entity-component reflections
         classpath group: 'org.terasology.nui', name: 'nui', version: '3.1.0-SNAPSHOT'
         classpath group: 'javax.servlet', name: 'javax.servlet-api', version: '3.0.1'
-        classpath group: 'org.terasology.gestalt', name: 'gestalt-entity-system', version: '7.0.6-SNAPSHOT'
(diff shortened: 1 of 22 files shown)


Cervator (approved):
Tested out, alongside actually building Android in Jenkins!

PR #629, opened 2021-10-16
fix: fix Android textures bug

### LINKED ISSUE #558: Textures load improperly (Android)
Issue: Textures are completely shuffled when exiting the game with the back button. Often you take on the appearance of an enemy ship that was nearby when you exited, and I was saw an item sprite be replaced with the desert planet turret's "edge of screen" icon. 

Repro:
1. On android, exit the game using the back button.
2. Reopen the game and reload your save (untested with new game)
3. Look at yourself, or nearby ships, or asteroids, or about 20% of things.

V2.0.0

LGK20, Android

BenjaminAmos:
# Description
When the game is closed using the back button on Android, the activity containing it is disposed. The runtime VM and all static variables are not disposed though, so lingering references remained in the SpriteManager cache to textures that didn't exist anymore. The SpriteManager cache is now cleared when the game's activity is destroyed.

I also added some further code to properly dispose of gestalt-managed textures, since the reason why textures appeared to be randomly swapped was because the underlying sprites referenced invalid OpenGL handles that had not been cleaned-up properly. By correctly disposing of these textures upon activity destruction, the invalid handles now show black rectangles instead.

A bit more context on the Android activity lifecycle (which is exposed by libGDX through the `SolApplication#pause` and `SolApplication#destroy` methods) can be found [here](https://developer.android.com/guide/components/activities/activity-lifecycle#onpause).

# Testing
Follow the testing instructions found in #558. The game should now appear normally, rather than with randomly-swapped textures.

# Notes
- This fixes #558
- This also fixes MovingBlocks/DestSolAndroid#9
- This also fixes a long-standing bug that I don't think there is an issue for, where all `SolUiControl` buttons show visual artifacts after closing the game using the back button.

### DIFF
--- a/engine/src/main/java/org/destinationsol/SolApplication.java
+++ b/engine/src/main/java/org/destinationsol/SolApplication.java
@@ -38,6 +38,7 @@ import org.destinationsol.game.SolGame;
 import org.destinationsol.game.SolNames;
 import org.destinationsol.game.WorldConfig;
 import org.destinationsol.game.drawables.DrawableLevel;
+import org.destinationsol.game.drawables.SpriteManager;
 import org.destinationsol.health.components.Health;
 import org.destinationsol.location.components.Angle;
 import org.destinationsol.location.components.Velocity;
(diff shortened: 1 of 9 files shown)


Cervator (approved):
Tested out, alongside actually building Android in Jenkins! 

I was able to launch the game on my physical Android device twice in a row without noticing any problems - although I'm mildly uncertain whether that "counts" since the second launch was from the icon left on my phone (can't remember if I previously had DS installed - could the two launches be different? Or maybe that was the install caused by running it from source)

PR #631, opened 2021-10-22
Adapted Main Menu screens to use NUI

BenjaminAmos:
# Description
This pull request replaces the pre-existing menu screens with NUI equivalents. All of the screens comprising the main menu, except for `InputMapScreen` have been ported. Whilst doing this, I found that some changes to NUIManager were needed to correctly maintain screen state. I also took the opportunity to improve UI asset loading slightly, as I found that it was duplicating quite a bit of that code previously. The `UIFont` class has been removed, which is a good thing since it was only ever intended to be a temporary fix for font scaling. Font scaling was implemented in a better way with the NUI Main Game Screen changes, so I have moved the rest of the code to use the scaled font producer instead.

~~I have not managed to replace `InputMapScreen` yet, as it is a more complex layout and my grasp of NUI UI layouts is still fairly immature. I am working on it though.~~ *`InputMapScreen` has now been implemented.*

## API Changes
 - `NUIScreenLayer#initialise` is now only called once when the UI asset is loaded by `NUIManager#createScreen`.
 - Added `NUIScreenLayer#onAdd`, which is called whenever the screen is added to the UI stack.
 - Added `Assets#getUIElement` to obtain a `UIElement` asset more easily.
 - Added `NUIManager#setScreen` for replacing the entire UI stack with a single screen.
 - `InputManager#setScreen` now accepts a null screen argument to clear all screens. This becomes more needed as we reach the point where only NUI screens are shown at one time.

# Testing
- Use a build of Destination Sol without these changes and start going through all the menu screens. Observe how they behave and appear.
- Then test against a build with these changes. The UI screens may look different but they should be similar and functionally identical.
## Some things to double-check
- Do all the key bindings work the same as before?
- Can you change the screen resolution successfully?
- Do the layouts appear consistent?
- Is there any text going off-screen? (note: the credits text is known to go off-screen with these changes due to its very large height)

# Notes
- This contributes towards https://github.com/MovingBlocks/DestinationSol/issues/455.
- You can see a visual comparison of the UIs before and after these changes [here](https://gist.github.com/BenjaminAmos/706f24eef73f84e3bbb8053b9d6e281b).
- Some elements of the code changes here still need a bit of polishing, so further minor changes may still be made.


engine/src/main/java/org/destinationsol/ui/nui/NUIManager.java
```diff
@@ -186,7 +197,10 @@ public void update(SolApplication solApplication) {
                 focusManager.getFocus().onKeyEvent(event);
             }
 
-            for (NUIScreenLayer uiScreen : uiScreens) {
+            // Create a copy of the list, as screens may be removed in response to key events.
+            // This does not happen for most key events, however it may occur with KeyActivatedButton instances,
+            // which already have special handling in the NUIScreenLayer class.
+            for (NUIScreenLayer uiScreen : new LinkedList<>(uiScreens)) {
```
NicholasBatesNZ, line 203: I'll assume there's a reason to create a new `LinkedList`...

NicholasBatesNZ (approved):
I haven't tested on mobile so I assume everything is working fine there. Bravo really and sorry it took so long to get to

engine/src/main/java/org/destinationsol/ui/nui/NUIManager.java
BenjaminAmos: A new list is created because you can't change the contents of the list whilst iterating over it I thought that the comment above explained that but maybe it isn't clear. If there's a better way of doing this, then I would gladly change that code.

NicholasBatesNZ: I was more wondering why a `LinkedList` specifically. But looking at it now I see that I was just tired missed that `uiScreens` is already a `LinkedList`. 👍

PR #633, opened 2021-10-24
build: add gradle library include builds

BenjaminAmos:
## Description
This pull request adds a convenience shortcut for including external gradle-built libraries from source into the main Destination Sol build. I find it useful for when I need to make and test changes to `gestalt` and `TeraNUI` in particular.

## Usage:
To include a gradle-built library from source, just clone it into the libs sub-directory, e.g. for `gestalt`:
```sh
git clone https://github.com/MovingBlocks/gestalt.git libs\gestalt
```

Any dependencies on that library will then use the source version, rather than fetching the dependency from an external repository.

## Credits
All credit for this goes to the authors of the original Terasology build file at https://github.com/MovingBlocks/Terasology/blob/develop/libs/subprojects.settings.gradle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@ modules/*
 modules/**/build.gradle
 !modules/core
 !modules/subprojects.gradle
+libs/*
+!libs/subprojects.gradle
 
 ## GWT
 war/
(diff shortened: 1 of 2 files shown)


Cervator (approved):
Have tested locally, can confirm 👍 

Poked around a bit more to refresh the Groovy stuff as well so we can fetch libs like in Terasology

PR #634, opened 2021-10-30
feat: upgrade the groovyw system to be closer in sync with Terasology

Cervator:
This is a _very optimistic_ and quick PR that essentially drops in most the latest Groovy utility from Terasology (for instance leaves out facades and the init task) so we can use it to fetch libs (to go with #633) - it also pulls in all other recent related changes, good or bad, hopefully more working than not. In some cases (like `common.groovy`) it is _a lot_ and I honestly haven't even semi-thoroughly eyeballed the diff - just trusting that since it works with TS it'll work with DS

I tried some quick search and replace for basic differences between TS and DS. I've probably still missed bits or erased some DS-specific details. Feel free to poke around and find some of all the things :-) 

Tested with:

* `groovyw lib get Gestalt` (and validating with `gradlew engine:dependencies` to see that source Gestalt is used)
  * **Note:** does not actually compile, am guessing due to a version diff in the default branch you get vs what DS `develop` uses - likely doesn't matter for this PR
* `groovyw module get TutorialGalaxyGeneration` (I already had warp - @BenjaminAmos should we move your other test module into the DS org? I couldn't spot it, but recall it)
* `groovyw module create MyNewModule` (made _some_ asset subdirs, probably not all the right ones)

Then refreshed Gradle a time or two to make sure there weren't any obvious errors.

If digging deeper surely some problems will show up, for instance `groovyw lib list` doesn't work as the GitHub paging API changed - pretty sure that's broken in TS as well. I'm trying to not spend too much effort on this as we have a potentially reworked Groovy utility system that entirely replaces the current setup.


config/groovy/common.groovy
```diff
@@ -408,24 +476,116 @@ class common {
...
+ * Small ANSI coloring utility.
+ * @see https://gist.github.com/tvinke/db4d21dfdbdae49e6f92dcf1ca6120de
+ * @see http://www.bluesock.org/~willg/dev/ansi.html
+ * @see https://gist.github.com/dainkaplan/4651352
+ */
+class Ansi {
+
+    static final String NORMAL         = "\u001B[0m"
+
+    static final String BOLD           = "\u001B[1m"
```
BenjaminAmos, line 562: I've had issues with rendering ANSI escape codes on Windows before but I can't find any issues while testing right now, so this should be alright, hopefully.

BenjaminAmos: This works under the Windows Terminal app but doesn't work under a plain Command Prompt. It just causes cosmetic issues though.

config/groovy/util.groovy
```diff
@@ -203,42 +266,47 @@ private void printListItemsCondensed(String[] items) {
...
+    Example: 'groovyw module create MySpaceShips' - would create that module
+    Example: 'groovyw module get caution - remote vampcat - would retrieve caution module from vampcat's account on github.'
+    Example: 'groovyw module get *' - would retrieve all the modules in the DestinationSol organisation on GitHub.
+    Example: 'groovyw module get ?r*' - would retrieve all the modules in the DestinationSol organisation on GitHub
+        that start with any single character, then an \\"r\\" and then end with anything else.
+        This should retrieve the draconic, organic and tribe repositories from the DestinationSol organisation on GitHub.
+ 
+    *NOTE*: On UNIX platforms (MacOS and Linux), the wildcard arguments must be escaped with single quotes e.g. groovyw module get '*'.
+    
+    *NOTE*: Item names are case sensitive. If you add items then `gradlew idea` or similar may be needed to refresh your IDE
```
BenjaminAmos, line 305: I don't actually use `gradlew idea` at all anymore, since IntelliJ's built-in gradle support just works now. Is this still standard practice for Destination Sol?

config/groovy/common.groovy
```diff
@@ -339,17 +398,26 @@ class common {
...
             new URL(currentPageUrl).openConnection().with { connection ->
                 connection.content.withReader { reader ->
                     slurper.parseText(reader.text).each { item ->
                         mappedPossibleItems.put(item.name, item.description)
+                        //println "Want to get item " + item.name
                     }
                 }
                 currentPageUrl = getLink(connection, "next")
+                // TODO: This comparison is vulnerable to a page request size of "100" or anything that starts with a 1, but just using 99 above ..
+                if (currentPageUrl.contains("page=1")) {
```
BenjaminAmos, line 411: `getLink` can return null if there is no such header present. GitHub doesn't return a `Link` header if the response was not paginated, such as when there are less than 99 repositories in the organisation.
```suggestion
                currentPageUrl = getLink(connection, "next")
                // TODO: This comparison is vulnerable to a page request size of "100" or anything that starts with a 1, but just using 99 above ..
                if (currentPageUrl != null && currentPageUrl.contains("page=1")) {
```

BenjaminAmos (changes requested):
I've tested this and it mostly works, apart from the search functionality which, as you mentioned, doesn't work. This is actually because the DestinationSol organisation doesn't have more than 99 repositories in it. I'll suggest a quick fix here and then I can merge it.

config/groovy/common.groovy
Cervator: There's a whole new system for colors with the PicoCLI approach in the new GooeyCLI system, which works nicely on Windows, so no worries there. Probably lovely _new_ worries in the new system though, when we get there! :-)

config/groovy/util.groovy
Cervator: Nah, pretty sure this is just obsolete and we never got around to removing it. There are still stray references to `gradlew idea` in some of the wikis too, I think.

BenjaminAmos (approved):
I've retested this and listing now works.

PR #636, opened 2021-11-05
NUI inventory screen

BenjaminAmos:
# Description
This pull request converts the in-game inventory screen to use NUI. It should function almost identically to before, however it may look a bit different. Any UI design critiques would be appreciated also.

# Testing
- This pull request affects all screens that use the inventory's functionality.
- The following screens are affected and should be double-checked against an older build to identify possible regressions:
  - `TutorialManager`
  - `BuyItemsScreen`
  - `SellItemsScreen`
  - `ChangeShipScreen`
  - `ChooseMercenaryScreen`
  - `GiveItemsScreen` (give items to a mercenary)
  - `HireShipsScreen`
  - `ShowInventory` (the "Items" screen)
  - `TakeItems` (take items from a mercenary)
- As a guide, at least the following things should be checked:
  - Is the appearence similar to before (or better)?
  - Do all of the key-bindings work as before?
  - Does the screen function the same as before (or better)?

# Notes
- This depends on #631.


BenjaminAmos:
I should have mentioned before that https://github.com/MovingBlocks/TeraNUI/pull/66 and https://github.com/MovingBlocks/TeraNUI/pull/67 contain bug fixes that should help with testing this pull request.

BenjaminAmos:
There is another bug fix for NUI which allows this pull request to render correctly: https://github.com/MovingBlocks/TeraNUI/pull/69.

BenjaminAmos:
This was replaced by #668.

PR #643, opened 2022-01-30
fix: fix respawn crash when the hero is transcendent

### LINKED ISSUE #568: Crash when respawning while in a starport transit animation
Issue: If you enter a starport and then try to respawn, you'll crash the game.

Repro:
1. Enter a starport
2. Respawn

Edit: https://media.discordapp.net/attachments/389093174921920512/784139389675503666/Screenshot_20201203-152827.png

V2.0.0

LGK20, Android

BenjaminAmos:
# Description
This pull request fixes a crash that occurs when the player attempts to respawn whilst trancendent (going through a star port).

# Testing
- Start a new game or continue an existing one.
- Fly to the nearest star port and enter it.
- Whilst travelling between star ports, open the in-game menu and press the `Respawn` button.
- The game should not crash and the player should be respawned successfully.

# Notes
This fixes #568.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/SolGame.java
+++ b/engine/src/main/java/org/destinationsol/game/SolGame.java
@@ -522,7 +522,7 @@ public class SolGame {
         respawnState.setPlayerRespawned(true);
         if (hero.isAlive()) {
             setRespawnState();
-            objectManager.removeObjDelayed(hero.getShip());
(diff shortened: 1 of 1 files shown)


BenjaminAmos:
I do wonder if the player should be allowed to re-spawn at all when travelling between star ports? Regardless of that, this does indeed still appear to fix the crash.

Issue #552, opened 2020-08-24
ECS graphics are misaligned


IsaacLic:
Entity sprites don't line up properly with their collision meshes. [This PR](https://github.com/MovingBlocks/DestinationSol/pull/551) added code to fix that, but it is still slightly off.

### How to reproduce

* In `DestinationSol\engine\src\main\resources\debugOptions.ini`, set `drawPhysicBorders` to be true
* Run the game
* You should see the collision meshes of each object. There should be a golden asteroid visible with an inaccurate collision mesh.

Issue #59, opened 2015-07-28
Ship Stuck in Planet


SimonC4:
I was in the imperialBig ship on the first planet, the rocky planet, and got stuck inside the planet. Kind of sucked in and then the ship disappeared and the camera kept drifting to the middle of the planet. 

Seems like it's happened a few times before: http://steamcommunity.com/app/342980/discussions/1/606068060828393540/

Cervator:
Yup have heard about this before, it just seems really tricky to reproduce and thus troubleshoot further.

Had coincidentally mentioned it to @grauerkoala earlier today in relation to #58 since it is probably some obscure bug like that which can cause it (but not that one)

Cervator:
Another report - suggests (again) maybe the Vanguard (aka imperialBig) in particular can hit this issue. Problem with its hit box?

http://steamcommunity.com/app/342980/discussions/1/520518053428477307/

Cervator:
More details from the last post, with possible reproduction instructions: 

"go to a flat spot on a planet then move your mouse in a clockwise circle around the ship for a minute or two thats how it worked for me"

So maybe you can sort of "wiggle" your way through the surface with a Vanguard.

Viech:
This just happened to me with the Beetle on `1.3.0`, so it's not Vanguard-specific.

Cervator:
Thanks for the extra detail @Viech :-)

Cervator:
And another: http://steamcommunity.com/app/342980/discussions/1/520519112496560688/

Tendency for ships to "jump" slightly / be able to skip physics/collissions briefly? Seems related.

crazywolf132:
What if you changed the rendering of the planet, so the it is no-longer renders as lots of little objects... what if you tried to do something more along the lines of pre-made planets. This would work by joining together the images and make a single .png and then load and scale that into the game.

crazywolf132:
Anyone got an image of this?

OxidizeMonk:
I got this problem too - when the tower on the snow planet shot me with a paralysing weapon, i teleported in the ground, under the tower.

crazywolf132:
Ok thank you @OxidizeMonk.
Out of respect to the developers, here is how to fix it.   @Cervator.
You need to re-do the collision boxes, probably make them 3 pixels bigger than what they need to be.
This will compensate for the code not joining them to pixel perfection.

Cervator:
[...]

PR #644, opened 2022-01-30
fix(mazes): fix incorrect box2d fixtures for maze tiles

### LINKED ISSUE #576: Bad collision boxes for maze

trying to go inside a maze

### What actually happened

noticed weird collision boxes implemented for the maze

### How to reproduce

*go to the maze at the bottom of the system you start in
* try to enter the maze
* notice weird collision boxes

Windows PC
v2.0.1

### Computer details
i3 3220
gtx 1050
8gb ram

BenjaminAmos:
# Description
This pull request fixes incorrect maze collisions by partially-reverting the maze tile changes in https://github.com/MovingBlocks/DestinationSol/commit/e5f6c68524ef8867d06c80c504bb127ccdde643a. This causes the sprite tile textures to be correctly flipped, although it prevent mazes from using animated sprites.

### Before
image
### After
image

# Testing
- In the `engine/src/main/resources/debugOptions.ini` file, set `drawPhysicBorders` to `true` and `spawnPlace` to `maze`.
- Start a new game.
- Fly left towards the nearest maze.
- Verify that all of the maze colliders are roughly the right shape and in the right place.
- Try flying through the maze passages.

# Notes
- This fixes #576.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/maze/MazeTileObject.java
+++ b/engine/src/main/java/org/destinationsol/game/maze/MazeTileObject.java
@@ -188,9 +188,9 @@ public class MazeTileObject implements SolObject {
                 tex.flip(!tex.isFlipX(), !tex.isFlipY());
                 backgroundTexture.flip(!backgroundTexture.isFlipX(), !backgroundTexture.isFlipY());
             }
-            RectSprite s = SpriteManager.createSprite(tex.name, MazeBuilder.TILE_SZ, 0, 0, new Vector2(), DrawableLevel.GROUND, 0, 0, SolColor.WHITE, false);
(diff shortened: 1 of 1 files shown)


Cervator:
I wonder if this also fixes #552 or #59 ? Different meshes tho.

BenjaminAmos:
Looking at this again, it seems like the maze collision issue was introduced in https://github.com/MovingBlocks/DestinationSol/commit/e5f6c68524ef8867d06c80c504bb127ccdde643a. The animated sprite code was never used after being implemented, so I've changed this PR to revert the changes made to `MazeTileObject` in that commit instead, which should be a better approach.

The approach previously in this PR 'fixed' the problem by never rotating the sprites. I only realised recently that this was incorrect. The new changes should actually fix the issue.

BenjaminAmos:
This pull request reverts the code to a known state where it previously worked, so there shouldn't be much risk in merging it without further delay.

PR #647, opened 2022-02-12
fix(world): make new world generation deterministic

BenjaminAmos:
# Description
When #607 was merged, it broke the determinism of world generation. This means that continuing an existing save would often end-up generating a completely different system than the one that was present when the game was saved. The precise reason for this issue is because the ordering of types returned by `ModuleEnvionment::getSubtypesOf` is undefined. I often found that the exact ordering of types in the returned list changed every time the game was run. The galaxy generation code is very sensitive to the ordering of these lists.

A simple solution to this problem would be to sort the lists by type name. This would not help though if additional feature generators were added later-on, since they could change the list ordering and size again. Because of this, I decided to save the generators used with the world configuration instead. This does mean that you will have to start a new game to see any new system/feature generators in use though.

I also updated some occurences of publicly-facing `ArrayList<T>` methods to the more generic `List<T>` in `GalaxyBuilder`, since I was modifying that code anyway.

It appears that some world generation tests were commented-out in #622. This pull request also restores those stubbed tests with working implementations again.

# Testing
- Start the game application
- Create a new game.
- Open the map to get a rough view of the system layout.
- Exit the game application completely through the main menu.
- Start the game application again.
- Continue from your previous save.
- Open the map and ensure that it contains an identical system layout to before.

# Notes
- With these changes, you will have to start a new game to have any new system/feature generators added subsequent to the world's creation included in the world generation process. The game should log an error if any generators are removed.

### DIFF
--- a/engine/src/main/java/org/destinationsol/SolApplication.java
+++ b/engine/src/main/java/org/destinationsol/SolApplication.java
@@ -350,7 +350,7 @@ public class SolApplication implements ApplicationListener {
         entitySystemManager.initialise();
 
         solGame.createUpdateSystems();
-        solGame.startGame(shipName, isNewGame, worldConfig, entitySystemManager);
+        solGame.startGame(shipName, isNewGame, entitySystemManager);
 
         if (!isNewGame) {
             try {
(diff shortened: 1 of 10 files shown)


BenjaminAmos:
@IsaiahBlanks, would you be interested in verifying this change? I'm not very familiar with the new world generation, so I might have made some mistakes somewhere.

BenjaminAmos:
These changes appear to work and are an improvement over the previous behaviour, so I see no issues in merging this right away.

PR #653, opened 2022-03-01
Rubble is mortal now!

Mystic-Slice:
# Description
The health component for the rubble was created but was not added to the entity itself. Fixed that. Now, the rubble is destructible.


BenjaminAmos (approved):
I've tested this myself and can confirm that this works as intended.

PR #654, opened 2022-03-02
fix #617(iii): Asteroids break into smaller pieces and then into tiny rocks

### LINKED ISSUE #617: Refactor new ECS rubble further
After merging #593 the "golden test asteroid" that spawns in front of the starting station now uses ECS-based rubble creation! There are some outstanding next steps however:

- [ ] The rubble starts with no momentum added - they should be flying all over the place (if they get bumped they also start acting bizarrely, like flying in circles)
- [ ] The rubble is invincible (same state before #593, normal asteroids are fine) and needs to be destructible
- [ ] The tubble goes straight to the tiniest size - a larger asteroid should first break into smaller pieces
- [ ] Smaller pieces should probably also have a chance to drop loot, but destroyed tiny rubble pieces should not?
- [ ] Maybe really a separate item: regular asteroids should be updated to the new system - but this may involve a fair bit more work including spawning, despawning, and random starting momentum/movement if any

If anybody new wants to work on this feel free - it isn't necessarily a Good First Issue since the topic is complex, but the area is relatively tiny and approachable and we can give more directions here or on Discord as needed in response to well-informed questions :-)

Mystic-Slice:
# Description
> Rubble created from destroying asteroid also now create rubble when destroyed. 
> Very tiny rocks do not create rubble further because the count of rubble produced by them will be 0.

# Testing
> * Break the golden asteroid.
> * Observe that it does not collapse into tiny pieces of rocks straight away. Instead, it acts like normal asteroids in the game.

**P.S:** The newly created pieces of asteroid do not break(fixed in #653). To test, add the health component to the rubble entities.


BenjaminAmos (commented):
I haven't tested this yet but a few things caught my eye whilst looking through the changes.

Mystic-Slice:
@BenjaminAmos Thnkx for the review. I have made the corrections.

Mystic-Slice:
Rebased to develop

BenjaminAmos (approved):
The fix works and the changes seem reasonable to me. I've never seen rubble as large as this before! I'll approve this but will wait for another review if possible, since I'm not too familiar with this area of the code,

PR #655, opened 2022-03-03
feat(input): remove fixed keyboard controls from GameOptions

BenjaminAmos:
# Description
This pull request moves the input control constants from `GameOptions` into a separate `DefaultOptions` enum. `GameOptions` has been changed to instead use a dynamic map of controls to their triggering inputs. This should allow for greater extensibility in the future, since modules can now define input controls. There is no way at present to persist the runtime-added controls yet.

An input control consists of an action that can be triggered by any number of input keys. The control itself contains the constant data describing what the control is and how it can be used. Controls are now associated with input types, which can be any of `KEYBOARD`, `MIXED`, `MOUSE` and `CONTROLLER`. Controls can be associated with multiple input types.

A control can now support an arbitrary number of keys as triggering inputs. This is implemented internally within `GameOptions` but currently nothing in the game supports it. Changing the input bindings from the main menu will replace all existing bindings for that control with the single binding chosen currently.

# Testing
- Play the game and make sure that all inputs work as before.
- For each input type, try changing the input bindings from the main menu. Ensure that the following operations work as before:
  - Display input binding
  - Change input binding
  - Save input bindings
  - Reset input bindings to defaults
- Restart the game after changing the input bindings and test again. Ensure that the changed inputs have persisted across saves.

# Future Improvements
These changes are not happening in this pull request but would be ideal as follow-ups.
- Support actually using multiple input bindings within the game e.g. all specified bindings should trigger a button, not just the first one.
- Populate the main menu inputs screen dynamically based on the input types specified by the controls.
- Allow the user to select both a primary button and a secondary button for each control from the main menu inputs screen.

# Notes
- This pull request moves all the keyboard input bindings which where previously stored in a series of constant variables. Because of this, reviewers should be aware that copy-paste errors are highly likely.
- It might be possible to port Terasology's input system to Destination Sol but that would require a considerably larger effort (and more ECS). This pull request attempts to adapt the existing code to be more future-proof instead.

### DIFF
--- a/engine/src/main/java/org/destinationsol/GameOptions.java
+++ b/engine/src/main/java/org/destinationsol/GameOptions.java
@@ -18,8 +18,17 @@ package org.destinationsol;
 import com.badlogic.gdx.Gdx;
 import com.badlogic.gdx.Input;
 import com.google.common.base.Enums;
+import org.destinationsol.input.DefaultControls;
+import org.destinationsol.input.InputControls;
 import org.destinationsol.menu.Resolution;
 import org.destinationsol.menu.ResolutionProvider;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import static java.util.Arrays.asList;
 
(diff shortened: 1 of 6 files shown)


soloturn:
Rebased onto current `develop`. Of CodeRabbit's 3 findings:

- **MERCENARY_INTERACTION label** was already fixed in e10a30d1 (now "Manage Mercenaries").
- **`Input.Keys.toString(-1)` crash risk** was also already addressed in e10a30d1 (invalid parsed keys are substituted with `Input.Keys.UNKNOWN` = 0, which is in-range and doesn't throw) — but that same change silently broke `save()`'s own `!= -1` filtering (dead code once invalid entries became 0 instead of -1), and `save()` had a separate structural bug on top: it only appended the built-up key string to `iniValues` inside the branch that checked the *last* input's validity, so one invalid trailing binding wiped out any earlier valid ones for that control. Fixed both in this push.
- **Mutable control storage** (`getControls()`/`getControl()` returning live references) wasn't addressed yet. Fixed: both now return cloned/unmodifiable snapshots, and `setControl()` clones its input before storing.

Compiles clean.

soloturn:
Filed #739 to track: persisting runtime-registered controls (the design needed was discussed above), plus the multiple-bindings/dynamic-screen/primary+secondary items from this PR's own "Future Improvements" section, and the Terasology-input-system-port idea from "Notes".

soloturn:
@NicholasBatesNZ addressed, along with CodeRabbit's overlapping/follow-on findings in the pushes above:

- `Input.Keys.toString(-1)` crash: fixed differently than suggested — invalid parsed keys are substituted with `Input.Keys.UNKNOWN` (0, in-range) rather than falling back to `getDefaultInputs()`, so an unrecognised binding shows as unbound instead of silently reverting to a default. `save()`'s serialization and `setControl()`'s validation are both updated to match (full `UNKNOWN..MAX_KEYCODE` range, not just `-1`).
- `MERCENARY_INTERACTION` label: already fixed.
- HashMap ordering: switched to `LinkedHashMap` rather than `EnumMap` — `EnumMap`'s key type is fixed to one enum, which would rule out `InputControls` ever holding a non-`DefaultControls` (module-registered) key, the whole point of this PR moving off the enum type. `LinkedHashMap` gets the same stable insertion-order behavior without that constraint.

Filed #739 for the runtime-persistence gap and the PR description's own Future Improvements/Notes items.

PR #656, opened 2022-03-09
feat: make the golden asteroid at spawn optional

BenjaminAmos:
# Description
This pull request makes spawning the golden asteroid conditional on enabling a debug option. This option is on by-default in from-source workspaces.

Whilst the ECS asteroid is a useful debugging tool, I do not believe that it is intended for regular players to use.

# Testing
- Play the game and the golden ECS asteroid should spawn as usual.
- Close the game.
- Change `spawnECSAsteriod` in `debugOptions.ini` to `false`.
- When starting a new game, the golden asteroid should no longer spawn.


BenjaminAmos:
This is a useful change that shouldn't affect anything else. No gameplay features use ECS yet, so it's completely separate. As such, I think it's fine to merge right away.

PR #661, opened 2022-06-06
doc: create bug report issue form

BenjaminAmos:
# Description
Since I was looking from the project documentation anyway, I thought I'd try-out the new (beta) issue forms. This makes the new issue process look a bit nicer, although that's about it. I've mostly re-used the existing issue template for this form, with some areas re-worded.

# Testing
You can preview the form [here](https://github.com/BenjaminAmos/DestinationSol/blob/issue-forms/.github/ISSUE_TEMPLATE/bug_report.yaml) and test it by creating a new issue on https://github.com/BenjaminAmos/DestinationSol.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,0 +1,65 @@
+name: "Bug report"
+description: "Report an issue with the game"
+labels: ["bug"]
+body:
+  - type: markdown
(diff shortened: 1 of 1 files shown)


BenjaminAmos:
For non-code PRs like this, is there a way of skipping CI so that it doesn't waste resources when there's absolutely nothing new to test?

NicholasBatesNZ (approved):
Woah this is neat!

PR #671, opened 2022-10-25
feat: display ship preview on new ship screen

BenjaminAmos:
# Description
This pull request adds a preview of the player's chosen ship to the new ship screen.
image

# Testing
- Start the game
- Select the `New Game` option
- View each starter ship option individually and ensure that the correct corresponding sprites are shown on-screen.
- (As is usual for UI changes, test that elements don't overlap too much at different resolutions)


NicholasBatesNZ (approved):
Neat change, I like it.
The game crashes if the spawn config has been edited and includes a ship that can't be found. I can't decide if that is fine or not, so I'll leave it up to you.

BenjaminAmos:
The game will crash regardless when playing the game if an invalid ship is specified but I think it's better to crash later-on than in the main menu. The asset not found exception should be caught in the menu now and a more useful error found in the logs.

PR #673, opened 2022-12-31
build: modernise and modularise gradle build scripts

BenjaminAmos:
# Description
This pull request updates gradle to version 7.6, removes uses of deprecated APIs and generally cleans things up. `apply plugin: ` usages have been replaced with pre-compiled script plugins. The IntelliJ project generation should be working again (gestalt-di broke it), although you'll have to use "Import Gradle Project" rather than `gradlew idea`.

I have tried to split-out most of the code that deals with compiling gestalt modules in the hope that it may someday be shared with Terasology again. Destination Sol specific additions are added via a plugin which includes the generic gestalt plugin.

All modules should be generated with sufficient metadata to be publishable to a maven repository. This should make it easier to obtain modules for compiling against when developing new modules that depend on them.

It's now possible, with a bit of work, to compile modules for the game from outside of the game's workspace. As far as I am aware this was probably intended but issues with the organisation of the gradle build scripts made it difficult to achieve previously.

Essentially, it changes module build files from their current 239-line-long state to the following:
```gradle
plugins {
    id 'destination-sol-module'
}
```
Almost everything else is done to support that change.

# Testing
Everything that is handled by the gradle scripts should be re-tested after this change. The most important things are:
 - Compiling the game's code
 - Generating a zipped desktop distribution
 - Generating an Android APK distribution
 - Building and running modules in-game (`warp` in particular)

# Notes
- I changed the copyright header template to say 2023. It's only a day away, so accurate enough.


BenjaminAmos:
The PR build is failing with an error that I cannot reproduce. I'm not sure as to the cause of it right now.
```log
Execution failed for task ':android:lintAnalyzeDebug'.
> Couldn't read file content: '/home/jenkins/agent/workspace/DestinationSol_engine_PR-673/android/assets/modules/core/build/dexes'.
```

BenjaminAmos:
This breaks the `groovyw` tool due to MovingBlocks/groovy-wrapper#6.

BenjaminAmos:
MovingBlocks/groovy-wrapper#7 fixes the `groovyw` tool but I wouldn't consider it a hard-dependency on merging this.

BenjaminAmos:
The groovy wrapper changes are in #680. The game doesn't need the groovy wrapper functioning to build.

PR #683, opened 2023-04-17
fix: fix pause key not pausing the game

BenjaminAmos:
# Description
This pull request fixes the pause key, so that it actually pauses and un-pauses the game again.

What it really does is remove the old pause key handling code in the original `MainGameScreen`, since it was interfering with the new pause key handling code from its NUI equivalent.

# Testing
- Start a new game (or continue an existing one).
- Move your ship around a bit.
- Press the <kbd>P</kbd> key to pause the game.
- Your ship should stop moving and the game should be paused.
- Press the <kbd>P</kbd>key again to un-pause the game.
- Your ship should continue moving.


Cervator (approved):
Tests out locally! Merging.

Am reminded Jenkins is broken until I get a new Android SDK on the build agent image. Adding to list with maybe resolution in just over a week :-)

PR #684, opened 2023-04-22
feat(ui): separate ability icons from their charges

BenjaminAmos:
# Description
This pull request allows for abilities to be assigned a separate icon to that used by its charges. Not all abilities use ability charges. You might also want to have different icons to represent the ability and the charges that it consumes

This is needed to fix an issue where the ability bar was not being displayed for abilities without an associated charge item..

# Testing
- Start a new game with any of the imperial ships (Imperial Large, Imperial Tiny, etc.).
- Verify that the ability bar is present and that it shows how many charges you have next to it.
- Checkout the warp module (`groovyw module get warp`) and switch it to the `ability-icons` branch.
- Verify that the ability bar is present when playing as the `Warp Quasar` ship.

# Notes
- This shouldn't require any changes to existing abilities.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/AbilityCommonConfig.java
+++ b/engine/src/main/java/org/destinationsol/game/AbilityCommonConfig.java
@@ -15,6 +15,8 @@
  */
 package org.destinationsol.game;
 
+import com.badlogic.gdx.graphics.g2d.TextureAtlas;
+import org.destinationsol.assets.Assets;
 import org.destinationsol.assets.sound.OggSound;
 import org.destinationsol.assets.sound.OggSoundManager;
 import org.destinationsol.assets.sound.PlayableSound;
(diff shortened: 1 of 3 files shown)


Cervator:
This makes sense and I can get through all the steps but - I not only lack an ability bar, I can't seem to trigger the ability at all, even waiting well over 10 minutes. I even dropped its cooldown down to 6 seconds and started fresh but nope. Nothing obvious in the logs. I honestly didn't even realize it had an ability!

Double checked that I had the changes locally both in engine and warp. Checked logs both times and nothing out of the ordinary appeared. Puzzling!

Imperial ships show the bar both before and after the change, and the ability itself works fine. With warp I just don't have a bar at all with the Quasar:

image

BenjaminAmos:
Does the research button show-up when playing as the `Warp Scout` ship? It's possible that the warp module wasn't compiled, so none of the code has actually been loaded. Make sure that you're using the [`ability-icons`](https://github.com/DestinationSol/warp/tree/ability-icons) branch of the warp module.

Re-building the warp module in IntelliJ or `gradlew :modules:warp:jar` might help.

The `Warp Endeavour` ship has a significantly smaller ability cooldown, so it might be easier to test with. Here's what it should look like, anyway.
image

Cervator:
That was it. Been so long :-) For Terasology I think we have some obscure Gradle piece figuring out dynamically which extra modules to build to run the game, but I guess `gradlew run` doesn't quite do that for DS (and I'm not really sure if that works in all cases or at all for TS anymore even)

Also, lol at 6 sec cooldown on that ability 😁 

image

Cervator (approved):

PR #703, opened 2025-01-04
feat!(factions): factions system re-work

BenjaminAmos:
# Description
Factions are now defined independently and can hold relations with each other. The player has their own faction which holds separate relationships with each faction. Factions holding a poor reputation will be hostile.

The original `Faction` enum has been removed and `FactionInfo`'s functionality has been moved into `FactionManager`, with data now held in a new `Faction` class.

This is a foundation for potential future work and does not expose any new gameplay.

The factions descriptions are placeholders. I am sure that they could be improved.

# Testing
- Start a new game with the `Loaded Imperial Large` preset.
- Show the factions display using the <kbd>Z</kbd> key.
- The starter station and guards should be from the `Imperial` faction.
- Flying around you should encounter hostile ships from the `Pirate` faction.
- Hostile ships from the `Desert` faction can be encountered on desert planets.

# Notes
- Hired mercenaries now have the same faction as the player.

# Known Issues
- Enemies spawned via `GalaxyBuilder` and `ChunkBuilder` may not be hostile if their native faction is not hostile to the player faction. This is unlikely to be encountered with the default configuration.


Cervator (approved):
Seems to work well, and as always the code and docs are well written. Feels a bit like getting back to the good ole days! :-)

I tested on PC so far, confirmed I can see Imperial and Pirate tags with `z` enabled - although I noted that the tag disappears when you enter planetary gravity so I couldn't see any Desert tags there (I imagine they'd work out in space)

Also confirmed I could make the Imperials mad by shooting at the starting station and guards 👍

PR #704, opened 2025-01-06
feat: make engines equippable

BenjaminAmos:
# Description
This pull request makes engines equippable in-game, allowing you to change the engine used by your ship. It adds mention of engines to the tutorial, as well as a warning for when no engine is currently equipped.

The implementation here is a bit more complicated since ships always come with an engine defined in their hull. We could decouple the two entirely, requiring ship definitions to contain an engine item to be spawnable, however, this would break all existing ship definitions that we currently have. For now, this half-way solution seems more palatable.

# Testing
- Start a new game as any ship.
- Check that an engine item is visible in your inventory.
- Close the game and edit your save file to add the following items: `core:minerEngine core:desertEngine`.
- Try equipping the `Miner Engine` item. You should still be able to move around.
- Unequip the `Miner Engine` item. You should not be able to move around.
- Try equipping the `Desert Engine` item. You should be able to move around again.
- Exit to the main menu (this should save your game).
- Continue your previous game and the `Desert Engine` engine should still be equipped.

# Notes
- Currently all ship engines are inheriting the default values, so do not expect to see changes in speed when switching engines. Adding some custom values to the engine's JSON definition file (implemented by #702) should allow for this in-future.


Cervator (approved):
Tests out fine 👍 

I also tried throwing away all my engines, exiting to the main menu, then continuing the game. Was happy to find that I had a copy of the default engine again - I can imagine it otherwise just being a question of time until somebody loses their engine and can't get a new one :-)

I wonder if we need a note about being able to do that (maybe in the warning label when you have no engine in small test), or if you have _no_ engine still allow a tiny amount of trust so you can limp to a station? 🤔 Doesn't have to be part of this PR, maybe a separate issue as a reminder.

BenjaminAmos:
The engine coming back is technically a bug but one I was aware of. It is to do with the complications I mentioned in the description. Until we are certain if this should deliberately be a feature, I will probably defer documenting it. As much as it is a bit unfair, I think the game gives you enough chances at self-sabotage as it is (sell all shields, amour and waste all money).

A tiny amount of thrust (emergency thrusters?) instead would probably be better. I will open an issue for that, as both issues should ideally be resolved together.

PR #709, opened 2025-03-22
fix: constantAllies and constantEnemies are now always friendly or hostile respectively

### LINKED ISSUE #707: All constantEnemies are friendly since faction update

Fix the bug which I originally thought was caused by my module

### What actually happened

After experimenting with removing the module altogether so only the "core" module remains, the constantEnemies still remained friendly.

### How to reproduce

Step 1: Start a new game or continue existing game
Step 2: Press tab and look at the map for pirate trucks and small pirate ships in the spawn system, and medium desert ships in the non-spawn systems. They should be blue.
Step 3 (optional): Fly up to one of them, then realize they are serious about not wanting you dead.

### Game version

Latest source build

### Log details

_No response_

### Operating System

Windows

### Additional Info

_No response_

BenjaminAmos:
# Description
This pull request partially reverts changes in #703 by forcing ships specified in `constantAllies` and `constantEnemies` to use the pre-defined factions to ensure predictable behaviour with regard to hostility.

# Testing
- Start a new game (any ship will do)
- Fly inwards until you encounter a pirate ship
- The pirate ship should be hostile and start shooting at you immediately

# Notes
This fixes #707 .

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/GalaxyFiller.java
+++ b/engine/src/main/java/org/destinationsol/game/GalaxyFiller.java
@@ -162,7 +162,7 @@ public class GalaxyFiller {
                 int count = (int) (shipConfig.density);
                 for (int i = 0; i < count; i++) {
                     // TODO: Select an appropriate ally faction based on the player faction.
-                    build(game, shipConfig, game.getFactionMan().getBuilderForHull(shipConfig.hull), false, system, angles);
+                    build(game, shipConfig, game.getFactionMan().getGenericAllyFaction(), false, system, angles);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Du4lity5151:
Works as expected. As far as I can concern in this commit, the only thing it does with constant enemies is changing their faction to a generic one with a disposition of -100, but the pirate and desert faction files also both have a negative disposition, so it is unclear why the constant enemies remained friendly before the faction change, and how applying the change resolved the issue.

BenjaminAmos:
This is a small change that changes things is behave as expected. Merging this should be fine.

PR #710, opened 2025-04-02
fix: fix default faction dispositions not being applied

### LINKED ISSUE #706: Mercenaries do not fire

Fight an enemy using a mercenary

### What actually happened

The mercenary does not fire, even when checked to ensure it's weapon is equipped and there is ammo, as well as attempting through the use of a heavy blaster, an ammo-free weapon.

### How to reproduce

Step 0 (optional): Add a mercenary in systemsConfig in the core module that uses a blaster or heavy blaster to ensure ammo is not an issue.
Step 1: Start a new game or continue game
Step 2: Travel to a station with a mercenary (ideally the one from step 0)
Step 3: Purchase mercenary
Step 3.5: Equip mercenary with a weapon and appropriate ammo if applicable.
Step 4: Fly up to an enemy, stay near it, do not fire at it, and watch as your unfortunate merc cannot do anything to save your life.

### Game version

Latest source build

### Log details

_No response_

### Operating System

Windows

### Additional Info

_No response_

BenjaminAmos:
# Description
The original implementation of the faction loading code did not correctly handle factions lacking explicit relations to each other. This lead to situations where a faction could be hostile to another but not reciprocally. Whilst such situations are allowed when explicit (it could lead to interesting gameplay mechanics), it should not be the case by default.

As commented in the new code, this pull request changes how faction relations are initially assigned when neither faction is aware of the other:
```java
// The simplified rules of uncertain Destination Sol diplomacy:
//   - If both are friendly, the stronger positivity will prevail.
//   - If both are hostile, the stronger hostility will prevail.
//   - If your enemy is hostile to you, you must be hostile to your enemy.
//   - Neutrality is considered friendly.
```

# Testing
Try following the instructions in #706 to ensure that the bug is no longer present. Mercenaries hired by the player should be correctly hostile to the player's enemies now.

# Notes
This fixes #706.

### DIFF
--- a/engine/src/main/java/org/destinationsol/game/FactionManager.java
+++ b/engine/src/main/java/org/destinationsol/game/FactionManager.java
@@ -217,7 +217,7 @@ public class FactionManager {
      * @return true, if f1 and f2 are enemies, otherwise false.
      */
     public boolean areEnemies(Faction f1, Faction f2) {
-        return f1 != null && f2 != null && f1.getRelation(f2) < 0;
+        return f1 != null && f2 != null && (f1.getRelation(f2) < 0 || f2.getRelation(f1) < 0);
     }
 
     private static class MyRayBack implements RayCastCallback {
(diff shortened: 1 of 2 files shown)


Du4lity5151:
Can confirm, this fixes the issue and does not appear to cause any side effects. Took down an entire imperial station and its guards post-hostility with several mercenaries who were using the same ship as the guards. It was hilarious to watch!

PR #724, opened 2026-01-22
feat(ui): introduce UIAnimatedImage for animated sprites and use in ship preview

### LINKED ISSUE #723: Spritesheets are displayed instead of animations if not in-game

While in the main menu, the game should display animations instead of spritesheets.

### What actually happened

if an animated ship sprite is displayed in the menu background or on the ship selection screen, the ship's spritesheet is shown instead of the animation.

### How to reproduce

Step 1: Download [vanguardelite](https://github.com/DestinationSol/vanguardelite) module and place it in the modules folder
Step 2: Launch the game
Step 3: Start a new game, then click on the ship button until it shows the Vanguard Elite

### Game version

Commit 52b23c4

### Log details

N/A

### Operating System

Windows

### Additional Info

Adding support for animations within the main menu would be the most logical solution. However, an alternative solution if not feasible would be to take a separate static sprite and display it instead of the spritesheet if available. Otherwise, ship animations function correctly while in-game.

BenjaminAmos:
# Description
This pull request introduces a new UI widget, called `UIAnimatedImage`, that allows animated images to be rendered within NUI UI screens. The API is similar to that of `UIImage`, however, the `image` binding has been replaced with three separate bindings:
    - `spritesheet` (`UITextureRegion`) - The spritesheet to use for rendering.
    - `frameDuration` (`float`) - The number of seconds to elapse between each frame shown
    - `frames` (`List<Rectanglei>`) - The pixel co-ordinates of each individual frame.

The new UI widget is used to render ship previews on the main menu, enabling support for animated ship sprites to be correctly displayed.

This pull request also changes the rendering logic for the main menu background ships to correctly render animated ship sprites.

# Testing
Follow the instructions #723 and verify that the expected behaviour is seen.

# Notes
- There appears to be a slight flickering when rendering the `imperialBigElite` ship from the `eliteVangard` module. It is unclear whether this is a spritesheet design or renderer issue.
- This fixes #723.

### DIFF
--- a/engine/src/main/java/org/destinationsol/menu/background/MenuBackgroundAsteroidManager.java
+++ b/engine/src/main/java/org/destinationsol/menu/background/MenuBackgroundAsteroidManager.java
@@ -16,6 +16,7 @@
 package org.destinationsol.menu.background;
 
 import com.badlogic.gdx.graphics.Color;
+import com.badlogic.gdx.graphics.g2d.Animation;
 import com.badlogic.gdx.graphics.g2d.TextureAtlas;
 import com.badlogic.gdx.math.Vector2;
 import com.badlogic.gdx.physics.box2d.Body;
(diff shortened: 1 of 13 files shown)


Du4lity5151:
This fix appears to work mostly correctly, however the animation plays slower in the background if I add the imperialBigElite as a background ship in menuBackgroundShipConfig. If you notice flickering in the siren lights themselves, then it might be normal as the lights flash for one frame and then go off for the next two frames, and repeats at 48 FPS. Otherwise, I did not notice any other issues.

PR #728, opened 2026-02-07
Steam Workshop Support

BenjaminAmos:
# Description
This pull request contains the changes needed to implement Steam Workshop support into Destination Sol. More specifically:
 - Extending JSON delta support to cover NUI UI files
 - Extracting the majority of desktop launcher logic held in `SolDesktop` into a shared `DesktopLauncher` utility class to be shared with the Steam facade.
 - Moving all of the Steam-related code and assets into a separate [DestSolSteam](https://github.com/MovingBlocks/DestSolSteam) facade, with its own separate repository.

# Testing
### Availability
- I am not sure how much I can describe publicly about the Steamworks side of things. Those with publishing access already have access but otherwise you can view the workshop by joining the [test Steam group](https://steamcommunity.com/groups/destination-sol-workshop-testing) and then visiting https://steamcommunity.com/app/342980/workshop/. The workshop page is not yet publicly accessible to all in order to avoid any confusion at this early stage.
- A build derived from this pull request has also been published to a 'private' release branch on Steam, accessible using the access code `steamworkshop`. This is updated manually and is unlikely to be up-to-date.
### Installation
- Firstly, fetch the Steam facade with `gradlew fetchSteam`.
### Obtaining modules from the workshop
- You can run the Steam facade with `gradlew runSteam`.
- Try creating a new game and navigating to the modules screen. Then press the `Steam Workshop` button.
- Select the `Browse` filter and subscribe to the `NUI Experiments` module.
- Close down the game and re-start it. Then go back to the modules screen and verify that the `NUI Experiments` module is now available.
- Start a new game with that module enabled and verify that a test UI screen is present by running the command `showNUIScreen nuiExperiments:warnButtons`.
- Go back to the main menu, return to the Steam Workshop screen and unsubscribe from the `NUI Experiments` module.
- Close down the game and re-start and verify that the module is no longer available.
### Uploading modules to the workshop
- Clone one of the modules from the [DestinationSol](https://github.com/DestinationSol/) organisation on GitHub into your `modules` directory.
- When on the Steam Workshop screen, click on the `Workshop Upload` button.
- Select the module you just cloned and click `Upload`.
- Enter an arbitrary update note and finally click on the `Upload` button again.
- You should now be able to see the module when you click on the `Your Uploads` button when on the Steam Workshop screen.
- You should also be able to see the workshop entry listed from within the Steam client somewhere. Where exactly is hard to predict though.

# Notes
- New modules may be available after subscribing without having to restart the game. This is experimental though and may not always work.
- Eventually, I would like to move all the modules apart from core that were bundled with the game on Steam instead onto the workshop.
- CI changes are untested. They can be re-visited at another time.


NicholasBatesNZ:
Thanks for this — really nice piece of work, and the split into a separate `DestSolSteam` facade reads well.

The main thing standing between it and mergeable is a rebase on `develop` after the Gradle 9.6.1 / Java 17 upgrade (#736). The only conflict is a single line in `SolDesktop.main()` where `e.printStackTrace()` became `logger.error("Failed to open splash screen", e)` — that change just needs carrying over into the new `DesktopLauncher`. I tried the merge locally and with that one line resolved, `:engine:compileJava`, `:desktop:compileJava` and `:engine:compileTestJava` all pass, so the rebase is genuinely trivial.

On the Jenkins failure — apologies, my earlier comment asked for a companion Android PR that you'd already opened. MovingBlocks/DestSolAndroid#34 is exactly that, and it's still open and mergeable. The red build here is `:android:compileDebugJavaWithJavac` failing on "AndroidModuleConfig is not abstract and does not override abstract method getModulePaths()", because CI clones the Android facade's default branch rather than your branch, so it builds against the un-updated version. So this is really a landing-order question between the two PRs rather than anything missing. If you'd rather not have to land them in lockstep, a `default` bridge method on `FacadeModuleConfig` would decouple them, but that's your call.

Two smaller things while you're in there: the new `Build Steam` stage's `when { tag ...; branch ... }` block is ANDed by declarative pipeline so it'll never fire — it probably wants `anyOf { }`; and `UIDeltaFormat.apply()` re-reads the base JSON from `assetData.getSource()` each call, so a second delta on the same `.ui` file overwrites the first rather than stacking on it.

soloturn:
Rebased on develop, and addressed the three remaining points from @NicholasBatesNZ's review:

- `DesktopLauncher.launchGame()` now uses `logger.error(...)` for a failed splash-screen init instead of `e.printStackTrace()` - carrying over the same fix develop made before this branch extracted the logic.
- Jenkinsfile's `Build Steam` stage now wraps its `tag`/`branch` conditions in `anyOf {}` so either one triggers it, instead of both being required (impossible) under declarative pipeline's default AND.
- `UIDeltaFormat.apply()` now accumulates JSON state per `UIData` instance across calls instead of re-reading the original base source every time, so a second delta on the same `.ui` file stacks onto the first instead of overwriting it.

`:engine:compileJava`/`:desktop:compileJava` verified clean. The Android/landing-order question is still open - that's a call for @BenjaminAmos.

PR #734, opened 2026-04-01
POC only: Bifrost

Cervator:
Not for merging, at least not yet! Details at https://siliconsaga.net/bifrost-first-contact/


BenjaminAmos:
I am well aware that this is only a proof of concept, however, would it be possible to implement this is an bundled external module, like the Steam Workshop integration in #728? That might require making this depend on some of the changes in #728 though. The idea is that, like with Steam integration, this is entirely optional without any code loaded if not desired.

It might make sense long-term to have some of the functionality in the core engine but as-is this demo makes a few instrusive changes to classes and UIs which could have been handled differently. Destination Sol is still largely a monolithic application but over time I have been trying to break it up into a more extensible architecture, even if it is nowhere near Terasology levels of extensibility.

